PR #243, opened 2025-09-03
fix: created time now gets updated properly. 

rehanthestar21:
# Description

Please provide a meaningful description of what this change will do, or is for.
Bonus points for including links to related issues, other PRs, or technical
references.

Note that by _not_ including a description, you are asking reviewers to do extra
work to understand the context of this change, which may lead to your PR taking
much longer to review, or result in it not being reviewed at all.

Please ensure commits conform to the [Commit Guideline](https://www.conventionalcommits.org/en/v1.0.0/)

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/ai_platform_engineering/knowledge_bases/rag/server/rag_api.py
+++ b/ai_platform_engineering/knowledge_bases/rag/server/rag_api.py
@@ -53,8 +53,8 @@ class CollectionConfig(BaseModel):
     url: Optional[str] = Field(None, description="Source URL for the collection")
     chunk_size: int = Field(10000, description="Size of text chunks", ge=100, le=50000)
     chunk_overlap: int = Field(2000, description="Overlap between chunks", ge=0, le=5000)
-    created_at: datetime.datetime = Field(default_factory=lambda: datetime.datetime.now(datetime.timezone.utc))
-    last_updated: datetime.datetime = Field(default_factory=lambda: datetime.datetime.now(datetime.timezone.utc))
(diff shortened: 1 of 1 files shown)


subbaksh (changes requested):
Lets refactor this code a little bit

Imo there are a few concepts here:
 - A data source (e.g. URL)
 - A document (i.e. a Webpage)
 - A collection (a collection of document embeddings) - This concept is unique to Milvus. but we can use and abstract it out

We need to store metadata in Redis about a few things
 - Collection config - the metadata for this should have the name, the metric type (sparse/dense), the similarity function (cosine etc.). This is something that wont be used as much, because most of the docs will be in one collection. Multiple collections maybe used in the future if for e.g. different agents wnat to store different data to query.
 
 - A Data source: This is a source which contains a set of documents. In our use-case its a URL. But it could also be S3 bucket, a local folder etc. Anything that contains Documents. We dont need to store any metadata here (see beloiw).
 
 - A Document: This is just a file (text, single webpage, json etc.). The document may or may not be chunked. We NEED to store metadata on each document, 
    - The name (e.g. web page name)
    - what the DataSource of the document is (e.g. URL)
    - when it was last updated.
    - Chunk size + strategy
    
The UI should show the list of documents, and the metadata. The collection configuration should be under "Advanced settings" , as most folks wont really change it much.

PR #253, opened 2025-09-07
fix: optimize memory usage by streaming page processing

sriaradhyula:
## Problem
The kb-rag-server was experiencing OOMKilled errors when processing large sitemaps (1072 pages) because it loaded all pages into memory simultaneously using `ascrape_all()`.

## Root Cause
- `ascrape_all(urls)` loads all 1072 pages into memory at once (~4-8GB)
- No memory management between page processing
- Container gets OOMKilled even with 8GB memory limit

## Solution
- Replace batch loading with streaming: `ascrape(page_url)` processes one page at a time
- Add garbage collection after each page to free memory
- Add error handling to continue processing on individual page failures
- Memory usage drops from 4-8GB to 4-8MB per page

## Testing
- Tested with platform-docs.outshift.io sitemap (1072 pages)
- No more OOMKilled errors
- Maintains same functionality with better memory efficiency

Fixes the OOMKilled containers during large sitemap ingestion reported in production.

### DIFF
--- a/ai_platform_engineering/knowledge_bases/rag/server/loader/loader.py
+++ b/ai_platform_engineering/knowledge_bases/rag/server/loader/loader.py
@@ -382,7 +382,7 @@ class Loader:
                 )
                 return
 
-            # Load documents from URLs
+            # Load documents from URLs with streaming processing
             for sitemap_url in sitemaps:
                 self.logger.info(f"Loading sitemap: {sitemap_url}")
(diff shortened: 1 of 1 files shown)


subbaksh:
Yes, this is a good one :)
This is a good solution, thres also lazy loading that the web loaders have for this as well:
https://python.langchain.com/docs/integrations/document_loaders/web_base/#lazy-load

PR #308, opened 2025-09-18
feat(helm): add missing agent deployments for aws, splunk, webex, komodor

sriaradhyula:
## Summary
Adds missing agent deployment configurations and bumps helm chart version to 0.2.3.

## Changes
- **Chart.yaml**: Bump version to 0.2.3 and add agent dependencies for aws, splunk, webex, komodor
- **values.yaml**: Add agent configurations following existing pattern

## Details
- Each new agent follows the same structure as existing agents (argocd, backstage, etc.)
- Uses  conditions for conditional deployment
- Enables deployment of these agents as separate pods instead of only within supervisor-agent

## Testing
- Helm template validation passes
- Follows existing agent configuration patterns
- Ready for deployment with enabledSubAgents configuration

Fixes missing agent deployments when enabledSubAgents are configured in override.yaml.

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -121,6 +121,58 @@ agent-slack:
     port: 8000
     pullPolicy: "Always"
 
+agent-aws:
+  nameOverride: "agent-aws"
+  image:
+    repository: "ghcr.io/cnoe-io/agent-aws"
+    pullPolicy: "Always"
+  mcp:
+    image:
+      repository: "ghcr.io/cnoe-io/mcp-aws"
+      tag: "stable"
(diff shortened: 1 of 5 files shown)


helm/values.yaml
suwhang-cisco, line 1: We need to add the new agents under `global. enabledSubAgents` as well (line 10)

suwhang-cisco (commented):
agent values look good but some missing from values.yaml as my comment above. Could you also populate values-external-secrets and values-secrets.example so people know the expected keys for the secrets in order for these new agents to work?

PR #311, opened 2025-09-19
feat: add SHARED key authentication option for A2A protocol

sriaradhyula:
## Changes
- Add shared key middleware for A2A authentication
- Move OAuth2 middleware to common auth module  
- Add Keycloak deployment configuration
- Update documentation for A2A authentication methods

## Documentation
- Added security documentation for A2A authentication
- Updated Docker Compose setup guide
- Added Keycloak deployment README


ai_platform_engineering/common/auth/shared_key_middleware.py
```diff
@@ -0,0 +1,116 @@
...
+load_dotenv()
+
+logger = logging.getLogger(__name__)
+
+USE_SHARED_KEY = os.getenv('USE_SHARED_KEY', 'false').lower() == 'true'
+
+if USE_SHARED_KEY:
+  SHARED_KEY = os.environ["SHARED_KEY"]
+  if not SHARED_KEY:
+    raise ValueError('SHARED_KEY is not set')
```
subbaksh: Instead of two env vars, why not just check if `SHARED_KEY` exists, and if it does use that :)

subbaksh: Maybe a good name would be `A2A_AUTH_SHARED_KEY`

subbaksh (approved):

PR #313, opened 2025-09-19
feat: add idpbuilder deployment configurations

sriaradhyula:
This PR adds idpbuilder deployment configurations with updated values and clean implementation.


helm/values.yaml
```diff
@@ -194,167 +194,167 @@ backstage-plugin-agent-forge:
         port: 7007
         protocol: TCP
 
-graphrag:
```
suwhang-cisco, line 197: why are we commenting out these values? we can set graphrag.enabled as false etc. but keeping the values would be helpful so user doesn't have to figure out which to uncomment but simply turn the flag on/off

PR #399, opened 2025-10-23
fix(auth): Fix shared key authentication not loading middleware

### LINKED ISSUE #398: Bug: Shared Key Authentication Not Working
# Bug Report: Shared Key Authentication Not Working

## Description

Shared key authentication is not being enforced despite being properly configured. When setting `A2A_AUTH_SHARED_KEY` with an actual secret key value, the authentication middleware never loads, allowing unauthenticated requests through.

## Steps to Reproduce

1. Configure shared key authentication:
   ```yaml
   env:
     A2A_AUTH_SHARED_KEY: "your-secret-key-here"
   ```

2. Deploy the supervisor agent

3. Test authentication:
   ```bash
   # Should fail with 401, but succeeds
   curl -X POST https://your-endpoint/protected-path
   ```

## Expected Behavior

- Requests without authentication should return `401 Unauthorized`
- Requests with valid Bearer token should succeed
- Requests with invalid token should return `401 Unauthorized`

## Actual Behavior

All requests succeed regardless of authentication because the middleware never loads.

## Root Cause

The issue is in `ai_platform_engineering/multi_agents/platform_engineer/protocol_bindings/a2a/main.py` at line 114:

```python
A2A_AUTH_SHARED_KEY = os.getenv('A2A_AUTH_SHARED_KEY', 'false').lower() == 'true'
```

This line checks if the environment variable equals the string `"true"` rather than reading the actual key value. This causes:

1. When you set `A2A_AUTH_SHARED_KEY=your-secret-key`, the check `"your-secret-key".lower() == 'true'` returns `False`
2. The middleware is never loaded since the if condition on line 124 evaluates to False
3. All requests pass through without authentication

Meanwhile, `shared_key_middleware.py` expects the actual key value:

```python
A2A_AUTH_SHARED_KEY = os.getenv('A2A_AUTH_SHARED_KEY')
```

This creates a contradiction where the same environment variable is expected to be both a boolean flag and an actual key value.

## Git History

This bug was introduced in commit **856badd1** (Sep 19, 2025 at 12:35 PM), just 6 minutes after commit **821d68e9** properly fixed the authentication setup.

**Commit 821d68e9** (correct):
```python
A2A_AUTH_SHARED_KEY = os.getenv('A2A_AUTH_SHARED_KEY')
```

**Commit 856badd1** (introduced bug):
```python
A2A_AUTH_SHARED_KEY = os.getenv('A2A_AUTH_SHARED_KEY', 'false').lower() == 'true'
```

## Proposed Solution

Revert line 114 to the correct implementation from commit 821d68e9:

```python
A2A_AUTH_SHARED_KEY = os.getenv('A2A_AUTH_SHARED_KEY')
```

This allows the middleware to load when a key is set (truthy value) and uses that same key for validation.

## Testing

After the fix:

```bash
# Without token - should fail
curl -X POST https://endpoint/path
# Returns: 401 Unauthorized ✓

# With valid token - should succeed
curl -H "Authorization: Bearer your-key" -X POST https://endpoint/path
# Returns: 200 or appropriate response ✓

# With invalid token - should fail
curl -H "Authorization: Bearer wrong-key" -X POST https://endpoint/path
# Returns: 401 Unauthorized ✓
```

## Environment

- Version: latest main branch (commit 856badd1 and after)
- Component: supervisor-agent / A2A protocol binding
- Impact: Critical - Authentication completely bypassed

## Additional Context

The same pattern used for OAuth2 (line 112) has a similar structure but doesn't have this issue because OAuth2 uses separate environment variables for the flag and configuration.

A future improvement could be to follow the OAuth2 pattern more closely or simply check if the key is set and non-empty rather than comparing to "true".

rayl15:
# Fix: Shared Key Authentication Middleware Not Loading

## Problem

Shared key authentication was completely non-functional due to a logic error introduced in commit 856badd1. The authentication middleware would never load, regardless of configuration, leaving all endpoints unprotected.

## Root Cause

Line 114 in `main.py` was checking if `A2A_AUTH_SHARED_KEY` equals the string `"true"`:

```python
A2A_AUTH_SHARED_KEY = os.getenv('A2A_AUTH_KEY', 'false').lower() == 'true'
```

When users set the actual secret key (e.g., `A2A_AUTH_SHARED_KEY=abc123...`), this check would always return `False`, preventing the middleware from loading.

## Solution

This PR reverts line 114 to the correct implementation from commit 821d68e9:

```python
A2A_AUTH_SHARED_KEY = os.getenv('A2A_AUTH_SHARED_KEY')
```

Now the middleware loads when a key is set (any truthy value) and uses that key for authentication validation.

## Changes

- **File**: `ai_platform_engineering/multi_agents/platform_engineer/protocol_bindings/a2a/main.py`
- **Line**: 114
- **Change**: Remove boolean comparison, just read the env var value

## Testing

Verified the fix with:

1. ✅ Requests without authentication → `401 Unauthorized`
2. ✅ Requests with valid Bearer token → Success
3. ✅ Requests with invalid token → `401 Unauthorized`
4. ✅ Public endpoints still accessible without auth

## Impact

- **Severity**: Critical security fix
- **Breaking Changes**: None
- **Backward Compatibility**: Fully compatible - restores intended behavior

## Related Issues

This fixes the authentication bypass that was preventing shared key authentication from working as documented.

---

**Note**: This is a simple one-line revert that restores the correct behavior from commit 821d68e9 which was accidentally broken in commit 856badd1.

Fixes #398

### DIFF
--- a/ai_platform_engineering/multi_agents/platform_engineer/protocol_bindings/a2a/main.py
+++ b/ai_platform_engineering/multi_agents/platform_engineer/protocol_bindings/a2a/main.py
@@ -111,7 +111,7 @@ app = a2a_server.build()
 ################################################################################
 A2A_AUTH_OAUTH2 = os.getenv('A2A_AUTH_OAUTH2', 'false').lower() == 'true'
 
-A2A_AUTH_SHARED_KEY = os.getenv('A2A_AUTH_SHARED_KEY', 'false').lower() == 'true'
+A2A_AUTH_SHARED_KEY = os.getenv('A2A_AUTH_SHARED_KEY')
 
 # Add CORSMiddleware to allow requests from any origin (disables CORS restrictions)
 app.add_middleware(


sriaradhyula:
@rayl15 - Thank you for the PR

sriaradhyula (approved):

PR #391, opened 2025-10-22
feat: refactor a2a_stream with common code

sriaradhyula:
## Summary

This PR refactors the a2a_stream functionality to use common code patterns across all agents, improving maintainability and consistency.

### 🏗️ **Common Infrastructure**
- **Moved common a2a functionality** from `common/` to `utils/a2a_common/` directory
- **Created BaseAgent class** with streaming support and common functionality:
  - LLM initialization and tracing setup
  - MCP client configuration  
  - Streaming response handling
  - Agent execution patterns
- **Created BaseAgentExecutor** for A2A protocol handling with streaming support
- **Added common streaming utilities** and helper functions

### 🔄 **Agent Refactoring**
- **Refactored multiple agents** to use the new common base classes (Komodor, Weather, Webex, etc.)
- **Updated imports** across the codebase to use the new utils structure
- **Maintained backward compatibility** while improving code organization

### 🐳 **Docker & CI Fixes**
- **Fixed Docker build context issues** in GitHub Actions workflows  
- **Updated all agent Dockerfiles** to use correct relative paths for the new structure
- **Resolved CI build failures** related to agent refactoring

### 🧹 **Testing & Cleanup**
- **Added comprehensive streaming test suite**
- **Resolved test suite issues** and enabled linting
- **Cleaned up unused files** and old backup code
- **Applied automatic linting fixes** across the codebase

## Benefits
- ✅ **Reduced code duplication** across agents
- ✅ **Consistent streaming patterns** for all a2a agents  
- ✅ **Easier maintenance** with centralized common functionality
- ✅ **Better testability** with shared base classes
- ✅ **Improved developer experience** with clear abstractions

## Docker Image Prebuilding
This PR uses the `prebuild/` branch prefix to trigger Docker image prebuilding, which will:
- Build and cache Docker images for all affected agents
- Speed up future builds and deployments  
- Validate Docker build fixes

## Breaking Changes
None - this maintains backward compatibility while providing the new common infrastructure.

---
**Note:** This is a continuation of the closed PR #354 with additional fixes and the prebuild branch prefix for Docker image caching.

PR #411, opened 2025-10-28
fix(a2a): Fix streaming artifact accumulation in A2ARemoteAgentConnectTool

### LINKED ISSUE #410: A2A streaming artifact accumulation broken - causing 2x duplicate agent calls
I have  been testing this and noticed something that was eating up a lot of response time. Turns out there's a bug in the A2A streaming client that's causing the supervisor to call each sub-agent **twice** for every request.

## The Problem

When I query something like \"List 5 JIRA issues\", the request takes about 45 seconds instead of the expected ~24 seconds. Looking at the logs, I saw this pattern:

```
05:47:36 - Starting A2A streaming send_message
05:47:52 - Received artifact from JIRA agent
05:47:52 - "No accumulated artifact text; falling back to non-streaming send_message"
05:48:07 - Second call to JIRA agent (duplicate!)
05:48:12 - Finally got the response
```

The supervisor is making **duplicate calls to every sub-agent** (JIRA, GitHub, AWS, ArgoCD, etc.) because it's not properly accumulating the streaming artifact text.

## Root Cause

The bug is in `ai_platform_engineering/utils/a2a/a2a_remote_agent_connect.py` around line 211.

The code does this:
```python
async for chunk in self._client.send_message_streaming(streaming_request):
    chunk_dump = chunk.model_dump(mode="json", exclude_none=True)
    
    # BUG: Trying to access .result on the Pydantic object
    result = getattr(chunk, "result", None)  # This always returns None!
    if result is None:
        continue  # Skips EVERY chunk
```

The issue is that `chunk` is a Pydantic model, and we call `chunk.model_dump()` to create a dict representation for logging. But then we try to use `getattr(chunk, "result", None)` on the original Pydantic object.

Pydantic models don't expose attributes like `.result` directly through `getattr()` the way you'd expect, so this **always returns `None`**. Every single chunk gets skipped, no artifact text is ever accumulated, and the system falls back to the non-streaming `send_message()` method - causing a second, duplicate call to the agent.

## The Fix

The fix is straightforward - work with the `chunk_dump` dict instead of the Pydantic object:

```python
async for chunk in self._client.send_message_streaming(streaming_request):
    chunk_dump = chunk.model_dump(mode="json", exclude_none=True)
    
    # FIX: Work with the dict, not the Pydantic model
    if not isinstance(chunk_dump, dict):
        continue
        
    result = chunk_dump.get("result")  # This works!
    if result is None:
        continue
    
    kind = result.get("kind")
    if kind == "artifact-update":
        art = result.get("artifact")
        if art:
            parts = art.get("parts")
            if parts:
                for part in parts:
                    if isinstance(part, dict):
                        text = part.get("text")
                        if text:
                            accumulated_text.append(text)  # Success!
```

## Impact

After applying this fix, I'm seeing:
- Artifact text properly accumulated: `"Accumulated artifact text chunk (1073 chars)"`
- No more duplicate agent calls
- Response time cut from ~45s to ~24s (**~45% faster!**)

This affects **all agents** (JIRA, GitHub, AWS, ArgoCD) and **all LLM providers** (Bedrock, OpenAI, etc.).

## Test Evidence

Here's what the logs look like after the fix:

```
2025-10-28 05:47:36 [a2a.client.tool] [INFO] Starting A2A streaming send_message.
2025-10-28 05:47:52 [a2a.client.tool] [INFO] Received A2A stream chunk: {...artifact...}
2025-10-28 05:47:52 [a2a.client.tool] [INFO] Accumulated artifact text chunk (1073 chars)
2025-10-28 05:47:57 [uvicorn.access] [INFO] "POST / HTTP/1.1" 200
```

No fallback, no duplicate calls. Just clean, fast responses.

---

I'm happy to submit a PR if you'd like! This was a fun bug to track down - those Pydantic model access patterns can be tricky sometimes. 😅

Let me know if you need any more details or have questions!

rayl15:
## Problem

The A2A remote agent connect tool currently fails to accumulate artifact text from streaming responses, which causes it to fall back to non-streaming `send_message()` for every agent call. This results in duplicate agent calls and wastes 40-45% of response time.

## Root Cause

The code attempts to use `isinstance(chunk, A2ATaskArtifactUpdateEvent)` to detect artifact updates, but the A2A protocol returns JSON-RPC wrapped responses, not direct `A2ATaskArtifactUpdateEvent` instances. This means:

1. The `isinstance()` check always fails
2. ALL streaming artifact chunks are skipped
3. `accumulated_text` remains empty
4. The fallback to non-streaming `send_message()` is triggered
5. Sub-agents (JIRA, GitHub, AWS, ArgoCD) get called twice

## Solution

This PR fixes the issue by working with the `chunk_dump` dict representation instead of trying to access attributes on the Pydantic chunk object. It properly extracts artifact text from the `result.artifact.parts` structure and handles both dict and object access patterns for maximum compatibility.

## Impact

- **Affects**: ALL agents (JIRA, GitHub, AWS, ArgoCD) and ALL LLM providers (Bedrock, OpenAI, etc.)
- **Performance**: ~50% improvement for all agent interactions (e.g., 45s → 24s for JIRA queries)
- **Scope**: Eliminates duplicate agent calls system-wide

## Testing

Verified in production deployment:
- Logs show "Accumulated artifact text chunk" messages 
- NO "falling back to non-streaming send_message" messages 
- Response times reduced by ~45% 

Fixes #410

### DIFF
--- a/ai_platform_engineering/utils/a2a/a2a_remote_agent_connect.py
+++ b/ai_platform_engineering/utils/a2a/a2a_remote_agent_connect.py
@@ -15,7 +15,6 @@ from a2a.types import (
     SendMessageRequest,
     SendStreamingMessageRequest,
     MessageSendParams,
-    TaskArtifactUpdateEvent as A2ATaskArtifactUpdateEvent,
 )
 
 from langchain_core.tools import BaseTool
@@ -207,16 +206,39 @@ class A2ARemoteAgentConnectTool(BaseTool):
             writer({"type": "a2a_event", "data": chunk_dump})
 
             try:
-                if isinstance(chunk, A2ATaskArtifactUpdateEvent):
-                    art = chunk.artifact
-                    if getattr(art, "parts", None):
-                        for part in art.parts:
-                            root = getattr(part, "root", None)
-                            text = getattr(root, "text", None) if root is not None else None
-                            if text:
-                                accumulated_text.append(text)
+              # Work with chunk_dump (dict) instead of chunk (Pydantic model)
+              # The chunk is a JSON-RPC wrapped response, not a direct A2ATaskArtifactUpdateEvent
+              if not isinstance(chunk_dump, dict):
+                continue
+
+              result = chunk_dump.get("result")
+              if result is None:
+                continue
+
+              # Check if this is an artifact-update event
+              kind = result.get("kind") if isinstance(result, dict) else None
+              if kind == "artifact-update":
+                art = result.get("artifact") if isinstance(result, dict) else None
+                if art:
+                  parts = art.get("parts") if isinstance(art, dict) else None
+                  if parts:
+                    for part in parts:
+                      text = None
+                      # Parts are dicts after model_dump()
+                      if isinstance(part, dict):
+                        text = part.get("text")
+                      else:
+                        # Fallback for object access
+                        text = getattr(part, "text", None)
+                        if not text:
+                          root = getattr(part, "root", None)
+                          if root:
+                            text = getattr(root, "text", None)
+                      if text:
+                        accumulated_text.append(text)
+                        logger.info(f"Accumulated artifact text chunk ({len(text)} chars)")
             except Exception as e:
-                logger.warning(f"Non-fatal error while handling stream chunk: {e}")
+              logger.warning(f"Non-fatal error while handling stream chunk: {e}", exc_info=True)
 
         final_response = " ".join(accumulated_text).strip()
         if not final_response:


sriaradhyula:
@rayl15 - Some of this is going to get refactored - https://github.com/cnoe-io/ai-platform-engineering/pull/391

I will merge this after #391 is merged

rayl15:
391 looks interesting. I will wait for it. Thanks @sriaradhyula

sriaradhyula:
@rayl15 - Can you please rebase changes and test with the streaming changes implemented?

rayl15:
Hi @sriaradhyula

 I rebased the branch and discovered something interesting , the fix is already present in main!

After PR #391's refactoring, the file moved to utils/a2a_common/a2a_remote_agent_connect.py and includes the same dict-based approach for handling artifact accumulation (lines 213-240). The conflicts shown are because this PR targets the old file location.

Since the issue is already resolved, feel free to close this PR. Thanks for including this fix in the refactoring work!

sriaradhyula:
## Closing — fix already on main

The streaming artifact accumulation bug described here was already fixed on `main` the day before this PR was opened.

**Commit that resolved it:** [`872b3aa`](https://github.com/cnoe-io/ai-platform-engineering/commit/872b3aa9dd1650068a9467a935bac5f0e025fa3e) — *feat: update a2a streaming and agent improvements* (Oct 27, 2025)

That commit removed the broken `TaskArtifactUpdateEvent as A2ATaskArtifactUpdateEvent` and `TaskStatusUpdateEvent as A2ATaskStatusUpdateEvent` imports that caused the `isinstance()` check to always fail, and switched to dict-based access — exactly the fix described here.

Additionally, `a2a_remote_agent_connect.py` is not in this PR's diff at all, and the PR has grown to ~1,940 changed files (mostly repo restructuring) that are now ~5 months stale against `main`. Closing to keep the PR queue clean.

PR #416, opened 2025-10-30
feat: Use more standard pattern for Helm ingress

### LINKED ISSUE #301: [Bug]: Helm ingress configuration not flexible
### Bug Description

I've installed the main Helm chart in an EKS cluster with ingress values configured the AWS Load Balancer controller installed.

The ingress values I've configured are:

```
supervisor-agent:
  ingress:
    enabled: true
    className: alb
```

The ingress that has been created has this spec:

```
spec:
  ingressClassName: alb
  rules:
  - host: supervisor-agent.local
    http:
      paths:
      - backend:
          service:
            name: ai-platform-engineering-supervisor-agent
            port:
              number: 8000
        path: /
        pathType: Prefix
```

This means that I cannot access the agent like this:

```
curl http://<load balancer DNS>
```

I have to do this:

```
curl -H 'Host: supervisor-agent.local' http://<load balancer DNS>'
```

The `host` field under `rules` in the Ingress specification is optional, but currently its being treated as required.

### Expected Behavior

The way Ingress is configured and templated can be improved to make this more flexible. An example of this can be found here:

https://github.com/aws-containers/retail-store-sample-app/blob/da39046444f93b7570a46ed4200564f6f7f11f22/src/ui/chart/templates/ingress.yaml

The `ingress.hosts` field in the values file should be changed to a array of strings which defaults to an empty array.

### Affected Version

v1.0.0

### Steps to Reproduce

See above

### Checklist

- [x] I have verified this does not duplicate an existing issue

niallthomson:
# Description

Fixes #301 

The pattern used to configure agent ingress with Helm has some issues:

1. It requires the user to specify hostnames for Ingress, which is not a required field. This means it can't be used with something like EKS and its load balancer controller using the simplest configuration.
2. It requires the user to specify the correct prefix path for ingress, which the user shouldn't need to know

This changes the way ingress is specified from this:

```
ingress:
  hosts:
    - host: agent.local
      paths:
        - path: /
          pathType: Prefix
```

To this:

```
ingress:
  hosts:
    - agent.local
```

Where `hosts` is optional.

It also changes the ingress `className` parameter to be an empty string by default, which is a more neutral default setting than `nginx`.

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [x] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/charts/ai-platform-engineering/Chart.lock
+++ b/charts/ai-platform-engineering/Chart.lock
@@ -1,40 +1,40 @@
 dependencies:
 - name: supervisor-agent
   repository: ""
-  version: 0.1.1
+  version: 0.2.0
 - name: agent
   repository: ""
-  version: 0.2.2
+  version: 0.3.0
 - name: agent
   repository: ""
-  version: 0.2.2
+  version: 0.3.0
 - name: agent
   repository: ""
-  version: 0.2.2
+  version: 0.3.0
 - name: agent
   repository: ""
-  version: 0.2.2
+  version: 0.3.0
 - name: agent
   repository: ""
-  version: 0.2.2
+  version: 0.3.0
 - name: agent
   repository: ""
-  version: 0.2.2
+  version: 0.3.0
 - name: agent
   repository: ""
-  version: 0.2.2
+  version: 0.3.0
 - name: agent
   repository: ""
-  version: 0.2.2
+  version: 0.3.0
 - name: agent
   repository: ""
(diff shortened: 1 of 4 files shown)


niallthomson:
Thanks @suwhang-cisco 

Bump agent chart to `0.3.0` and parent chart to `0.4.0`?

PR #459, opened 2025-11-11
single config file + script

sbraicu:
# Description

Single config file and script to customize and launch deployment

## Type of Change

- [ ] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

For chart changes, you can test pre-release versions before merging:
- **Base repo contributors:** Create a branch starting with `pre/` for automatic pre-release builds
- Pre-release charts are published to `ghcr.io/cnoe-io/pre-release-helm-charts`

## Checklist


niallthomson:
Transplanting my comment from the previous PR that was closed:

I like that this is reducing the size of the compose file.

I'd especially love it if we could get rid of the duplicate agents (p2p and slim). Would it be possible to simplify each agent to something like this (no transport in the service or container name, removed `profiles`):

```
agent-argocd:
    image: ghcr.io/cnoe-io/agent-argocd:${IMAGE_TAG:-stable}
    container_name: agent-argocd
    env_file: [.env]
    ports: ["8004:8000"]
    environment:
      - A2A_TRANSPORT=p2p
      - MCP_MODE=http
      - MCP_HOST=mcp-argocd
      - MCP_PORT=8000
    depends_on: [mcp-argocd]
```

Then model an "overlay" file like this (maybe called `docker-compose.slim.yaml`):

```
agent-argocd:
    environment:
      - A2A_TRANSPORT=slim
      - SLIM_ENDPOINT=${SLIM_ENDPOINT:-http://slim-dataplane:46357}
    depends_on: [mcp-argocd, slim-dataplane]
```

The `docker-compose.slim.yaml` could also contain all the slim dataplane containers, further simplifying the main compose file?

Then the script could do something like:

```
if [ "$TRANSPORT" = "slim" ]; then
    docker compose -f docker-compose.yaml -f docker-compose.slim.yaml up $SHOULD_RUN
else
    docker compose -f docker-compose.yaml up -d $SHOULD_RUN
fi
```

sbraicu:
Thanks for your comment and yes, it is a good idea for sure. 

However this would imply the refactoring of all agents and might take some time. 

In the meantime this script and simplified way of configuring & deploying what is needed would help a lot, and is probably a good intermediary stept to what you described.

docker-compose.yaml
```diff
@@ -1,1284 +1,758 @@
...
-      - ./charts/ai-platform-engineering/data/prompt_config.deep_agent.yaml:/app/prompt_config.yaml
-    profiles:
-      - p2p
-      - p2p-basic
-      - p2p-tracing
-      - rag_only
-    # The following block uses the extended 'depends_on' syntax to wait for agents to be healthy.
-    # 'condition: service_healthy' waits for health checks to pass (with start_period + retries timeout).
-    # 'condition: service_started' just waits for the container to start (faster but less reliable).
-    depends_on:
```
sriaradhyula, line 18: Can we keep these, even if it is dynamic?

niallthomson:
> However this would imply the refactoring of all agents and might take some time.

I don't understand what would need refactored. The environment variables would be the same, you're just using Docker Compose functionality to merge multiple compose files instead of using 1 huge one?

sbraicu:
> > However this would imply the refactoring of all agents and might take some time.
> 
> I don't understand what would need refactored. The environment variables would be the same, you're just using Docker Compose functionality to merge multiple compose files instead of using 1 huge one?

Yes, we will do that next. We probably should discuss the best way to move forward and keep everything still manageable - there are  > 20 docker compose profile files and a huge docker compose file in the main folder as you said. There should be a way make it better.
Until then, I hope using the script for deployment would hide some of the complexity for now.

docker-compose.yaml
sbraicu: As discussed, let's move forward and remove the "depends on" as it would break the flexibility. We should find a better way to be sure that the supervisor agent is reconciling soon enough if it comes up before the others - for now maybe a disclaimer would be also ok since docker compose deployment is meant for development only. 
And since we already use a script for deployment we should investigate ways of using that script to be sure everything comes up in the right order

PR #471, opened 2025-11-18
fix(executor): filter sub-agent completion signals and restore context_id

sriaradhyula:
## Summary

This PR fixes the issue where the supervisor was prematurely closing the event queue when sub-agents completed, preventing it from calling additional tools (e.g., Jira after PagerDuty).

## Changes

- **Filter completion signals from sub-agents**: Sub-agents no longer send `final=True` or `state=TaskState.completed` signals. The supervisor filters these out to prevent treating sub-agent completion as its own completion.
- **Remove `last_event_was_tool_result` flag**: No longer needed since we prevent the root cause (sub-agents sending completion signals) rather than detecting and ignoring them.
- **Restore `context_id` parameter**: Ensures conversation continuity across supervisor and sub-agents by passing the same `context_id` to all agents in a multi-agent conversation.
- **Simplify completion handling**: Supervisor only processes its own completion signals.
- **Update tool notification handling**: Use metadata-based detection instead of text parsing.
- **Remove unused feature flags**: `ENABLE_ARTIFACT_STREAMING`, `STREAM_SUB_AGENT_TOOL_OUTPUT`.

## Testing

- ✅ Verified that supervisor can call multiple tools sequentially (e.g., PagerDuty then Jira)
- ✅ Verified that sub-agent tool notifications are correctly forwarded to clients
- ✅ Verified that `context_id` is consistently maintained across supervisor and sub-agents

## Related Issues

Fixes issue where supervisor was prematurely closing the queue when sub-agents completed.

### DIFF
--- a/ai_platform_engineering/multi_agents/platform_engineer/prompts.py
+++ b/ai_platform_engineering/multi_agents/platform_engineer/prompts.py
@@ -16,10 +16,11 @@ def load_prompt_config(path="prompt_config.yaml"):
     return {}
 
 config = load_prompt_config()
-print("DEBUG: config keys:", list(config.keys()))
-print("DEBUG: system_prompt_template length:", len(config.get('system_prompt_template', '')))
-print("DEBUG: system_prompt_template content:")
-print(repr(config.get('system_prompt_template', '')))
+# Debug prints removed to reduce log noise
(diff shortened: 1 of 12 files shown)


sriaradhyula:
OBE

PR #646, opened 2026-01-15
fix(a2a): prevent streaming output duplication

sriaradhyula:
## Summary
Fixed duplicate content appearing in A2A streaming responses.

## Root Cause
During streaming, content was being yielded twice:
1. **Streaming phase**: Each `AIMessageChunk` token was yielded and accumulated
2. **Post-stream phase**: The final parsed response was yielded with the same accumulated content

This caused the `partial_result` artifact to contain duplicate content.

## Fix
- **agent.py**: Clear content from final response if streaming chunks were already sent
  - Check `len(accumulated_ai_content) > 1` to detect streaming mode
  - Set `final_response['content'] = ''` to avoid duplication
- **agent_executor.py**: Add `clear_accumulators` signal handling for retry/fallback scenarios
  - Skip streaming for duplicate content (exact match check)

## Testing
```bash
curl -s -X POST http://localhost:8000/ \
  -H "Content-Type: application/json" \
  -d '{"jsonrpc": "2.0", "id": "test", "method": "message/send", "params": {"message": {"messageId": "msg-1", "role": "user", "parts": [{"text": "Write Superman to /tmp/test.txt, then grep for Superman."}]}}}' | jq '.result.artifacts[-1].parts[0].text'
```

Before: Content appeared twice in response
After: Content appears once, no duplication

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/ai_platform_engineering/multi_agents/platform_engineer/protocol_bindings/a2a/agent.py
+++ b/ai_platform_engineering/multi_agents/platform_engineer/protocol_bindings/a2a/agent.py
@@ -254,7 +254,7 @@ class AIPlatformEngineerA2ABinding:
       # SYNTHESIS RETRY CONFIGURATION
       # If synthesis fails (orphaned tool calls, timeout), retry before failing
       # ========================================================================
-      max_synthesis_retries = int(os.getenv("MAX_SYNTHESIS_RETRIES", "2"))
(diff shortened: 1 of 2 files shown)


sriaradhyula:
Recreating PR from main branch

PR #777, opened 2026-02-06
feat: single-node deep agent architecture

artdroz:
## Summary

- Simplify deployment to single-node deep agent architecture with in-process MCP tool loading via STDIO
- Add task_config.yaml for deterministic self-service workflows (EC2, EKS, GitHub repos, ArgoCD deploy, etc.)
- Add PolicyMiddleware for tool call authorization (ASP-based policy engine)
- Integrate RAG knowledge base tools into single-node supervisor
- Merge isRagEnabled() functions supporting both NEXT_PUBLIC_ENABLE_RAG and NEXT_PUBLIC_RAG_ENABLED env vars
- Add local github mcp server only for single node mode

### Single-Node Deep Agent (`deep_agent_single.py`)
- Single container runs the supervisor with all MCP tools loaded in-process via STDIO transport
- No inter-container A2A communication needed — all sub-agents run as in-process tool calls
- `task_config.yaml` drives deterministic workflows for self-service operations
- `PolicyMiddleware` enforces tool authorization via `policy/policy.lp` (ASP/clorm)
- RAG tools integrated directly into the supervisor when `ENABLE_RAG=true`

### Helm Chart (`charts/ai-platform-engineering`)
- `global.deploymentMode: "single-node"` (default) vs `"multi-agent"` — controls whether sub-agents deploy as separate pods or run in-process
- In single-node mode, only the supervisor-agent pod runs; all agent sub-chart deployments are skipped
- CAIPE UI, RAG stack are optional via tags/conditions

### Docker Compose (`docker-compose.single-node.dev.yaml`)
- `caipe` service — always runs, single deep agent container (command: `platform-engineer-single`)
- `caipe-ui` — optional profile
- RAG services — optional profile: `rag_server`, `web_ingestor`, `rag-redis`, `milvus-standalone`, `etcd`, `milvus-minio`
- Graph RAG — optional profile: adds `neo4j`, `agent_ontology`, `backstage_ingestor`
- Tracing — optional profile: Langfuse stack

### CI Fixes
- Fix duplicate `ragEnabled` property/function in `ui/src/lib/config.ts` and `ui/src/hooks/use-rag-health.ts`
- Fix 33 ruff lint errors (unused imports, unused variables, f-strings without placeholders)

## Single-Node Architecture

In single-node mode, there are **no separate agent containers**. The architecture is:

```
┌──────────────────────────────────────────┐
│     supervisor-agent pod                 │
│     (platform-engineer-single)           │
│                                          │
│  ┌─ ArgoCD MCP tools (in-process)        │
│  ├─ GitHub MCP tools (in-process)        │
│  ├─ Jira MCP tools (in-process)          │
│  ├─ AWS MCP tools (in-process)           │
│  ├─ Slack MCP tools (in-process)         │
│  ├─ ... all other MCP tools              │
│  └─ RAG tools (if ENABLE_RAG=true)       │
└──────────────────────────────────────────┘
         ↑
         │ HTTP :8000
         ↓
┌──────────────────────────────────────────┐
│     caipe-ui pod (optional)              │
│     NEXT_PUBLIC_A2A_BASE_URL → :8000     │
└──────────────────────────────────────────┘

┌──────────────────────────────────────────┐
│     RAG infra (optional, via rag-stack)  │
│     rag-server, rag-redis, milvus        │
└──────────────────────────────────────────┘
```

### Helm Chart
```bash
helm pull oci://ghcr.io/cnoe-io/pre-release-helm-charts/ai-platform-engineering \
  --version 0.2.16-single-node-v5-7
```

### Docker Images
Only **2 application images** are needed for single-node:

| Component | Image |
|-----------|-------|
| Supervisor (single deep agent) | `ghcr.io/cnoe-io/prebuild/ai-platform-engineering:single-node-v5-8` |
| CAIPE UI | `ghcr.io/cnoe-io/prebuild/caipe-ui:single-node-v5-8` |

> **Note:** Individual agent images (agent-argocd, agent-github, etc.) are **not used** in single-node mode. All MCP tools are loaded in-process inside the supervisor container.

## Test plan
- [x] Ruff linting passes (`make lint`)
- [x] Next.js UI build succeeds (`npm run build` in ui/)
- [x] Pre-release Helm chart publishes successfully
- [x] Pre-release Docker images build successfully
- [x] Helm install on cluster with single-node mode
- [x] Verify supervisor starts with `platform-engineer-single` and loads MCP tools
- [x] Verify CAIPE UI connects to supervisor
- [x] Verify RAG services start and connect (if enabled)

### DIFF
--- a/charts/ai-platform-engineering/Chart.yaml
+++ b/charts/ai-platform-engineering/Chart.yaml
@@ -1,20 +1,20 @@
 apiVersion: v2
 # pyproject.toml version
-appVersion: 0.2.32 # Do NOT modify. It will be updated automatically using the release workflow
+appVersion: 0.2.31 # Do NOT modify. It will be updated automatically using the release workflow
 name: ai-platform-engineering
 description: Parent chart to deploy multiple agent subcharts as different platform agents
 sources:
   - https://github.com/cnoe-io/ai-platform-engineering/charts
 # Chart version for ai-platform-engineering parent chart
-version: 0.2.32 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+version: 0.2.31 # Do NOT bump this. It will be updated automatically using the PR or release workflow
 dependencies:
   # AI Platform Engineer Multi-Agent
   - name: supervisor-agent
     # Chart version for agent subchart
-    version: 0.2.32 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.2.31 # Do NOT bump this. It will be updated automatically using the PR or release workflow
   # Single agent chart used multiple times with different aliases
   - name: agent
-    version: 0.2.32 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.2.31 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     alias: agent-argocd
     tags:
       - agent-argocd
@@ -24,7 +24,7 @@ dependencies:
       - child: agentExports.data
         parent: global.enabledSubAgents.argocd
   - name: agent
-    version: 0.2.32 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.2.31 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     alias: agent-aws
     tags:
       - agent-aws
@@ -33,7 +33,7 @@ dependencies:
       - child: agentExports.data
         parent: global.enabledSubAgents.aws
   - name: agent
-    version: 0.2.32 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.2.31 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     alias: agent-backstage
     tags:
       - agent-backstage
@@ -43,7 +43,7 @@ dependencies:
       - child: agentExports.data
         parent: global.enabledSubAgents.backstage
   - name: agent
-    version: 0.2.32 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.2.31 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     alias: agent-confluence
     tags:
       - agent-confluence
@@ -52,7 +52,7 @@ dependencies:
       - child: agentExports.data
         parent: global.enabledSubAgents.confluence
   - name: agent
-    version: 0.2.32 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.2.31 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     alias: agent-github
     tags:
       - agent-github
@@ -62,7 +62,7 @@ dependencies:
       - child: agentExports.data
         parent: global.enabledSubAgents.github
   - name: agent
-    version: 0.2.32 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.2.31 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     alias: agent-gitlab
     tags:
       - agent-gitlab
@@ -71,7 +71,7 @@ dependencies:
       - child: agentExports.data
         parent: global.enabledSubAgents.gitlab
   - name: agent
-    version: 0.2.32 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.2.31 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     alias: agent-jira
     tags:
       - agent-jira
@@ -80,7 +80,7 @@ dependencies:
       - child: agentExports.data
         parent: global.enabledSubAgents.jira
   - name: agent
-    version: 0.2.32 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.2.31 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     alias: agent-komodor
     tags:
       - agent-komodor
@@ -89,7 +89,7 @@ dependencies:
       - child: agentExports.data
         parent: global.enabledSubAgents.komodor
   - name: agent
-    version: 0.2.32 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.2.31 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     alias: agent-pagerduty
     tags:
       - agent-pagerduty
@@ -98,7 +98,7 @@ dependencies:
       - child: agentExports.data
         parent: global.enabledSubAgents.pagerduty
   - name: agent
-    version: 0.2.32 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.2.31 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     alias: agent-slack
     tags:
       - agent-slack
@@ -107,7 +107,7 @@ dependencies:
       - child: agentExports.data
         parent: global.enabledSubAgents.slack
   - name: agent
-    version: 0.2.32 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.2.31 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     alias: agent-splunk
     tags:
       - agent-splunk
@@ -116,7 +116,7 @@ dependencies:
       - child: agentExports.data
         parent: global.enabledSubAgents.splunk
   - name: agent
-    version: 0.2.32 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.2.31 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     alias: agent-victorops
     tags:
       - agent-victorops
@@ -124,7 +124,7 @@ dependencies:
       - child: agentExports.data
         parent: global.enabledSubAgents.victorops
   - name: agent
-    version: 0.2.32 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.2.31 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     alias: agent-webex
     tags:
       - agent-webex
@@ -133,7 +133,7 @@ dependencies:
       - child: agentExports.data
         parent: global.enabledSubAgents.webex
   - name: agent
-    version: 0.2.32 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.2.31 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     alias: agent-netutils
     tags:
       - agent-netutils
@@ -142,7 +142,7 @@ dependencies:
       - child: agentExports.data
         parent: global.enabledSubAgents.netutils
   - name: agent
-    version: 0.2.32 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.2.31 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     alias: agent-weather
     tags:
       - agent-weather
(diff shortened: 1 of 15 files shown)


artdroz:
## Backwards Compatibility Analysis — Files Changed vs Main

This comment enumerates every file modified on `main` (not net-new) and explains how each change preserves backwards compatibility.

---

### 1. `agent.py` (multi-node executor binding)

**Changes:** Reads `user_email` from `self._pending_user_email` attribute (set by executor) and adds it to `config['metadata']`. The `stream()` method signature is **unchanged** (`query, context_id, trace_id=None`). The `@trace_agent_stream` decorator is untouched.

**Backwards compat:** No signature change. `_pending_user_email` defaults to `None` via `getattr`, so when the attribute is not set (e.g. in main's executor), the code silently skips it. Adding a key to `config['metadata']` is additive — downstream consumers ignore unknown keys.

---

### 2. `agent_executor.py` (multi-node executor)

**Changes:**
- `_parse_execution_plan_text`: Extended from 1 pattern (markdown checkbox `- [x]`) to 3 patterns (emoji+agent, bullet+emoji, markdown checkbox). The **original checkbox pattern is retained as pattern 3** — if no emoji patterns match, the original regex runs.
- `_format_execution_plan_text`: Changed output from `- [x] step` to `- ✅ step`. This is a format change but only affects how the executor emits plan updates, not how it parses them (the parser handles both).
- `execute()`: Extracts `user_email` from the `"by user: email\n\n..."` prefix in the query text. Sets `self.agent._pending_user_email` before calling `stream()`. If the prefix is absent, `user_email` is `None` and the attribute is never set — no change in behavior.

**Backwards compat:** Parser is strictly additive (old pattern still works). Formatter output changed but the UI parser handles both formats. `user_email` extraction is no-op when prefix is absent. The `stream()` call is unchanged.

---

### 3. `a2a_remote_agent_connect.py`

**Changes:**
- Agent card field access: Uses `getattr()` with fallback for `supports_authenticated_extended_card` / `supportsAuthenticatedExtendedCard` to handle both Pydantic v1 and v2 field naming.
- URL override: After fetching agent card, overrides `self._agent_card.url` with the base URL used to reach it (agents often advertise `0.0.0.0` which doesn't work cross-container).
- User email propagation: Prepends `"by user: email\n\n"` to message text and adds `user_email` to metadata dict. If `user_email` is `None`, no prefix is added and metadata key is skipped.

**Backwards compat:** `getattr` fallback is strictly more robust. URL override only fires when the advertised URL differs from the connection URL. User email prefix is conditional on `user_email` being set (it's an existing parameter that was previously unused in message construction).

---

### 4. `base_langgraph_agent.py`

**Changes:** Adds `_load_mcp_tools()` method and `_load_mcp_tools_with_cleanup_handling()` static method. These are new methods extracted from `_setup_mcp_and_graph` to allow reuse for single-node subagent graph creation.

**Backwards compat:** Purely additive — new methods on the base class. No existing methods modified. Subclasses that don't call these methods are unaffected. The `include_fallback` parameter defaults to `True` preserving original behavior.

---

### 5. `ContextPanel.tsx`

**Changes:**
- `parseExecutionTasks()`: Extended from 1 pattern (emoji+agent `⏳ [ArgoCD] ...`) to 3 patterns (emoji+agent, bullet+emoji, markdown checkbox). Patterns are tried in order with fallthrough — **Pattern 1 is identical to main's single pattern**.
- `emojiToStatus()`: Extracted as a helper. Fixes a minor visual issue in main where `⏳` mapped to `in_progress` instead of `pending`. On main, `⏳` and `🔄` were grouped together and both mapped to `in_progress`. Since `⏳` semantically means "waiting/not started," they are now separated: `⏳` → `pending`, `🔄` → `in_progress`. In practice the impact is minimal since plan items quickly transition through statuses.
- HITL completion guard: `hasFinalResult` memo added; tasks are only force-marked as completed when `final_result` artifact is received, not when the stream pauses for HITL.
- CSS overflow fixes: Added `overflow-hidden`, `min-w-0`, `shrink-0`, `line-clamp-2`, `title` attributes to prevent long task descriptions from breaking the layout.

**Backwards compat:** Pattern 1 (single-node emoji+agent format) is preserved as the primary parser path. New patterns only activate if pattern 1 matches nothing. CSS changes are visual-only. HITL guard is additive.

---

### 6. `a2a-sdk-client.ts`

**Changes:**
- Added `HITLInputField`, `HITLFormData`, `HITLDecision` interfaces (new exports).
- `sendMessageStream()`: Added optional `metadata` parameter (defaults to `undefined`). When provided, appends a `DataPart` to the message. Existing callers pass no metadata — no behavior change.
- Added `sendHITLResponse()`: New method that formats form decisions as DataPart metadata and sends via `sendMessageStream`. The message text is constructed from actual `user_inputs` values (not hardcoded `"Form submitted"`).

**Backwards compat:** New interfaces are additive exports. `sendMessageStream` signature change uses optional parameter with default — all existing call sites work unchanged. `sendHITLResponse` is new and only called from the new HITL flow in `ChatPanel`.

---

### 7. `api-middleware.ts`

**Changes:**
- Added `import { getConfig } from '@/lib/config'`.
- `getAuthenticatedUser()`: When SSO is disabled (`getConfig('ssoEnabled')` returns false) and no session exists, returns a fallback anonymous user instead of throwing 401. This allows MongoDB persistence to work in local dev without SSO.
- `validateRequired()`: Changed from `!data[field]` (falsy check) to `data[field] === undefined || data[field] === null` (nullish check). This fixes a bug where `false`, `0`, or `""` were treated as missing.

**Backwards compat:** When SSO is enabled (production), the fallback path is never reached — `getConfig('ssoEnabled')` returns true and the existing 401 logic runs. The `validateRequired` fix is a bugfix that makes validation more correct.

---

### 8. `ChatPanel.tsx`

**Changes:**
- HITL form state: Added `contextId` to `pendingUserInput` state. Added `hitlFormRequested` flag to distinguish HITL stream pause from normal completion.
- `handleUserInputSubmit`: Rewritten to use `A2ASDKClient.sendHITLResponse()` which sends form values as both message text and DataPart metadata, then streams the agent's continuation.
- HITL restoration: Added position-based check to avoid re-showing form after user already answered. Added `isFinal` check to skip restoration on completed workflows.
- Input dismissal: Clears pending form when user types directly.

**Backwards compat:** All HITL changes are additive. Non-HITL message flow (`submitMessage`) is unchanged. The `handleUserInputSubmit` replacement calls the same underlying `sendMessageStream` with additional metadata — the agent receives the same message text plus optional DataPart.

---

### 9. `MetadataInputForm.tsx`

**Changes:**
- Added `response?: string` field to `UserInputMetadata` interface.
- Field type detection: Changed from `field.field_type || (auto-detect)` to checking `field.field_values` directly. If `field_values` is present, renders a select regardless of `field_type`.

**Backwards compat:** The `response` field is optional — existing code that doesn't set it is unaffected. The field type detection change makes rendering more reliable (always shows dropdown when options exist) without breaking any existing form definitions.

---

### 10. `chat-store.ts`

**Changes:** When loading messages from MongoDB, checks if any event has `artifact.name === 'UserInputMetaData'`. If so, `isInterrupted` is set to `false` instead of being derived from `!isFinal`. This prevents HITL messages from showing as "interrupted."

**Backwards compat:** Only affects messages with HITL events. Non-HITL messages retain the original `isInterrupted` logic.

---

### 11. `agent-config-store.ts`

**Changes:** Added `importFromYaml()` method and renamed "Agent Skills" to "Agentic Workflows" in comments.

**Backwards compat:** `importFromYaml` is a new method. Comment-only renames don't affect runtime. No existing methods modified.

---

### 12. `docker-compose.dev.yaml`

**Changes:** Fixed SSO env var names from `CAIPE_UI_OIDC_ENABLED`/`CAIPE_UI_OIDC_*` to `SSO_ENABLED`/`OIDC_*` to match what the Next.js app actually reads via `lib/config.ts`. Added `OIDC_GROUP_CLAIM`, `OIDC_REQUIRED_GROUP`, `OIDC_REQUIRED_ADMIN_GROUP`, `OIDC_ENABLE_REFRESH_TOKEN`.

**Backwards compat:** This is a **bugfix** — the old `CAIPE_UI_OIDC_*` names were incorrect and SSO was silently disabled. The app's `lib/config.ts` reads `SSO_ENABLED` and `OIDC_*` directly. New vars are additive with defaults.

---

### 13. Sub-agent files (backstage, confluence, github, jira, webex agents)

**Changes:** Various MCP configuration improvements:
- **Backstage/Jira**: Fixed `project_path` (was pointing to `mcp_backstage/` instead of `mcp/`). Added missing `ATLASSIAN_EMAIL` env var.
- **GitHub**: Removed Copilot HTTP MCP (`get_mcp_http_config()` now returns `None`). Added `get_mcp_config()` with go-based `github-mcp-server` via STDIO transport. **In multi-node (`MCP_MODE=http`), the base class never reaches `get_mcp_config()` — it probes localhost HTTP, finds no server, and falls back to `get_additional_tools()` which provides gh CLI, matching main's effective behavior.** In single-node (`MCP_MODE=stdio`), `get_mcp_config()` is called and provides full MCP tool coverage via the local go server. Import path changed from relative (`from agent_github.tools`) to absolute.
- **Webex**: Fixed `server_path` default and MCP config structure (was returning nested dict, now flat).
- **Confluence**: Added missing `ATLASSIAN_EMAIL` env var.

**Backwards compat:** Multi-node GitHub behavior is preserved: gh CLI remains the primary tool. The Copilot HTTP MCP config is removed but was not the effective tool source — gh CLI (via `get_additional_tools()`) was the working tool. Single-node adds new STDIO MCP as an additive capability. Other sub-agent fixes correct broken MCP tool loading. Agent APIs (`get_agent_name`, `get_system_instruction`, etc.) are unchanged.

---

### 14. Build files (`Dockerfile`, `Dockerfile.caipe-ui`, `entrypoint.sh`)

**Changes:**
- **Dockerfile**: Added Go toolchain stage for github-mcp-server. Added `uv`, `terraform`, `unzip`. Extended `PYTHONPATH` for single-node imports. Added `GOPATH`.
- **Dockerfile.caipe-ui**: Pinned build stages to `BUILDPLATFORM` to avoid QEMU crashes. Changed `public/` ownership for entrypoint script. Updated comments.
- **entrypoint.sh**: Generates `env-config.js` at startup with `NEXT_PUBLIC_*` vars (was previously just logging them).

**Backwards compat:** Dockerfile additions are purely additive — multi-node containers work identically with the extra tools present but unused. `BUILDPLATFORM` pinning is a build-time fix that produces the same runtime image. `entrypoint.sh` change generates an additional static file that the app optionally loads.

---

### 15. Infrastructure files (Helm charts, CI, pyproject.toml, Makefile, etc.)

- **Helm values/templates**: Added `deploymentMode` conditional to skip agent deployments in single-node mode. Added `task-config` volume mount. Added `agent-weather` config block. These are gated by `global.deploymentMode` defaulting to `"multi-agent"` so existing deployments are unaffected.
- **helm-pre-release.yml**: Strip pre-release suffix from base version before computing next version. Bugfix.
- **pyproject.toml**: Added `deepagents>=0.3.8`, `nest-asyncio`, `clorm` dependencies. Relaxed `langchain`/`langchain-core` version pins. Additive.
- **Makefile**: Added `run-a2a`, `run-a2a-client`, `uv-prep` targets. Additive.
- **deepagents/state.py**: Added `id` field to `Todo` TypedDict. Required by deepagents library.
- **.gitignore**: Added `task_config.yaml` exception for charts data dir.

**Backwards compat:** All additive or gated behind `deploymentMode`. Existing multi-node Helm deployments, CI pipelines, and dependency resolution are unaffected.

---

### Summary

All modified files preserve backwards compatibility through:
1. **Additive patterns** — new functionality alongside existing code paths
2. **Optional parameters** — new params default to `None`/`undefined`
3. **Conditional guards** — new behavior gated on config values (`deploymentMode`, `ssoEnabled`, `MCP_MODE`)
4. **Bugfixes** — correcting broken behavior (SSO env vars, field validation)
5. **Unchanged signatures** — public APIs and method signatures are preserved
6. **Mode-aware tooling** — multi-node uses gh CLI (matching main), single-node adds MCP STDIO as new capability

PR #784, opened 2026-02-07
fix(ui): force dynamic rendering for runtime env injection

sriaradhyula:
## Summary

- **Root cause**: PR #779 introduced `PublicEnvScript` as a Server Component to inject `NEXT_PUBLIC_*` env vars into `window.__RUNTIME_ENV__` at request time. However, the root layout was **statically rendered at build time**, so `PublicEnvScript` captured an empty `process.env` and injected `window.__RUNTIME_ENV__={}` into the HTML. This broke SSO, branding, and all other runtime config in production (Docker/K8s) where env vars are only available at runtime, not build time.

- **Fix**: Call `await headers()` in `RootLayout` to force dynamic rendering, ensuring `PublicEnvScript` reads `process.env` at every request. Also convert static `metadata` export to `generateMetadata()` so the page title reflects runtime branding (e.g. the configured app name instead of the default).

## Test plan

- [ ] Verify `window.__RUNTIME_ENV__` contains all `NEXT_PUBLIC_*` vars (browser console)
- [ ] Verify SSO login/logout flow works in production deployment
- [ ] Verify `UserMenu` appears in the top-right after login
- [ ] Verify page title reflects runtime `NEXT_PUBLIC_APP_NAME` value
- [ ] Verify branding (logo, tagline, spinner color) loads correctly from runtime env

### DIFF
--- a/ui/src/app/layout.tsx
+++ b/ui/src/app/layout.tsx
@@ -1,4 +1,5 @@
 import type { Metadata } from "next";
+import { headers } from "next/headers";
 import { Inter, JetBrains_Mono, Source_Sans_3, IBM_Plex_Sans } from "next/font/google";
 import { ThemeProvider } from "@/components/theme-provider";
 import { AuthProvider } from "@/components/auth-provider";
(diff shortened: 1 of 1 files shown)


sriaradhyula:
Closing: fully superseded by #783 which includes the same layout.tsx dynamic rendering fix plus the complete config architecture overhaul.

PR #785, opened 2026-02-08
fix(ui): sync conversation ID between client and MongoDB to fix share dialog

sriaradhyula:
## Summary

- **Bug**: Share dialog showed "Legacy Conversation" on a **brand new database** with MongoDB enabled. Every conversation appeared as legacy, making sharing impossible.
- **Root cause**: `createConversation()` in the Zustand store generated a UUID on the client, but `POST /api/chat/conversations` generated a **different** UUID server-side. The client stored conversations under ID `abc-123` while MongoDB stored them under `xyz-789`. When the Share dialog fetched `/api/chat/conversations/abc-123/share`, MongoDB returned 404, incorrectly triggering the legacy fallback.
- **Fix**: Pass the client-generated UUID to the server API so both sides use the same ID. Also improved ShareDialog error handling to distinguish between "conversation not yet synced" (MongoDB mode 404) vs "genuinely legacy" (localStorage mode).

## Changes

| File | Change |
|------|--------|
| `ui/src/types/mongodb.ts` | Added optional `id` field to `CreateConversationRequest` |
| `ui/src/app/api/chat/conversations/route.ts` | Use client-provided `id` if present, otherwise generate UUID |
| `ui/src/store/chat-store.ts` | Pass the locally generated ID to `apiClient.createConversation()` |
| `ui/src/components/chat/ShareDialog.tsx` | Don't treat 404 as legacy when MongoDB is enabled; don't treat network errors as legacy |

## Test plan

- [ ] Create a new conversation on a fresh MongoDB database
- [ ] Click Share on the new conversation — should show share UI, not "Legacy Conversation"
- [ ] Verify conversation ID in browser matches MongoDB `_id` field
- [ ] In localStorage-only mode, legacy message should still appear for unsaved conversations
- [ ] Network errors during share load should not show legacy message

### DIFF
--- a/ui/src/app/api/chat/conversations/route.ts
+++ b/ui/src/app/api/chat/conversations/route.ts
@@ -90,7 +90,7 @@ export const POST = withErrorHandler(async (request: NextRequest) => {
 
     const now = new Date();
     const newConversation: Conversation = {
-      _id: uuidv4(),
+      _id: body.id || uuidv4(), // Use client-provided ID if given, otherwise generate
       title: body.title,
       owner_id: user.email,
       created_at: now,
--- a/ui/src/components/chat/ShareDialog.tsx
+++ b/ui/src/components/chat/ShareDialog.tsx
@@ -84,12 +84,23 @@ export function ShareDialog({
           }
         }
       } else if (response.status === 404) {
-        // Conversation doesn't exist in MongoDB (legacy local conversation)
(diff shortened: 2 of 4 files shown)


sriaradhyula:
Moved to PR #783 (prebuild/runtime-config-and-ui-improvements branch) instead.

PR #827, opened 2026-02-19
fix(webloader): fix sitemap discovery for subdirectory paths

kevkantes:
## Summary
Fixes sitemap crawling for sites that have sitemaps in subdirectories rather than at the domain root.

## Problem
The sitemap crawler was constructing sitemap URLs by extracting only the domain from the start URL, ignoring any path component. For example:
- Start URL: `https://example.com/docs/`
- Attempted: `https://example.com/sitemap.xml` (may redirect to auth page)
- Should try: `https://example.com/docs/sitemap.xml` (correct location)

When `/sitemap.xml` at the root redirected to an authentication page, the crawler would:
1. Parse the HTML login page as XML
2. Find 0 URLs in the "sitemap"
3. Complete with 0 pages crawled
4. Provide no clear error message to the user

## Solution
Changed sitemap URL construction from:
```python
parsed = urlparse(self.start_url)
base_url = f"{parsed.scheme}://{parsed.netloc}"  # Domain only
```

To:
```python
base_url = self.start_url.rstrip("/")  # Full path preserved
```

This preserves the path component when constructing both `/sitemap.xml` and `/robots.txt` URLs, allowing sitemaps to be discovered in subdirectories.

## Testing
- ✅ All existing tests pass (31 passed, 3 skipped)
- ✅ Manually tested with subdirectory URLs - now finds and crawls pages successfully
- ✅ Verified backward compatibility with root-level sitemaps

## Impact
- **Backward compatible**: Sites with root-level sitemaps continue to work
- **New functionality**: Sites with subdirectory sitemaps now work correctly
- **Better error handling**: Failures are now properly detected rather than silent 0-page completions

### DIFF
--- a/ai_platform_engineering/knowledge_bases/rag/ingestors/src/ingestors/webloader/loader/scrapy_worker.py
+++ b/ai_platform_engineering/knowledge_bases/rag/ingestors/src/ingestors/webloader/loader/scrapy_worker.py
@@ -195,8 +195,13 @@ class WorkerSpider(Spider):
       # Determine if we have a subdirectory path
       has_subdirectory = parsed.path and parsed.path.rstrip("/") != ""
 
-      # Try subdirectory sitemap.xml first (if there's a path)
-      sitemap_url = f"{subdirectory_base}/sitemap.xml"
+      # Check if user provided a direct sitemap URL
+      if self.start_url.endswith("sitemap.xml") or self.start_url.endswith("sitemap.xml.gz"):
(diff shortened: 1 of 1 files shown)


subbaksh:
Also fixed: the code would silently fall back to crawling just the start URL if sitemap discovery failed for sub directory. Now it tries subdirectory then root for both sitemap.xml and robots.txt, and fails with a clear error listing all URLs it checked

PR #836, opened 2026-02-23
feat: make skill-templates independent of caipe-ui.enabled

sriaradhyula:
## Problem
The skill-templates ConfigMap is tied to caipe-ui.enabled, creating unnecessary coupling and making configuration unclear.

## Solution
- Remove dependency on `caipe-ui.enabled` flag
- Use only `caipe-ui.skills.enabled` (defaults to `true`)
- Skills can now be deployed independently of caipe-ui state
- Simplifies configuration and makes intent clearer

## Changes
- Modified condition in skill-templates-configmap.yaml to check only caipe-ui.skills.enabled
- Default value of true maintains backward compatibility
- Skills are now independently controllable

## Breaking Changes
None - defaults to true for backward compatibility

### DIFF
--- a/charts/ai-platform-engineering/values.yaml
+++ b/charts/ai-platform-engineering/values.yaml
@@ -362,6 +362,8 @@ backstage-plugin-agent-forge:
 
 ######### CAIPE UI configuration #########
 caipe-ui:
+  skills:
+    enabled: true
   nameOverride: "caipe-ui"
   image:
     repository: "ghcr.io/cnoe-io/caipe-ui"


charts/ai-platform-engineering/templates/skill-templates-configmap.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if index .Values "caipe-ui" "enabled" | default false }}
+{{- if index .Values "caipe-ui" "skills" "enabled" | default true }}
```
suwhang-cisco, line 1: Any reason why the default is now `true`? Also could we add this new value field in the values.yaml with the desired default for clarity? I'm also wondering if this cm shouldn't be here but under caipe-ui chart as this is specific to caipe-ui, but we can keep it here for now.

sriaradhyula: We need this value to be default

PR #888, opened 2026-03-03
feat: add custom mention prompts

cisco-erilutz:
## Context

Adds custom mention prompts to the channel configs so that mention behavior can be tuned


kevkantes:
We need to add this to the helm chart as well like the other prompts so we can configure this on the fly

sriaradhyula (approved):

PR #890, opened 2026-03-03
feat(supervisor-agent): add per-client supervisor customization via clientProfiles

kevkantes:
## Summary

- **DISABLED_AGENTS blocklist (Python):** New env var that filters agents out of the supervisor's enabled list, applied after the existing ENABLE_* gate
- **clientProfiles Helm value:** A map where each entry auto-generates an additional supervisor Deployment + Service with its own OAUTH2_CLIENT_ID and DISABLED_AGENTS
- **Shared Helm partials:** Deployment and Service specs extracted into _deployment.tpl and _service.tpl so default and profile instances render from the same template
- **Duplicate client ID validation:** Helm fails with a clear error if a client ID appears in both the default OAUTH2_CLIENT_ID and a profile
- **NOTES.txt:** Prints routing setup instructions (e.g. Istio JWT claim-based routing) when profiles are configured

### DIFF
--- a/charts/ai-platform-engineering/Chart.yaml
+++ b/charts/ai-platform-engineering/Chart.yaml
@@ -6,7 +6,7 @@ description: Parent chart to deploy multiple agent subcharts as different platfo
 sources:
   - https://github.com/cnoe-io/ai-platform-engineering/charts
 # Chart version for ai-platform-engineering parent chart
-version: 0.2.33-rc.helm.4 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+version: 0.2.33-rc.helm.5 # Do NOT bump this. It will be updated automatically using the PR or release workflow
 dependencies:
   # AI Platform Engineer Multi-Agent
   - name: supervisor-agent


charts/ai-platform-engineering/charts/supervisor-agent/templates/NOTES.txt
```diff
@@ -0,0 +1,39 @@
...
+
+{{- range $name, $profile := .Values.clientProfiles }}
+  - {{ include "supervisorAgent.fullname" $ }}-{{ $name }}
+    Client ID : {{ $profile.clientId }}
+    Disabled  : {{ $profile.disabledAgents | default "none" }}
+{{- end }}
+
+ACTION REQUIRED: Configure your ingress or service mesh to route
+traffic to each profile's Service based on the OAuth2 client ID.
+For Istio, use RequestAuthentication to extract the JWT "cid" claim
```
cisco-erilutz: I don't think we're allowed to create these with our setup. I tried when setting up temp auth, and either this or something like this wasn't allowed since it's governed at a higher level. We should definitely confirm that we could actually create this

kevkantes: Okay yea, confirmed we can't. However I do think this is the "right" way to do it for others

We can use envoyfilter for our use case to extract the cid and route it to the right place. The supervisor is still doing JWT validation, so I think we should be okay

sriaradhyula:
Over come new changes with custom agents

PR #898, opened 2026-03-04
feat(aigateway): add Vault integration and single-node agent configuration

artdroz:
## Summary

- **Vault integration for AIGateway**: Add `vault_utils.py` with HashiCorp Vault AppRole authentication and KV v2 operations. LiteLLM API keys are unconditionally stored/retrieved from Vault whenever credentials are available (no feature flag). Includes `rotate_llm_api_key` tool for key rotation with model/budget preservation.
- **Single-node agent configuration**: Support `ENABLE_<AGENT>` env vars for agent opt-in/out and `SUBAGENT_<NAME>_MODEL` for per-subagent LLM model overrides in `deep_agent_single.py`. Add `single-node-agent-env` ConfigMap template to generate these from Helm chart values.
- **Helm chart updates**: Rename deployment mode from `multi-agent` to `multi-node`. Add `agentgateway-crds` and `agentgateway` chart dependencies (gated behind `global.agentgateway.enabled`). Externalize AIGateway system prompt to `prompt_config.aigateway_agent.yaml`. Mount sub-agent prompt ConfigMaps in supervisor deployment for single-node mode. Document Vault env vars as part of `llmSecrets` (single secret pattern matching the dev deployment).
- **Docker Compose single-node dev fixes**: Forward `OPENAI_API_VERSION`, add Vault env vars, add MongoDB persistence for caipe-ui, fix `NEXTAUTH_URL` and port mapping.

### Files changed (19 files, +1509 / -597)

| Area | Files |
|------|-------|
| AIGateway agent | `agent.py`, `tools.py`, `vault_utils.py` (new) |
| Single-node orchestrator | `deep_agent_single.py` |
| Helm chart | `Chart.yaml`, `values.yaml`, `single-node-agent-env.yaml` (new), `supervisor-agent-env.yaml`, `supervisor-agent/deployment.yaml`, agent subchart templates |
| Prompt config | `data/prompt_config.aigateway_agent.yaml` (new) |
| Task config | `data/task_config.yaml` |
| Docker Compose | `docker-compose.single-node.dev.yaml` |
| Dependencies | `pyproject.toml`, `uv.lock` |

## Test plan

- [x] Verified `create_llm_api_key` stores key in Vault via AppRole auth (logs confirm full store/retrieve cycle)
- [ ] Verify `rotate_llm_api_key` rotates key and preserves model/budget in Vault
- [ ] Verify `ENABLE_<AGENT>=false` disables agent in single-node mode
- [ ] Verify `SUBAGENT_<NAME>_MODEL` overrides LLM model per subagent
- [ ] `helm template` renders correctly for single-node and multi-node modes
- [ ] Existing multi-node deployment unaffected


ai_platform_engineering/agents/aigateway/agent_aigateway/tools.py
```diff
@@ -1,607 +1,952 @@
...
+    msg += f"curl -X POST '{LITELLM_API_URL}/v1/embeddings' \\\n"
+    msg += f"  -H 'Authorization: Bearer {api_key}' \\\n"
+    msg += "  -H 'Content-Type: application/json' \\\n"
+    msg += f"  -d '{{\"model\": \"{full_model_name}\", \"input\": [\"test\"]}}'\n"
+    msg += "```\n\n"
+    msg += "**Python:**\n```python\n"
+    msg += "import openai\n"
+    msg += f"client = openai.OpenAI(api_key=\"{api_key}\", base_url=\"{LITELLM_API_URL}\")\n"
+    msg += f"response = client.embeddings.create(model=\"{full_model_name}\", input=[\"test\"])\n"
+    msg += "```\n\n"
```
suwhang-cisco: nit picking but can we also put this in a function similar to `_format_chat_code_examples` for code consistency? I also wonder if this is something that should also go to a config file so a user can easily modify, but can just stay here for now

charts/ai-platform-engineering/charts/supervisor-agent/templates/deployment.yaml
```diff
@@ -156,6 +158,16 @@ spec:
               mountPath: /app/task_config.yaml
               subPath: task_config.yaml
               readOnly: true
+            {{- if eq (default "single-node" .Values.global.deploymentMode) "single-node" }}
```
suwhang-cisco: Can we ensure all defaults are "multi-node" for now for consistency, or even better get rid of the default completely from everywhere?

charts/ai-platform-engineering/values.yaml
```diff
@@ -25,6 +25,64 @@ tags:
...
+    confluence: true
+    github: true
+    jira: true
+    komodor: true
+    pagerduty: true
+    rag: false
+    slack: true
+    splunk: true
+    weather: false
+    webex: true
```
suwhang-cisco: Could we put this under:
```
singleNodeConfigs:
```
or something similar instead and also get it out of `global:`? global variable is only required when you need to pass down the values to the subagents, but in single-node case, you don't have any subagent charts (except remote agents but that already has `.Values.remoteAgent` field within that subchart).

That way it clearer to the user that these variables only get used in single-agent case. Furthermore, `enabledSubAgents` currently might clash when using `multi-node` as this is auto imported from subcharts currently

charts/ai-platform-engineering/values.yaml
```diff
@@ -25,6 +25,64 @@ tags:
...
+  # All Vault env vars should be added to llmSecrets (either .data for createLlmSecret
+  # or .externalSecrets.data for ExternalSecrets), matching the pattern used for all
+  # other env vars. Required env vars:
+  #   VAULT_ADDR           - e.g. "https://keeper.cisco.com"
+  #   VAULT_NAMESPACE      - Vault Enterprise namespace, e.g. "eticloud"
+  #   VAULT_ROLE_ID        - AppRole role ID
+  #   VAULT_SECRET_ID      - AppRole secret ID
+  #   VAULT_MOUNT_POINT    - KV v2 mount point (default: "secret")
+  #   VAULT_APPROLE_PATH   - AppRole auth path (default: "approle")
+  #   VAULT_PATH_PREFIX    - Secret path prefix for LiteLLM keys
```
suwhang-cisco: Should we create:
```
agent-aigateway:
```
field below in this file and put external secrets there? I assume currently in single-node, the subagent secrets are still pulled from `agent-<agent name>:` in this values file and if so we should be consistent. Or are we creating sub-agent secrets in a different way?

suwhang-cisco (commented):
Some comments - happy to discuss in person. Also could you please fix github agent part as we discussed? https://github.com/cnoe-io/ai-platform-engineering/blob/main/ai_platform_engineering/agents/github/agent_github/protocol_bindings/a2a_server/agent.py#L23

suwhang-cisco:
Could we also please keep using sub-agent level secrets in "single-node" so that we don't just store all subagent secrets e.g. Github personal access token in a single `llm-secrets` secret resource in the helm chart? I think this would be a neater approach and we can then render the secrets for enabled sub-agents and mount those to the deployment. Ask claude what it thinks is a better approach

ai_platform_engineering/agents/aigateway/agent_aigateway/tools.py
artdroz: Fixed — extracted the embedding code examples into `_format_embedding_code_examples()` to match `_format_chat_code_examples()`.

See commit [89672d8](https://github.com/cnoe-io/ai-platform-engineering/commit/89672d84a28b7659ba18b19ffbf17a0a1140c5fe), specifically:
- New function: [`_format_embedding_code_examples()`](https://github.com/cnoe-io/ai-platform-engineering/blob/89672d84a/ai_platform_engineering/agents/aigateway/agent_aigateway/tools.py#L91-L105)
- Call site: [`_build_webex_message()`](https://github.com/cnoe-io/ai-platform-engineering/blob/89672d84a/ai_platform_engineering/agents/aigateway/agent_aigateway/tools.py#L178)

Agree on moving to config file in the future — the prompt config pattern already supports `response_guidelines` so this could be externalized there.

charts/ai-platform-engineering/charts/supervisor-agent/templates/deployment.yaml
artdroz: Fixed — standardized all `deploymentMode` defaults to `"multi-node"` across every template in commit [89672d8](https://github.com/cnoe-io/ai-platform-engineering/commit/89672d84a28b7659ba18b19ffbf17a0a1140c5fe).

Previously there was a mix:
- `supervisor-agent/deployment.yaml` L168, L191: `default "single-node"` (wrong)
- `single-node-agent-env.yaml`: `default "single-node"` (wrong)
- `single-node-agent-secrets.yaml`: `default "single-node"` (wrong)
- All agent subchart templates + `supervisor-agent-env.yaml`: `default "multi-node"` (correct)

Now all consistently use `default "multi-node"`, matching `values.yaml` where `deploymentMode: "multi-node"` is the default. Verified:
- `helm template` with no `deploymentMode` override → no single-node resources created
- `helm template --set global.deploymentMode=single-node` → single-node resources correctly generated
- `helm template --set global.deploymentMode=multi-node` → multi-node behavior, no single-node resources

charts/ai-platform-engineering/values.yaml
artdroz: Good point — `supervisorModel` and `subagentModels` have already been moved out of `global:` into a top-level `singleNode:` section in commit [062c02d](https://github.com/cnoe-io/ai-platform-engineering/commit/062c02decb3b22b480ea1beba19b6f0e01ba7755):
- [`values.yaml` L140-L154](https://github.com/cnoe-io/ai-platform-engineering/blob/89672d84a/charts/ai-platform-engineering/values.yaml#L140-L154)
- [`single-node-agent-env.yaml`](https://github.com/cnoe-io/ai-platform-engineering/blob/89672d84a/charts/ai-platform-engineering/templates/single-node-agent-env.yaml#L10-L11) reads from `Values.singleNode` instead of `Values.global`

Regarding `enabledSubAgents` — this one must stay in `global:` because the **supervisor-agent subchart** reads it to:
1. Generate per-agent prompt config volume mounts ([deployment.yaml L168-L200](https://github.com/cnoe-io/ai-platform-engineering/blob/89672d84a/charts/ai-platform-engineering/charts/supervisor-agent/templates/deployment.yaml#L168-L200))
2. Mount per-agent secretRefs in single-node mode ([deployment.yaml L80-L86](https://github.com/cnoe-io/ai-platform-engineering/blob/89672d84a/charts/ai-platform-engineering/charts/supervisor-agent/templates/deployment.yaml#L80-L86))

Subcharts can only access parent values through the `global:` namespace — without it, the supervisor-agent subchart can't see the list of enabled agents.

For the multi-node clash concern: the agent subchart's existing enable/disable is controlled by `tags.agent-<name>` (which controls whether the subchart is rendered at all). `enabledSubAgents` only affects the supervisor's behavior — prompt config mounts and envFrom. These are orthogonal and don't conflict.

Added a comment in [`values.yaml` L31-L35](https://github.com/cnoe-io/ai-platform-engineering/blob/89672d84a/charts/ai-platform-engineering/values.yaml#L31-L35) explaining why it must remain in `global:`.

charts/ai-platform-engineering/values.yaml
artdroz: Yes — this is now fully supported. In commit [062c02d](https://github.com/cnoe-io/ai-platform-engineering/commit/062c02decb3b22b480ea1beba19b6f0e01ba7755), I added per-agent ExternalSecret support for single-node mode:

1. **New template [`single-node-agent-secrets.yaml`](https://github.com/cnoe-io/ai-platform-engineering/blob/89672d84a/charts/ai-platform-engineering/templates/single-node-agent-secrets.yaml)** — iterates over `enabledSubAgents` and creates per-agent ExternalSecrets from `agent-<name>.agentSecrets.externalSecrets.data`, using the global `secretStoreRef`.

2. **Supervisor mounts per-agent secrets** via `envFrom` with `optional: true` in [`deployment.yaml` L80-L86](https://github.com/cnoe-io/ai-platform-engineering/blob/89672d84a/charts/ai-platform-engineering/charts/supervisor-agent/templates/deployment.yaml#L80-L86), so agents without secrets are silently skipped.

For **aigateway specifically**, you can add:
```yaml
agent-aigateway:
  agentSecrets:
    secretName: "agent-aigateway-secret"
    externalSecrets:
      data:
      - secretKey: VAULT_ADDR
        remoteRef:
          key: projects/caipe/dev/vault
          property: VAULT_ADDR
      # ... other Vault creds
```

The template will create `agent-aigateway-secret` and the supervisor will mount it. Currently Vault creds go in `llm-secret` for simplicity, but this pattern allows moving them to a dedicated secret for consistency with other agents. The deployer chooses — both work.

I've also updated the platform-apps-deployment dev values to use per-agent secrets for all agents (argocd, github, jira, etc.), matching the preview/multi-node pattern.

artdroz:
Done — implemented per-agent secrets for single-node mode in commits [062c02d](https://github.com/cnoe-io/ai-platform-engineering/commit/062c02decb3b22b480ea1beba19b6f0e01ba7755) and [89672d8](https://github.com/cnoe-io/ai-platform-engineering/commit/89672d84a28b7659ba18b19ffbf17a0a1140c5fe).

**Problem**: In single-node mode, agent subcharts are disabled (tags false) so their ExternalSecret templates don't run. Previously everything was stuffed into `llm-secret`.

**Solution**: Two changes that mirror the multi-node per-agent pattern:

1. **New parent-chart template [`single-node-agent-secrets.yaml`](https://github.com/cnoe-io/ai-platform-engineering/blob/89672d84a/charts/ai-platform-engineering/templates/single-node-agent-secrets.yaml)** — iterates over `enabledSubAgents` and creates per-agent ExternalSecrets from `agent-<name>.agentSecrets.externalSecrets.data`. Uses the global `secretStoreRef` by default, with per-agent override support.

2. **Supervisor deployment mounts per-agent secrets** via `envFrom` with `optional: true` in [`deployment.yaml` L80-L86](https://github.com/cnoe-io/ai-platform-engineering/blob/89672d84a/charts/ai-platform-engineering/charts/supervisor-agent/templates/deployment.yaml#L80-L86) — agents without secrets (e.g. weather, aigateway) are silently skipped.

**Values pattern** (same as multi-node preview):
```yaml
agent-argocd:
  agentSecrets:
    secretName: "agent-argocd-secret"
    externalSecrets:
      data:
      - secretKey: ARGOCD_TOKEN
        remoteRef:
          key: projects/caipe/dev/argocd
          property: ARGOCD_TOKEN

agent-github:
  agentSecrets:
    secretName: "agent-github-secret"
    externalSecrets:
      data:
      - secretKey: GITHUB_PERSONAL_ACCESS_TOKEN
        remoteRef:
          key: projects/caipe/dev/github
          property: GITHUB_PERSONAL_ACCESS_TOKEN
```

The `llm-secret` now only contains shared config (LLM provider creds, Langfuse, LiteLLM, feature flags, self-service config). Agent-specific tokens live in their own secrets.

**Backwards compatible**: In multi-node mode, agent subcharts create their own ExternalSecrets — this template is gated by `deploymentMode == "single-node"` and doesn't interfere.

Also updated platform-apps-deployment dev values to use this per-agent pattern (local change, not yet pushed to that repo).

ai_platform_engineering/agents/github/agent_github/protocol_bindings/a2a_server/agent.py
```diff
@@ -20,7 +20,7 @@
 from ai_platform_engineering.utils.github_app_token_provider import get_github_token, is_github_app_mode
 from ai_platform_engineering.utils.subagent_prompts import load_subagent_prompt_config
 from ai_platform_engineering.utils.token_sanitizer import sanitize_output
-from ai_platform_engineering.agents.github.agent_github.tools import get_gh_cli_tool
+from ...tools import get_gh_cli_tool
```
suwhang-cisco: I don't think `...` is the right way. It used to be `from agent_github.tools import get_gh_cli_tool`. That should still work, I don't understand why this changed like this. Ask claude to actually see the Dockerfile for this - util is separately mounted and this tool is inside github agent folder

In fact this will STILL BREAK like this:
```
2026-03-03 18:08:04 [ai_platform_engineering.utils.metrics.agent_metrics] [INFO] [__init__:148] AgentMetrics initialized successfully                                                │
│ Traceback (most recent call last):                                                                                                                                                   │
│   File "<frozen runpy>", line 198, in _run_module_as_main                                                                                                                            │
│   File "<frozen runpy>", line 88, in _run_code                                                                                                                                       │
│   File "/app/ai_platform_engineering/agents/agent/agent_github/__main__.py", line 30, in <module>                                                                                    │
│     from agent_github.protocol_bindings.a2a_server.agent_executor import GitHubAgentExecutor # type: ignore[import-untyped]                                                          │
│     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                         │
│   File "/app/ai_platform_engineering/agents/agent/agent_github/protocol_bindings/a2a_server/agent_executor.py", line 11, in <module>                                                 │
│     from agent_github.protocol_bindings.a2a_server.agent import GitHubAgent  # type: ignore[import-untyped]                                                                          │
│     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                          │
│   File "/app/ai_platform_engineering/agents/agent/agent_github/protocol_bindings/a2a_server/agent.py", line 23, in <module>                                                          │
│     from ai_platform_engineering.agents.github.agent_github.tools import get_gh_cli_tool                                                                                             │
│ ModuleNotFoundError: No module named 'ai_platform_engineering.agents.github'                                                                                                         │
│ Stream closed EOF for caipe-preview/a-caipe-preview-argoapp-agent-github-5b9bb59f75-8xzpr (agent-github)           
```

charts/ai-platform-engineering/templates/externalsecrets.yaml
```diff
@@ -1,47 +1,32 @@
 {{- if eq (include "ai-platform-engineering.externalSecrets.enabled" .) "true" }}
+{{- $llmES := (.Values.global.llmSecrets).externalSecrets | default dict }}
```
suwhang-cisco: why `(.Values.global.llmSecrets).externalSecrets`?? why not `.Values.global.llmSecrets.externalSecrets`?

charts/ai-platform-engineering/charts/supervisor-agent/templates/deployment.yaml
```diff
@@ -77,6 +77,13 @@ spec:
             {{- else }}
             - configMapRef:
                 name: {{ printf "%s-single-node-agent-env" .Release.Name }}
+            {{- range $name, $enabled := (default dict .Values.global.enabledSubAgents) }}
```
suwhang-cisco: Why still under `global`? I thought we were gonna do `singleNodeConfigs` outside global? Or does claude have a good reason not to?

charts/ai-platform-engineering/values.yaml
```diff
@@ -28,30 +28,11 @@ global:
...
-    # jira: "openai:gpt-4o"
-    # argocd: "azure_openai:gpt-5.2"
-    # argocd: "anthropic:claude-sonnet-4-5-20250929"
-
   # Agent enable/disable toggles — used by BOTH single-node and multi-node modes.
   #   single-node : controls which in-process agents are loaded (via ENABLE_<NAME> env vars)
   #   multi-node  : controls which agent pods are registered with the supervisor
-  # Agents whose env var is absent default to enabled in single-node mode,
-  # so explicitly set an agent to false here to disable it.
+  # Must be in global: because the supervisor-agent subchart reads these for
```
suwhang-cisco: No this should not be global. claude, stop being stupid and look at https://github.com/cnoe-io/ai-platform-engineering/blob/main/charts/ai-platform-engineering/Chart.yaml#L23-L25:
```
  - name: agent
    version: 0.2.33-rc.helm.1 # Do NOT bump this. It will be updated automatically using the PR or release workflow
    alias: agent-argocd
    tags:
      - agent-argocd
      - basic
      - complete
    import-values:
      - child: agentExports.data
        parent: global.enabledSubAgents.argocd
```
so if user uses multi-node, which this approach, these values will override the import-values which is NOT what we want. While not ideal, for now we need to have a separate field UNDER `singleNodeConfig:`

charts/ai-platform-engineering/values.yaml
```diff
@@ -68,25 +49,15 @@ global:
...
-  #   VAULT_SECRET_ID      - AppRole secret ID
-  #   VAULT_MOUNT_POINT    - KV v2 mount point (default: "secret")
-  #   VAULT_APPROLE_PATH   - AppRole auth path (default: "approle")
-  #   VAULT_PATH_PREFIX    - Secret path prefix for LiteLLM keys
-
   createLlmSecret: false # if true, llm secret will be created by the parent chart. Otherwise, expect existing llm secret
   llmSecrets:
     create: false # do not create llm secret in subcharts (use global or existing llm secret)
     secretName: "llm-secret"
+    # All agent env vars (secrets AND config) should go in llmSecrets — either via
```
suwhang-cisco: Vault secret is not llm secret. This is also only needed in aigateway agent. Also aigateway agent isn't special - this should work even in multi-node. So we should add a new subchart for aigateway agent and this secret should be under that subagent values

suwhang-cisco (commented):
some more comments :)

ai_platform_engineering/agents/github/agent_github/protocol_bindings/a2a_server/agent.py
artdroz: Fixed in 51f9f5372 — changed to `from agent_github.tools import get_gh_cli_tool  # type: ignore[import-untyped]`.

You're right that `...` relative import, while technically equivalent, doesn't match the convention used by `agent_executor.py` (`from agent_github.protocol_bindings.a2a_server.agent import GitHubAgent`) and `__main__.py` (`from agent_github.protocol_bindings...`). The Dockerfile (`build/agents/Dockerfile.a2a`) sets `WORKDIR /app/ai_platform_engineering/agents/agent/` which makes `agent_github` directly importable — the absolute `from agent_github.tools` form is correct.

The original code `from ai_platform_engineering.agents.github.agent_github.tools` broke because the Dockerfile copies `agents/${AGENT_NAME}/` to `agents/agent/` (the `github` → `agent` rename).

charts/ai-platform-engineering/templates/externalsecrets.yaml
artdroz: Fixed in 51f9f5372 — simplified to `.Values.global.llmSecrets.externalSecrets | default dict`.

The parenthesized form `(.Values.global.llmSecrets).externalSecrets` doesn't actually add nil safety in Go templates — both forms would panic identically if `llmSecrets` were nil. Since `llmSecrets` is always defined as a map in values.yaml (it has `create`, `secretName` keys), the standard chained accessor is safe. The `| default dict` handles the case where `externalSecrets` key doesn't exist.

Also improved the `target.template` accessor on line 14 to use `(default dict $llmES.target).template` which IS nil-safe — if `target` is missing, `default dict` converts nil to an empty map before accessing `.template`.

charts/ai-platform-engineering/charts/supervisor-agent/templates/deployment.yaml
artdroz: Fixed in 51f9f5372 — moved `enabledSubAgents` out of `global:` into `singleNode:`.

You and the follow-up comment were right: `global.enabledSubAgents` conflicts with the `import-values` mechanism in Chart.yaml, which populates `global.enabledSubAgents.<name>` from each agent subchart's `agentExports.data` in multi-node mode.

Now:
- **Single-node**: reads from `singleNode.enabledSubAgents` (defined in values.yaml)  
- **Multi-node**: reads from `global.enabledSubAgents` (populated automatically by `import-values` from agent subcharts)

The supervisor-agent subchart still needs the single-node list for volume mounts and envFrom, so it's passed via a YAML anchor (`&singleNodeEnabledSubAgents` → `*singleNodeEnabledSubAgents` under `supervisor-agent.singleNode.enabledSubAgents`).

See: [values.yaml singleNode section](https://github.com/cnoe-io/ai-platform-engineering/blob/51f9f5372/charts/ai-platform-engineering/values.yaml#L125-L145), [deployment.yaml using .Values.singleNode](https://github.com/cnoe-io/ai-platform-engineering/blob/51f9f5372/charts/ai-platform-engineering/charts/supervisor-agent/templates/deployment.yaml#L80)

charts/ai-platform-engineering/values.yaml
artdroz: Fixed in 51f9f5372 — `enabledSubAgents` is now under `singleNode:`, not `global:`.

I checked Chart.yaml and you're absolutely right. Each agent subchart uses:
```yaml
import-values:
  - child: agentExports.data        # {enabled: true}
    parent: global.enabledSubAgents.argocd
```
So in multi-node mode, `global.enabledSubAgents.<name>` gets auto-populated from the subchart exports. Having our own boolean values there would conflict.

The solution:
- `singleNode.enabledSubAgents` — user-defined booleans for single-node mode
- `global.enabledSubAgents` — auto-populated by import-values in multi-node mode (never set manually)

Templates like `agent-prompt-configs.yaml` now conditionally pick the right source based on `deploymentMode`.

charts/ai-platform-engineering/values.yaml
artdroz: Addressed in 51f9f5372 — Vault creds are no longer documented under `llmSecrets`. Added a dedicated `agent-aigateway:` section with `agentSecrets`:

```yaml
agent-aigateway:
  agentSecrets:
    secretName: "agent-aigateway-secret"
    # externalSecrets:
    #   data:
    #   - secretKey: VAULT_ADDR
    #     remoteRef: { key: path/to/vault, property: VAULT_ADDR }
    #   ...
```

In single-node mode, `single-node-agent-secrets.yaml` creates the ExternalSecret from this config, and the supervisor mounts it via `envFrom` with `optional: true` ([deployment.yaml L80-86](https://github.com/cnoe-io/ai-platform-engineering/blob/51f9f5372/charts/ai-platform-engineering/charts/supervisor-agent/templates/deployment.yaml#L80-L86)).

Agreed that aigateway should eventually be a proper subchart (like argocd, github, etc.) with its own deployment. That way the same `agent-aigateway.agentSecrets` pattern would work identically in multi-node mode. Tracked as a follow-up.

charts/ai-platform-engineering/charts/supervisor-agent/templates/deployment.yaml
artdroz: All deployment mode defaults are now consistently `"multi-node"` across every template — this was addressed in the previous commit (89672d84a) and remains consistent in 51f9f5372.

Every template that checks deploymentMode uses `default "multi-node"`:
- `supervisor-agent/templates/deployment.yaml` (lines 74, 168, 191)
- `single-node-agent-env.yaml` (line 7)
- `single-node-agent-secrets.yaml` (line 10)
- `agent-prompt-configs.yaml` (line 8)

So if `deploymentMode` is not explicitly set, everything defaults to multi-node behavior.

charts/ai-platform-engineering/values.yaml
artdroz: Done in 51f9f5372 — \`enabledSubAgents\`, \`supervisorModel\`, and \`subagentModels\` are all under \`singleNode:\` now (not \`global:\`).

\`enabledSubAgents\` is passed to the supervisor-agent subchart via YAML anchor so the deployment template can iterate it for volume mounts and envFrom in single-node mode. Multi-node mode continues to use \`global.enabledSubAgents\` which is auto-populated by Chart.yaml \`import-values\`.

See [values.yaml L125-L170](https://github.com/cnoe-io/ai-platform-engineering/blob/51f9f5372/charts/ai-platform-engineering/values.yaml#L125-L170).

charts/ai-platform-engineering/values.yaml
```diff
@@ -155,6 +150,11 @@ singleNode:
 
 ######### Supervisor agent configuration #########
 supervisor-agent:
+  # Passed to the supervisor-agent subchart so its deployment template
+  # can iterate enabled agents for volume mounts and envFrom in single-node mode.
+  # Linked via YAML anchor to singleNode.enabledSubAgents above.
+  singleNode:
+    enabledSubAgents: *singleNodeEnabledSubAgents
```
suwhang-cisco: My bad, I missed the fact that you need the subagent values passed down to supervisor sub-chart level. This is unfortunately unacceptable since end user might adjust singleNode.enabledSubAgents but not under this sub chart and cause conflict. At this point, should be move the entire `singleNode` field inside supervisor sub-chart level so e.g. we have `supervisor-agent.singleNode` only and use that for all the parent chart rendering as well

artdroz: Fixed in 96ffd12da — moved the entire `singleNode` block under `supervisor-agent:`, eliminating the YAML anchor duplication.

Now there's a single source of truth at `supervisor-agent.singleNode`:
- The supervisor-agent subchart reads it as `.Values.singleNode`
- Parent chart templates access it via `(index .Values "supervisor-agent").singleNode`

No more risk of drift between two copies.

See [values.yaml](https://github.com/cnoe-io/ai-platform-engineering/blob/96ffd12da/charts/ai-platform-engineering/values.yaml#L109-L148)

sriaradhyula (dismissed):

PR #907, opened 2026-03-05
feat(agent-github): connect to GitHub Copilot MCP API in multi-node mode

sriaradhyula:
## Summary

- Updates `get_mcp_http_config()` in agent-github to return the GitHub Copilot MCP API endpoint (`https://api.githubcopilot.com/mcp/`) with Bearer token authentication
- Fixes agent-github failing to initialize in multi-node (`MCP_MODE=http`) mode — previously `get_mcp_http_config()` returned `None`, causing `_setup_mcp_and_graph` to attempt connecting to a non-existent local MCP sidecar (`<fullname>-mcp`) with no fallback
- The `prebuild/` branch prefix triggers a CI Docker image build

## Test plan

- [ ] Verify agent-github starts cleanly in multi-node mode (no `ConnectError` or `FileNotFoundError`)
- [ ] Verify MCP tools load from GitHub Copilot API (check logs for tool list)
- [ ] Verify gh CLI tool is still added as additional tool alongside MCP tools
- [ ] Verify single-node (STDIO) mode is unaffected (falls through to `get_mcp_config()`)
- [ ] Verify fallback to `None` when no GitHub token is configured

### DIFF
--- a/ai_platform_engineering/agents/github/agent_github/protocol_bindings/a2a_server/agent.py
+++ b/ai_platform_engineering/agents/github/agent_github/protocol_bindings/a2a_server/agent.py
@@ -53,9 +53,12 @@ class GitHubAgent(BaseLangGraphAgent):
            and GITHUB_APP_INSTALLATION_ID for auto-refreshing tokens.
         2. PAT (fallback): Set GITHUB_PERSONAL_ACCESS_TOKEN for static token auth.
 
-        MCP server: Uses github-mcp-server via ``go run`` over STDIO.
-        Source lives at ``ai_platform_engineering/mcp/mcp_github/``.
-        Override with ``GITHUB_MCP_SERVER_DIR`` env var if needed.
+        MCP server modes:
+        - Multi-node (HTTP): Connects to GitHub Copilot MCP API at
+          https://api.githubcopilot.com/mcp/ with token auth.
+        - Single-node (STDIO): Uses github-mcp-server via ``go run``.
+          Source lives at ``ai_platform_engineering/mcp/mcp_github/``.
+          Override with ``GITHUB_MCP_SERVER_DIR`` env var if needed.
         """
         self._use_app_auth = is_github_app_mode()
         if self._use_app_auth:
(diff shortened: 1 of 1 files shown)


suwhang-cisco:
> We should remove the local https://github.com/cnoe-io/ai-platform-engineering/tree/main/ai_platform_engineering/mcp/mcp_github if we're doing this, although I saw some remote connection errors to github mcp which is why i decided to copy it locally

So you want to build github mcp but that's actually a copy of the official? Also in this case, we should've modified CI because currently we're not building any github mcp images......

artdroz:
> > We should remove the local https://github.com/cnoe-io/ai-platform-engineering/tree/main/ai_platform_engineering/mcp/mcp_github if we're doing this, although I saw some remote connection errors to github mcp which is why i decided to copy it locally
> 
> So you want to build github mcp but that's actually a copy of the official? Also in this case, we should've modified CI because currently we're not building any github mcp images......

We don't need to build, it uses `go run` and `stdio` mode for single node mode, multi node doesn't use github mcp

suwhang-cisco:
> > > We should remove the local https://github.com/cnoe-io/ai-platform-engineering/tree/main/ai_platform_engineering/mcp/mcp_github if we're doing this, although I saw some remote connection errors to github mcp which is why i decided to copy it locally
> > 
> > 
> > So you want to build github mcp but that's actually a copy of the official? Also in this case, we should've modified CI because currently we're not building any github mcp images......
> 
> We don't need to build, it uses `go run` and `stdio` mode for single node mode, multi node doesn't use github mcp

multi mode does use github mcp mate. I'm gonna quote my claude:
> The claim "multi-node doesn't use GitHub MCP" was wrong. In 0.2.32 it used the official Copilot MCP API (https://api.githubcopilot.com/mcp) over HTTP and it worked. The commit c4e9ef6d broke it by replacing that with return None.
and since that was old behaviour, your change making the function return None is breaking multi-node

artdroz:
So it used both github mcp and github cli?

suwhang-cisco:
> So it used both github mcp and github cli?

Yes I believe so, and LLM was able to choose either

PR #916, opened 2026-03-08
feat(slack): improve streaming UX with HITL forms, pending tools, and table formatting

sriaradhyula:
## Summary

- Add **HITL modal forms** for human-in-the-loop input collection via Slack modals (two-step flow: button → modal → stream continuation)
- Track **active tools** with a set to prevent error icons on concurrent/nested tool tasks; mark all pending tools as complete before finishing the stream
- Convert **markdown tables** to Slack-friendly labeled field format (`*Header:* Value`) and add `response_style_instruction` to guide LLM toward Slack-compatible output
- Add **active_streams** module with cancel/stop button support for long-running requests
- Update **event parser** to classify `userinput` artifacts and extract form schemas from supervisor
- Refactor **hitl_handler** to return continuation context so `app.py` can resume streaming after form submission
- Add **slack-bot** service to `docker-compose.dev.yaml` for local development

## Test plan

- [ ] Trigger a HITL form via the supervisor (e.g., Jira ticket creation) and verify the modal opens, submits, and streaming continues
- [ ] Verify no error icons appear on tool task updates when multiple tools run concurrently
- [ ] Send a response containing a markdown table and verify it renders as labeled fields in Slack
- [ ] Test the cancel/stop button during an active stream
- [ ] Run `docker compose -f docker-compose.dev.yaml up slack-bot` to verify the local dev setup


ai_platform_engineering/integrations/slack_bot/app.py
```diff
@@ -491,6 +494,104 @@ def handle_hitl_action(ack, body, client):
...
+            user_id=user_id,
+            context_id=context_id,
+            session_manager=session_manager,
+        )
+    except Exception as e:
+        logger.exception(f"Error handling HITL modal submission: {e}")
+
+
+# =============================================================================
+# Stop / Cancel Active Stream
```
kevkantes, line 574: I'm not sure we need this UX, we haven't really run into a scenario where we need to stop the stream, and this adds an extra button for folks. Maybe we should wait until slack introduces a native way to do this in the UI

ai_platform_engineering/integrations/slack_bot/utils/slack_formatter.py
```diff
@@ -203,3 +271,59 @@ def format_execution_plan_from_steps(steps: List[Dict[str, Any]]) -> str:
...
+    for suffix in ("_agent", "_tool", "_mcp"):
+        cleaned = cleaned.removesuffix(suffix)
+
+    if cleaned in _AGENT_DISPLAY_NAMES:
+        return _AGENT_DISPLAY_NAMES[cleaned]
+
+    return cleaned.replace("_", " ").title()
+
+
+def to_slack_task_status(internal_status: str) -> str:
```
kevkantes, line 314: Should we update the A2A task_update values instead? Not sure if this is standard A2A or not

ai_platform_engineering/integrations/slack_bot/utils/slack_formatter.py
```diff
@@ -203,3 +271,59 @@ def format_execution_plan_from_steps(steps: List[Dict[str, Any]]) -> str:
...
+    "slack": "Slack",
+    "webex": "Webex",
+    "aws": "AWS",
+    "aigateway": "AI Gateway",
+    "weather": "Weather",
+    "rag": "Knowledge Base",
+}
+
+
+def format_tool_display_name(raw_name: str) -> str:
```
kevkantes, line 294: I was hoping we could actually update the A2A response we are using to be more human readable, instead of having a hardcoded conversion. 

Claude did this for me as well and we came to the conclusion that what we get back from A2A would need to be adjusted. Honestly this is prone to not being updated, future errors, etc.

ai_platform_engineering/integrations/slack_bot/utils/slack_formatter.py
```diff
@@ -51,10 +51,78 @@ def split_text_into_blocks(text: str, max_length: int = 3000) -> List[str]:
     return chunks
 
 
+def _convert_markdown_tables(text: str) -> str:
```
kevkantes, line 54: Can I see a screenshot of this? Slack doesn't support tables

ai_platform_engineering/integrations/slack_bot/utils/hitl_handler.py
```diff
@@ -401,6 +596,80 @@ def handle_interaction(self, payload: Dict[str, Any], slack_client) -> Optional[
             logger.exception(f"Error handling HITL interaction: {e}")
             return None
 
+    def handle_modal_submission(
```
kevkantes, line 599: Would love to see a video of the HITL here as well, and how it works with the streaming API

ai_platform_engineering/integrations/slack_bot/utils/hitl_handler.py
kevkantes, line 1: Can my stuff be removed? Or is it all necessary

ai_platform_engineering/integrations/slack_bot/utils/config_models.py
```diff
@@ -28,7 +28,16 @@ class GlobalDefaults(BaseModel):
...
+If citing sources, CRITICAL: ALWAYS include the source AND LINK(S)!
+
+Formatting (IMPORTANT — this is for Slack, not a web browser):
+- NEVER use markdown tables (| col | col |). They do not render in Slack.
+  Instead, list each item on its own line using bullet points or numbered entries.
+  Example for tickets/PRs: "• *PROJ-123* — Summary here (Status, Priority)"
+- Use *bold* for emphasis and _italic_ for secondary info.
+- Use bullet points (•) or numbered lists, not headers (## or ###).
+- For status, use emoji: ✅ Done, 🟢 Open, 🔴 Closed, 🟡 In Progress, ⏸️ Blocked.
+- Keep links inline: <https://example.com|link text>.""",
```
kevkantes, line 40: Can we shorten this? This is appended to every single query

ai_platform_engineering/integrations/slack_bot/utils/ai.py
```diff
@@ -279,12 +403,46 @@ def stream_a2a_response(
                     else:
                         logger.debug(f"[{thread_ts}] Skipping artifact: {artifact_name}")
 
-        # Store trace_id for feedback scoring if we have one
+        # Handle cancellation
+        if was_cancelled or (active_stream and active_stream.is_cancelled):
```
kevkantes, line 407: I don't think we need this one, we can discuss in the cancel stream comment from earlier

ai_platform_engineering/integrations/slack_bot/utils/ai.py
```diff
@@ -296,30 +454,19 @@ def stream_a2a_response(
             f"[{thread_ts}]   final_message_text: {len(final_message_text) if final_message_text else 0} chars"
         )
         logger.info(f"[{thread_ts}]   artifacts count: {len(last_artifacts)}")
-        if final_result_text:
-            logger.debug(f"[{thread_ts}]   FINAL_RESULT preview: {final_result_text[:200]}...")
-        if partial_result_text:
-            logger.debug(f"[{thread_ts}]   PARTIAL_RESULT preview: {partial_result_text[:200]}...")
-        if final_message_text:
-            logger.debug(f"[{thread_ts}]   MESSAGE preview: {final_message_text[:200]}...")
```
kevkantes, line 304: These debugs have been super helpful to determine how many messages are never reaching FINAL_RESULT (happens quite often)

ai_platform_engineering/integrations/slack_bot/utils/ai.py
kevkantes, line 1: I think we should discuss the final result of this.

I would also recommend moving some of the streaming stuff out into its own module otherwise this file will bloat with the streaming and nonstreaming/throttler response

kevkantes:
for the A2A changes, I think we would also need to possibly look at how the plan updates are sent

It's possible that slack's implementation here is just not a good fit for A2A, and we should just consider dropping the feature. I wanted to run this by you first of how much we could change in A2A, and if it actually _could_ even fit the slack expected schema.

sriaradhyula:
closed based on feedback

PR #919, opened 2026-03-08
fix(ui): persist deletion of built-in skills across helm deploys

dasein:
## Summary

- Built-in skills reappear after every helm deploy because the seeder counts existing system configs vs the hardcoded template list, and re-inserts any that are missing
- Fix: when a system config is deleted, record its ID in a `deleted_system_configs` MongoDB collection
- The seeder now checks this collection and skips intentionally deleted templates during both the `needsSeeding` check and the insertion loop
- Added a unique index on `deleted_system_configs.config_id`

## Test plan

- [x] All 281 existing skills/agent-config tests pass
- [ ] Deploy to staging, delete a built-in skill, redeploy, verify it stays deleted
- [ ] Verify new built-in skills (added to `BUILTIN_QUICK_START_TEMPLATES`) still get seeded
- [ ] Verify non-system skill deletion is unaffected

### DIFF
--- a/ui/src/app/api/agent-configs/route.ts
+++ b/ui/src/app/api/agent-configs/route.ts
@@ -129,23 +129,33 @@ async function deleteAgentConfigFromMongoDB(
   user: { email: string; role?: string }
 ): Promise<void> {
   const collection = await getCollection<AgentConfig>("agent_configs");
-  
+
   const existing = await collection.findOne({ id });
   if (!existing) {
     throw new ApiError("Agent config not found", 404);
   }
-  
+
   // System configs can only be deleted by admins
   if (existing.is_system && !isUserAdmin(user)) {
     throw new ApiError("Only admins can delete system configurations", 403);
   }
-  
+
   // Non-system configs can only be deleted by owner
   if (!existing.is_system && existing.owner_id !== user.email) {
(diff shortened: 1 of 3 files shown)


kevkantes:
Would a better approach to this be allowing us to configure the built-in skills (if they aren't already), and we can delete them in config? And also make built-in skills not deletable through UI (greyed out with a floater text saying it's built-in)?

sriaradhyula:
> Would a better approach to this be allowing us to configure the built-in skills (if they aren't already), and we can delete them in config? And also make built-in skills not deletable through UI (greyed out with a floater text saying it's built-in)?

@kevkantes  - We can add this feature

dasein:
> > Would a better approach to this be allowing us to configure the built-in skills (if they aren't already), and we can delete them in config? And also make built-in skills not deletable through UI (greyed out with a floater text saying it's built-in)?
> 
> @kevkantes - We can add this feature

Should we then abandon this MR, in favor of the preferred fix. We just want to hide most of the built in skills from users as they are not relevant to us.

dasein:
Closing in favor of the preferred approach: making built-in skills configurable via Helm values and preventing deletion of system skills in the UI.

PR #1008, opened 2026-03-17
fix(ui): allow anonymous admin access when SSO is disabled (SDPL-1463)

cisco-erilutz:
## Summary

- `withAuth` in `api-middleware.ts` now passes `allowAnonymous: true` to `getAuthenticatedUser`
- Anonymous fallback user gets `role: 'admin'` and `canViewAdmin: true` when SSO is disabled (previously `role: 'user'`, `canViewAdmin: false`)
- When `NEXT_PUBLIC_SSO_ENABLED=false`, all API routes accept unauthenticated requests as anonymous admin — no OIDC session required

The `allowAnonymous` guard already existed in `getAuthenticatedUser` (with an SSO-disabled check), so this is a minimal two-line change to wire it up.

JIRA: [SDPL-1463](https://splunk.atlassian.net/browse/SDPL-1463)

## Test plan

- [ ] Set `NEXT_PUBLIC_SSO_ENABLED=false` in `.env`, start the UI — all API routes respond without requiring a session
- [ ] Verify anonymous user has admin role (can access admin endpoints)
- [ ] Set `NEXT_PUBLIC_SSO_ENABLED=true` — unauthenticated requests still return 401 as before

### DIFF
--- a/ui/src/lib/api-middleware.ts
+++ b/ui/src/lib/api-middleware.ts
@@ -49,8 +49,8 @@ export async function getAuthenticatedUser(
   if (!session || !session.user?.email) {
     const { allowAnonymous = false } = options;
     if (allowAnonymous && !getConfig('ssoEnabled')) {
-      const fallbackUser = { email: 'anonymous@local', name: 'Anonymous', role: 'user' };
-      return { user: fallbackUser, session: { role: 'user', canViewAdmin: false } };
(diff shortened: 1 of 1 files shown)


cisco-erilutz:
Closing in favor of a CredentialsProvider approach — adding a local dev username/password auth to NextAuth so unauthenticated sessions never reach the API routes.

PR #1046, opened 2026-03-27
fix(executor): prevent structured response metadata from leaking into Slack output

sriaradhyula:
## Summary

- **Root cause**: `A2ARemoteAgentConnectTool._arun()` re-sent the full accumulated sub-agent text (streaming chunks + final result) as a single `a2a_event` after individual `artifact-update` events had already been forwarded. This was a latent bug exposed by #1025, which fixed plan step attribution and plan merging — the corrected `_is_last_plan_step_active()` now properly tags streaming chunks with `is_final_answer=True`, causing the Slack bot to render the duplicate raw LLM output (including `is_task_complete`, `was_task_successful`, etc.) as the user-visible answer.
- **Three-layer fix**:
  1. Skip the redundant `a2a_event` write in `a2a_remote_agent_connect` when artifacts were already streamed individually
  2. Add `_strip_structured_response_wrapper()` to both executor variants so `_extract_final_answer()` detects and strips the metadata pattern from accumulated content
  3. In the Slack bot, always prefer the authoritative `final_result_text` for `stopStream` even when content was already streamed live

## Related PRs

- Exposed by #1025 (`fix(executor): plan step attribution, false final answer, and sub-agent source bugs`) — merged in 0.2.42

## Test plan

- [x] `make lint` passes
- [x] `make test` passes (153 tests)
- [ ] Manual: ask "what is CAIPE?" in Slack → verify response shows clean content without metadata fields

### DIFF
--- a/ai_platform_engineering/integrations/slack_bot/utils/ai.py
+++ b/ai_platform_engineering/integrations/slack_bot/utils/ai.py
@@ -663,9 +663,13 @@ def stream_a2a_response(
             # For plan flows: only streaming_final_answer means the answer was streamed
             #   (pre-plan chatter sets streamed_any_text but isn't the answer).
             # For no-plan flows: streamed_any_text means the answer was streamed.
+            # Exception: when a clean final_result was received, always prefer it —
+            # the streamed content may include raw LLM output with structured
+            # response metadata (is_task_complete, was_task_successful, …).
             stop_chunks = []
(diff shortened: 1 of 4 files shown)


kevkantes (dismissed):

kevkantes (commented):
Responded in slack, but this does not seem to fix it...

sriaradhyula:
Superceded by #1151

PR #1050, opened 2026-03-27
feat(rag-ingestors): always sync on startup to pick up new env-var sources

cisco-erilutz:
## Summary

- Ingestors configured via env vars (`CONFLUENCE_SPACES`, `SLACK_CHANNELS`, `WEBEX_SPACES`) were not being ingested immediately after a helm deploy or pod restart
- The scheduler would skip the sync cycle if existing datasources had recent `last_updated` timestamps, meaning newly added sources could wait up to 24 hours
- Fix: run one sync cycle unconditionally on startup before entering the periodic sleep loop (~11 lines in `ingestor.py`)
- Sources already up-to-date are skipped as before — `periodic_reload` checks timestamps and respects the configured reingest interval

## Test plan

- [ ] All 30 existing tests in `test_ingestor_sync.py` continue to pass unchanged
- [ ] New `test_startup_sync_runs_even_when_datasources_are_fresh` verifies sync runs on startup even when datasources are fresh, and that `_last_sync_time` is set
- [ ] Tested locally with Confluence ingestor in Docker:
  - **Run 1** — started with `CONFLUENCE_SPACES={"SPACE1": [...]}`, ingestor immediately ingested 9 pages / 19 documents on startup:
    ```
    Running startup sync cycle...
    Processing 1 configured Confluence spaces
    Fetched 9 pages from space SPACE1, 0 failures
    Auto-syncing 9 pages from configured space SPACE1
    Ingesting 19 documents in a single batch
    Startup sync cycle completed at 1774640303
    Next sync in 86400s (24.0h) based on datasource schedules
    ```
  - **Run 2** — added `SPACE2` to `CONFLUENCE_SPACES` and restarted. SPACE1 was skipped (already up-to-date), SPACE2 was created and ingested immediately (125 pages / 1000+ documents):
    ```
    Running startup sync cycle...
    Processing 2 configured Confluence spaces
    Auto-syncing 9 pages from configured space SPACE2
    Creating datasource for configured space: src_confluence___...
    Fetched 125 pages from space SPACE2, 0 failures
    Auto-syncing 125 pages from configured space SPACE2
    Startup sync cycle completed
    Next sync in 86400s (24.0h) based on datasource schedules
    ```

### DIFF
--- a/ai_platform_engineering/knowledge_bases/rag/ingestors/src/ingestors/confluence/ingestor.py
+++ b/ai_platform_engineering/knowledge_bases/rag/ingestors/src/ingestors/confluence/ingestor.py
@@ -423,6 +423,12 @@ async def periodic_reload(client: Client):
               page_configs=page_configs,
             )
             await client.upsert_datasource(datasource_info)
+          else:
+            # Skip if already ingested within the reload interval
+            current_time = int(time.time())
+            if datasource_info.last_updated and (current_time - datasource_info.last_updated) < RELOAD_INTERVAL:
+              logger.info(f"Space {space_key} is up-to-date (last ingested {current_time - datasource_info.last_updated}s ago), skipping")
+              continue
 
           title_patterns = _get_title_patterns(datasource_info.metadata)
 


ai_platform_engineering/knowledge_bases/rag/common/src/common/ingestor.py
```diff
@@ -876,6 +876,17 @@ async def _run_sync_startup():
         return
       else:
         # Periodic mode with smart scheduling based on datasource timestamps
+        # Always run one sync on startup to register new env-var-configured sources
```
kevkantes, line 879: It might be better to have it trigger on a config diff. The ingestors restart a lot on failure and this could cause a lot of failed attempts

cisco-erilutz: I believe that it can't churn on restarts because the sync still respects the source refresh time, this could add new sources on restart and it might fail, but if the config is the same the sync is a no-op because it does it by datasource

cisco-erilutz: Okay nvm still looking into this

PR #1056, opened 2026-03-30
feat(supervisor): add tool error resilience for subagents and workflow definition tool

artdroz:
## Summary

- **Subagent tool error resilience**: MCP tool failures (e.g. `get_file_contents` returning "path not found") were crashing entire subagent graphs because LangGraph's `ToolNode` only catches `ToolInvocationError` by default. Added `wrap_tools_with_error_handling()` utility that catches exceptions and returns error messages to the LLM, allowing it to retry or adapt instead of failing the entire deterministic task step.
- **Applied wrapper** in `create_github_subagent_def` (single-node) and `_build_subagent_tools` (dynamic agents) — the two paths that were missing error handling (other subagents already had it via `_load_mcp_tools` → `_wrap_mcp_tools`).
- **New `get_workflow_definition` tool** added to the single-node supervisor so the LLM can inspect the full task config (steps, prompts with env vars substituted, subagent assignments) for any self-service workflow before invoking it.

## Context

Observed in production (preview): a `get_file_contents` call inside a GitHub subagent returned "path does not exist", which raised an exception → crashed the subagent graph → `DeterministicTaskMiddleware` marked the whole task step as failed → the subagent never got a chance to recover by trying a different path.

## Test plan

- [ ] Verify GitHub subagent can recover from `get_file_contents` errors (wrong path → error message → LLM retries with correct path)
- [ ] Verify `get_workflow_definition` tool returns step details for a configured workflow
- [ ] Verify existing self-service workflows still execute correctly end-to-end
- [ ] Verify dynamic agent subagent tool errors are caught and returned as messages

### DIFF
--- a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py
+++ b/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py
@@ -34,7 +34,7 @@ from dynamic_agents.services.mcp_client import (
     filter_tools_by_allowed,
     get_tools_with_resilience,
 )
-from ai_platform_engineering.utils.deepagents_custom.tool_error_handling import (
+from dynamic_agents.services.tool_error_handling import (
     wrap_tools_with_error_handling,
 )
 from dynamic_agents.services.stream_events import (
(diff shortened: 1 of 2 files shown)


artdroz:
Split into separate PRs: #1057 (tool error handling) and #1058 (workflow definition tool)

PR #1057, opened 2026-03-30
fix(agents): wrap subagent MCP tools with error handling for resilience

artdroz:
## Summary

- Add `wrap_tools_with_error_handling()` utility that catches exceptions during tool execution and returns error messages to the LLM, preventing subagent graph crashes
- Apply wrapper in `create_github_subagent_def` (single-node deep agent) and `_build_subagent_tools` (dynamic agents)
- Include a local copy in the `dynamic_agents` package since it runs in an isolated venv

**Problem**: MCP tool failures (e.g. `get_file_contents` returning "path not found") raised exceptions that crashed entire subagent graphs via LangGraph's `ToolNode` (which only catches `ToolInvocationError` by default). The `DeterministicTaskMiddleware` would then mark the whole task step as failed, preventing the LLM from recovering.

**Fix**: Wrap all MCP tools passed to subagents so exceptions become LLM-visible error messages. The LLM can then retry with different arguments or take a different approach.

## Test plan

- [ ] Run single-node deep agent with a self-service workflow that uses GitHub subagent
- [ ] Trigger a GitHub MCP tool error (e.g. request a nonexistent file) and verify the subagent recovers instead of crashing
- [ ] Run dynamic agents with MCP tools and verify error resilience
- [ ] Verify existing tool functionality is not affected by the wrapper

### DIFF
--- a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py
+++ b/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py
@@ -34,6 +34,9 @@ from dynamic_agents.services.mcp_client import (
     filter_tools_by_allowed,
     get_tools_with_resilience,
 )
+from dynamic_agents.services.tool_error_handling import (
+    wrap_tools_with_error_handling,
+)
 from dynamic_agents.services.stream_events import (
     make_input_required_event,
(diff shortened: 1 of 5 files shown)


artdroz:
Recombining into a single PR; dynamic agents fix moving to openshell-sandbox-session-tools branch

PR #1051, opened 2026-03-28
fix(supervisor): single-node all-in-one mode — agent routing, persistence, and unified deep agent

sriaradhyula:
## Summary

This PR fixes a series of issues with the **single-node (all-in-one) supervisor mode** and unifies the multi-node and single-node implementations into a single codebase.

#### 1. Agents reporting "currently unavailable" in single-node mode
All MCP-based agents (Jira, ArgoCD, Confluence, Backstage, Komodor, PagerDuty, Slack, Splunk, Webex) were resolving their MCP server host to `localhost:8000` instead of their actual sidecar containers. Root cause: `mcp_config.py` lookup order is `<AGENT>_MCP_HOST` → `MCP_HOST` → `localhost`, but the all-in-one `docker-compose.dev.yaml` was missing per-agent host mappings. Fixed by adding explicit `<AGENT>_MCP_HOST` env vars to the `caipe-supervisor-all-in-one` service.

#### 2. Slack bot streaming final response leak
The final streamed `[FINAL ANSWER]` metadata was being forwarded to the Slack channel as a raw message. Fixed by filtering structured response markers before posting.

#### 3. Orphaned tool call metadata leak in structured response mode
After context compression or interrupted subagent calls, `AIMessage` entries with dangling tool calls remained in checkpoint history. On recovery, injecting a placeholder `AIMessage` via `as_node="model"` prevents `UnboundLocalError` in `model_to_tools` routing.

#### 4. Missing `_NullAgentRegistry` for single-node startup
When `CAIPE_DISTRIBUTED_AGENTS=false`, the platform registry module attempted remote connectivity checks at import time. Added `_NullAgentRegistry` stub that satisfies the registry interface with no-ops.

#### 5. MongoDB checkpoint persistence not wired in single-node
`deep_agent_single.py` was hardcoded to `InMemorySaver()`, ignoring `LANGGRAPH_CHECKPOINT_TYPE` and `LANGGRAPH_CHECKPOINT_MONGODB_URI`. Fixed by calling `create_checkpointer()` (same as multi-node).

#### 6. Redis cross-thread store not wired in single-node
`create_store()` was never called in the single-node path. Fixed; now wires `AsyncRedisStore` via `langgraph-redis` container with `LANGGRAPH_STORE_TYPE=redis`.

#### 7. UI debug routes left in production build
`/api/debug/auth-status` and `/api/debug/session` routes exposed session internals. Removed. Token expiry guard and JWT validation hardened.

[...]

PR #1072, opened 2026-03-31
fix(agent-registry): include system_prompt in subagent dict

sriaradhyula:
## Summary

- Add `system_prompt` field to the subagent dict returned by `generate_subagents()` in `agent_registry.py`
- One-line change: downstream consumers (deep agent, orchestrator) now have the subagent's prompt available alongside `name`, `description`, and `graph`

## Test plan

- [ ] Run `make test-multi-agents` — all agent registry tests pass
- [ ] Confirm subagent dict contains `system_prompt` key when registry is initialized

> Split from #1051.


sriaradhyula:
Folded into PR #1069 — system_prompt fix + regression tests are included there.

PR #1075, opened 2026-04-01
fix(rag): remove unconditional fetch_document call-to-action from search tool description

sriaradhyula:
## Problem

After `6d825e70` (feat(rag): improve MCP search tools and UI), the supervisor was calling `fetch_document` for every search result on every query, adding ~20s of unnecessary latency per request (observed: 10+ sequential `fetch_document` calls x ~2s each).

### Before 6d825e70

Each search result was built by naively truncating raw chunk text to 500 chars and appending a **conditional inline marker** directly to results that were actually cut off:

```python
text = page_content[:500]
if len(page_content) > 500:
    text += "... [truncated, use fetch_document with document_id to get full content]"
```

The model had a **per-result signal**: it could see exactly which results were truncated and knew precisely which ones warranted a fetch_document call. It only fetched when it saw the marker.

### After 6d825e70

The per-result truncation logic was replaced with format_search_result() from the new snippet_utils.py. This function does query-aware snippet extraction with term highlighting — a genuine improvement to result quality. The inline [truncated] marker was removed as **collateral damage** — the commit message makes no mention of this behavioral change.

With the inline per-result signal gone, the model treats the unconditional standing instruction in the tool description as "always call fetch_document for every result". The snippets look complete and well-formatted — the model has no way to know 90% of each document is missing — so it fetches everything.

## Options Considered

**Option A — Prompt config rule only**
Add a fetch_document — Use Sparingly rule to prompt_config.deep_agent.yaml.
Rejected as primary fix: Prompt instructions are soft signals. The model tends to trust a tool own description over system prompt rules. Two contradictory signals are unreliable.

**Option B — Replace instruction with softer wording**
Change to "Only use fetch_document if the snippet is clearly insufficient."
Rejected: The model cannot judge sufficiency from a snippet. The new format_search_result output looks complete and well-formatted by design — there is no visible signal indicating the snippet is partial.

**Option C — Remove the call-to-action entirely (this PR)**
Remove the unconditional instruction from the tool description entirely. The fetch_document tool still exists and is still described in its own docstring. The model will use it when a user explicitly requests full document content. It will not call it reflexively after every search.
Chosen: Matches the intended behavior of the original inline marker — conditional use, not unconditional.

**Option D — Disable fetch_document entirely**
Set fetch_document_enabled: false in RAG server config.
Rejected: Too aggressive. Legitimate use cases exist.

**Option E — Restore conditional inline marker in snippet_utils.py**
Re-add [content truncated] to format_search_result output when the full document is significantly larger than the snippet.
Deferred: More invasive change to snippet_utils.py. Could be a follow-up if the model still over-fetches in edge cases.

## The Fix

One line in _make_search_description():

- Before: Each result includes text_content (truncated to 500 chars) ... Use fetch_document with document_id to get full content.
- After: Each result includes text_content (highlighted snippet) — accurate description of what format_search_result() actually returns.

## Related

- 6d825e70 — commit that introduced the regression (collateral damage from snippet refactor)
- PR #1074 — companion fix excluding format_markdown tool in structured response mode (same class of problem: unnecessary tool calls adding latency)
- Supervisor prompt_config.deep_agent.yaml also has a fetch_document — Use Sparingly rule added as secondary belt-and-suspenders guard

## Test Plan

- [ ] Send "show me deployment options in caipe" — verify fetch_document called 0-2 times instead of 10+
- [ ] Send "what is ArgoCD?" — verify no fetch_document calls (search snippets sufficient)
- [ ] Explicitly ask "give me the full content of document X" — verify fetch_document is still called when needed
- [ ] Verify response quality is unchanged

### DIFF
--- a/ai_platform_engineering/integrations/slack_bot/tests/test_ai_plan_streaming.py
+++ b/ai_platform_engineering/integrations/slack_bot/tests/test_ai_plan_streaming.py
@@ -435,9 +435,12 @@ class TestToolEndNoTaskUpdate:
 
 
 class TestStreamWithoutPlanStreamsMarkdown:
-    """When no plan exists, STREAMING_RESULT streams markdown_text normally."""
+    """When no plan exists, STREAMING_RESULT is buffered (not streamed live) to avoid
+    leaking intermediate tool outputs or response metadata into Slack.
+    The clean FINAL_RESULT is delivered via stopStream instead."""
 
-    def test_markdown_streamed_without_plan(self):
+    def test_no_plan_streaming_result_buffered_final_result_in_stop_chunks(self):
+        """No-plan flow: STREAMING_RESULT events are buffered; FINAL_RESULT goes in stopStream."""
         events = [
             _task_event(),
             _streaming_result("Hello "),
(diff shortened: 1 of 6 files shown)


subbaksh (approved):

sriaradhyula:
resolved by another PR #1151

PR #1106, opened 2026-04-02
fix(dev): local dev no-SSO mode, Jira ADF descriptions, Docker networking

cisco-erilutz:
## Summary

Bug fixes for local development and AWS Bedrock deployments using restricted IAM roles. These are general fixes that apply to any deployment.

### 1. Jira MCP: Parse ADF description format

**Files**: \`ai_platform_engineering/agents/jira/mcp/mcp_jira/models/jira/issue.py\`, \`mcp_jira/tools/jira/search.py\`

Jira Cloud API v3 returns the \`description\` field as an Atlassian Document Format (ADF) JSON object, not a plain string. The model was storing the raw dict, making descriptions unreadable in tool responses. Added \`is_adf_format\` / \`adf_to_text\` conversion so descriptions render as human-readable text regardless of whether the Jira instance returns ADF or plain strings.

Also replaced bare \`print()\` debug statements in \`search.py\` (which included full request headers containing Base64-encoded credentials) with \`logger.debug()\` calls, and set the default log level to \`WARNING\` so they are silent unless explicitly enabled.

### 2. No-SSO local dev: anonymous fallback user gets admin role

**Files**: \`ui/src/lib/api-middleware.ts\`, \`ui/src/app/(app)/admin/page.tsx\`, \`ui/src/app/api/dynamic-agents/chat/{start-stream,cancel,resume-stream,restart-runtime}/route.ts\`

When \`SSO_ENABLED=false\`, the anonymous fallback user was given \`role: 'user'\` instead of \`role: 'admin'\`, causing the admin page to load blank and all admin API routes to return 403. Fixed to return \`role: 'admin', canViewAdmin: true\` for the anonymous local-dev user.

\`withAuth\` and \`getAuthFromBearerOrSession\` now pass \`allowAnonymous: !ssoEnabled\` — making the intent self-documenting at the call site. The anonymous fallback only fires when \`SSO_ENABLED=false\`; in SSO-enabled deployments \`allowAnonymous\` evaluates to \`false\` and the anonymous path is dead code.

The four dynamic-agents chat routes similarly pass \`allowAnonymous: !getServerConfig().ssoEnabled\` explicitly.

The admin page \`useEffect\` hooks now skip the \`status === "authenticated"\` guard when SSO is disabled, so admin data loads on mount.

### 3. Docker networking: A2A_BASE_URL defaults to internal service name; \`caipeUrl\` / \`getInternalA2AUrl\` split

**Files**: \`docker-compose.dev.yaml\`, \`ui/src/app/api/agents/tools/route.ts\`, \`ui/src/lib/config.ts\`

The \`caipe-ui\` container's \`A2A_BASE_URL\` defaulted to \`http://localhost:8000\`, which resolves to the UI container itself rather than \`caipe-supervisor\`. Fixed to \`http://caipe-supervisor:8000\`.

\`caipeUrl\` in \`getServerConfig()\` was being computed from \`A2A_BASE_URL\`, which is the internal Docker service URL — but \`caipeUrl\` is embedded into \`window.__APP_CONFIG__\` and sent to the browser, which can't reach internal service names. Split into two separate concerns:

- \`caipeUrl\` (browser-facing, in \`__APP_CONFIG__\`) now reads only \`NEXT_PUBLIC_A2A_BASE_URL\`, falling back to \`http://localhost:8000\`
- New \`getInternalA2AUrl()\` helper reads \`A2A_BASE_URL\`, falling back to \`http://caipe-supervisor:8000\` — use this in server-side API routes that proxy to the supervisor

The \`/api/agents/tools\` proxy route was using \`caipeUrl\` for its server-side fetch. Fixed to use \`getInternalA2AUrl()\`.

### 4. Skills volume mounted in caipe-ui container

**File**: \`docker-compose.dev.yaml\`

The \`caipe-ui\` container was missing the volume mount for \`./charts/ai-platform-engineering/data/skills\`. The UI reads skill templates from the filesystem at \`../charts/ai-platform-engineering/data/skills\` relative to its working directory — this path is not included in the Next.js image build and must be mounted at runtime.

### 5. Dockerfile: patch \`langchain_aws\` to forward \`base_model_id\` in \`_as_converse\`

**Files**: \`build/Dockerfile\`, \`ai_platform_engineering/dynamic_agents/build/Dockerfile\`

\`ChatBedrock._as_converse\` creates a \`ChatBedrockConverse\` instance without forwarding \`base_model_id\`. When the Converse API is used (e.g. \`beta_use_converse_api=True\`), the new \`ChatBedrockConverse\` instance doesn't know the base model and falls back to calling \`bedrock:GetInferenceProfile\` to resolve it. This fails for IAM roles that only grant \`bedrock:InvokeModel\` / \`bedrock-runtime:Converse\` on an application inference profile ARN.

The fix adds \`base_model_id=self.base_model_id\` to the \`ChatBedrockConverse(...)\` constructor call in \`_as_converse\`, bypassing the \`GetInferenceProfile\` lookup. The patch is applied in both the supervisor and dynamic-agents Dockerfiles at build time. Both include a pre-check that fails the build if the target line has changed upstream, and a post-check that verifies the substitution was applied.

### 6. Dynamic agents: mount AWS credentials, pass \`AWS_BEDROCK_BASE_MODEL_ID\`

**Files**: \`docker-compose.dev.yaml\`, \`docker-compose.yaml\`, \`docker-compose.single-node.dev.yaml\`

The \`dynamic-agents\` container was missing the \`~/.aws\` credential mount and \`AWS_BEDROCK_BASE_MODEL_ID\` env var needed for the langchain-aws patch to work. Added to all compose files so custom agents using AWS Bedrock work out of the box.

### 7. Dynamic agents config: MCP server examples and model cleanup

**File**: \`ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/config.yaml\`

- Removed cross-region Bedrock model entries (\`global.*\` profile IDs) that don't work with restricted IAM roles
- Added commented-out docker-compose mode MCP server entries for RAG, GitLab, and Jira — users can uncomment to enable, requires the corresponding compose profile

### 8. Fix \`.env.example\` for local dev

**File**: \`.env.example\`

- \`MONGODB_URI\` hostname corrected from \`mongodb\` to \`caipe-mongodb\` (the actual Docker Compose service name)
- \`NEXTAUTH_SECRET\` given a placeholder value so NextAuth doesn't fail on startup with an empty secret
- \`ALLOW_DEV_ADMIN_WHEN_SSO_DISABLED=true\` added so the admin panel is accessible out of the box when running without SSO

## Test plan

- [ ] Run \`docker compose --profile caipe-supervisor --profile caipe-ui up\` with \`SSO_ENABLED=false\` using the updated \`.env.example\` as a starting point
- [ ] Verify admin page loads without SSO session
- [ ] Verify dynamic agents chat works without SSO session
- [ ] Verify Jira issue descriptions render as plain text (not raw JSON) when fetched via the Jira MCP tool
- [ ] Verify no credential-containing print output appears in \`mcp-jira\` container logs during Jira searches
- [ ] Verify \`caipe-ui\` can reach \`caipe-supervisor\` at its internal Docker service name
- [ ] Build supervisor and dynamic-agents images; verify the \`langchain_aws\` patch applies cleanly
- [ ] Confirm custom agents using AWS Bedrock do not trigger \`GetInferenceProfile\` when \`AWS_BEDROCK_BASE_MODEL_ID\` is set
- [ ] Verify RAG/GitLab/Jira MCP servers appear in custom agent tool picker after uncommenting in config.yaml

### DIFF
--- a/ui/src/lib/__tests__/config.test.ts
+++ b/ui/src/lib/__tests__/config.test.ts
@@ -169,8 +169,8 @@ describe('getServerConfig', () => {
       expect(getServerConfig().ssoEnabled).toBe(false);
     });
 
-    it('should read A2A_BASE_URL', () => {
-      process.env.A2A_BASE_URL = 'https://my-supervisor:8000';
+    it('should read NEXT_PUBLIC_A2A_BASE_URL', () => {
+      process.env.NEXT_PUBLIC_A2A_BASE_URL = 'https://my-supervisor:8000';
       expect(getServerConfig().caipeUrl).toBe('https://my-supervisor:8000');
     });
 
@@ -668,10 +668,10 @@ describe('getServerConfig', () => {
   // ---------- Production defaults ----------
 
   describe('production defaults (when no A2A/RAG URL set)', () => {
-    it('should use k8s service URLs for caipeUrl in production', () => {
+    it('should use default caipeUrl when no NEXT_PUBLIC_A2A_BASE_URL is set', () => {
       process.env.NODE_ENV = 'production';
       clearEnv('A2A_BASE_URL');
-      expect(getServerConfig().caipeUrl).toBe('http://caipe-supervisor:8000');
+      expect(getServerConfig().caipeUrl).toBe('http://localhost:8000');
     });
 
     it('should use k8s service URLs for ragUrl in production', () => {
@@ -1263,7 +1263,7 @@ describe('end-to-end: layout injection → client read', () => {
     expect(getConfig('storageMode')).toBe('mongodb');
     expect(getConfig('spinnerColor')).toBe('#4ecdc4');
     expect(getConfig('supportEmail')).toBe('support@grid.cisco.com');
-    expect(getConfig('caipeUrl')).toBe('http://caipe-supervisor:8000');
+    expect(getConfig('caipeUrl')).toBe('http://localhost:8000');
 
     // Secrets must NOT be in the script
     expect(script).not.toContain('admin:secret');


ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/config.yaml
```diff
@@ -47,11 +47,6 @@ models:
     provider: "anthropic-claude"
     description: "Fast and efficient for simpler tasks"
 
-  - model_id: "global.anthropic.claude-haiku-4-5-20251001-v1:0"
```
cisco-erilutz, line 50: add this model back in please

ai_platform_engineering/multi_agents/platform_engineer/deep_agent_single.py
```diff
@@ -934,11 +934,38 @@ async def create_confluence_subagent_def(prompt_config: dict = None) -> dict:
     return await create_subagent_def(agent, "confluence", "Confluence: wiki documentation", prompt_config)
 
 
+async def create_gitlab_remote_subagent_def(prompt_config: dict = None) -> dict:
```
cisco-erilutz, line 937: can you add a comment here saying that this is just a stub subagent def at the moment and needs building out?

cisco-erilutz:
Closing to reopen with a clean comment section.

PR #1127, opened 2026-04-05
fix(ui): skip Bearer token for internal rag-server URLs

sriaradhyula:
## Summary

- The Duo OIDC access_token has a key ID absent from Duo's public JWKS. Forwarding it as Authorization: Bearer to an in-cluster rag-server causes persistent 401s on every REST call (/v1/mcp/custom-tools, /v1/datasources, etc.) — the rag-server cannot validate the token.
- The rag-server already handles this: when no auth header is present and the request comes from a trusted CIDR (TRUSTED_NETWORK_CIDRS=10.0.0.0/8), it grants TRUSTED_NETWORK_DEFAULT_ROLE. The Next.js pod always has a 10.x.x.x address.
- Add isInternalRagServer() to detect cluster-local and private-IP RAG server URLs. For those targets getRbacHeaders() omits the Authorization header, letting trusted-network auth handle it. External deployments continue to receive the Bearer token.

## Observed failure (caipe prod)

Token validation failed for provider 'ui': Key ID 'ce989860...' not found in JWKS (repeated every 2 min)
caipe-ui logs: API Error: Error [ApiError]: Unauthorized

## Test plan

- [ ] Deploy to preview; verify Knowledge Bases tab loads without 401 errors in rag-server logs
- [ ] Confirm supervisor RAG tool calls still work (unaffected — MCP calls never sent a token)
- [ ] Verify an external RAG server deployment (non-cluster URL) still receives Authorization: Bearer

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/ui/src/app/api/rag/[...path]/route.ts
+++ b/ui/src/app/api/rag/[...path]/route.ts
@@ -3,27 +3,27 @@ import { getServerSession } from 'next-auth';
 import { authOptions } from '@/lib/auth-config';
 
 /**
- * RAG API Proxy with JWT Bearer Token Authentication
+ * RAG API Proxy
  *
- * Proxies requests from /api/rag/* to the RAG server with JWT authentication.
- * The RAG server validates the JWT token and fetches user claims (email, groups)
- * from the OIDC userinfo endpoint, caching them in Redis.
+ * Proxies requests from /api/rag/* to the RAG server.
  *
- * Authentication:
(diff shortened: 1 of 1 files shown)


subbaksh (changes requested):
Doesnt this mean everyone is knowledge base admin?
Since UI backend is within trusted network and proxies request with no auth header

sriaradhyula:
Abandoning this fix. This is not a correct fix

Issue #682, opened 2026-01-26
Implement RBAC for RAG Retrieval Chunk Scopes


sriaradhyula:
## Overview
Implement Role-Based Access Control (RBAC) for RAG retrieval to enforce chunk-level scoping and access validation during inference time.

## Objective
Ensure that RAG retrieval respects user permissions and only returns chunks that the user is authorized to access.

## Key Components
- Define chunk scope boundaries and ownership
- Implement RBAC policies for chunk access
- Add access validation during retrieval
- Enforce scoping at inference time
- Audit and logging for access control

## Success Criteria
- RBAC policies enforced for all chunk retrievals
- Unauthorized access attempts logged and blocked
- Performance impact minimal
- Comprehensive test coverage

PR #672, opened 2026-01-22
feat: RAG RBAC

subbaksh:
# Description

 - **RBAC Backend:** Implemented role-based authorization (admin, ingestor, viewer) with permission checks on all server endpoints
 - **User Authentication:** Integrated OAuth2 proxy for user authentication with role extraction from headers
 - **UI Enhancements:** Added user profile display, role visibility, and permission-aware UI components
 - **Configuration:** Updated docker-compose with OAuth2 proxy service and RBAC-related environment variables

## Type of Change

- [x] New Feature
- [ ] Bugfix
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Pre-release Helm Charts (Optional)

For chart changes, you can test pre-release versions before merging:
- **Base repo contributors:** Create a branch starting with `pre/` for automatic pre-release builds
- **Fork contributors:** Ask a maintainer to add the `helm-prerelease` label
- Pre-release charts are published to `ghcr.io/cnoe-io/pre-release-helm-charts`
- Cleanup happens automatically when the PR closes or label is removed

## Checklist

- [x] I have read the [contributing guidelines](CONTRIBUTING.md)
- [ ] Existing issues have been referenced (where applicable)
- [ ] I have verified this change is not present in other open pull requests
- [ ] Functionality is documented
- [ ] All code style checks pass
- [ ] New code contribution is covered by automated tests
- [ ] All new and existing tests pass

PR #1138, opened 2026-04-06
feat(rag): comprehensive RBAC with userinfo-based auth fallback

sriaradhyula:
## Problem

The rag-server's comprehensive RBAC system (merged in #672) is fully implemented — it fetches user identity from the OIDC userinfo endpoint, extracts email and groups, and maps them to roles (admin/ingestor/viewer) with Redis caching. **None of this works in prod because JWKS validation fails before userinfo is ever called.**

### Auth flow in `rbac.py` (`require_authenticated_user`):

```
1. validate_token() via JWKS       ← FAILS for Duo access_token
2. fetch_userinfo() → email/groups ← never reached
3. map groups → role               ← never reached
4. cache in Redis                  ← never reached
```

### Why JWKS fails

Duo issues access tokens signed with a key whose ID (`ce989860...`, a 64-char hex SHA-256) is **not published** in Duo's public JWKS for the configured client. This is expected — access tokens are intended for Duo's own resource servers, not for third-party JWKS validation. The rag-server's attempt to validate them via JWKS is architecturally incorrect for this token type.

### Consequences in prod (plat-prod-use2-1)

- Every caipe-ui → rag-server REST call returns `401 Unauthorized`
- Knowledge Bases tab is completely broken for all users
- The supervisor works only because it uses MCP (no auth header → trusted network grants admin)
- PR #1127 worked around this by stripping the auth header entirely for internal URLs — but this bypasses the RBAC system and gives all UI users the same `TRUSTED_NETWORK_DEFAULT_ROLE` regardless of their actual group membership

## Root Cause

`validate_token()` in `auth.py` is a hard prerequisite for the userinfo fetch. There is no fallback: if JWKS validation fails, the request is rejected with 401 immediately.

```python
# rbac.py — require_authenticated_user()
provider, access_claims = await auth_manager.validate_token(token)  # hard fail
# ...
userinfo = await auth_manager.fetch_userinfo(token, provider)       # never reached
```

## Correct Fix

In `auth.py`, when JWKS validation fails, fall back to the userinfo endpoint directly. If the OIDC provider's userinfo endpoint accepts the token and returns valid claims, the token is implicitly valid — the provider has verified it server-side.

```python
async def validate_token(self, token: str) -> Tuple[OIDCProvider, Dict[str, Any]]:
    for provider in self.providers.values():
        try:
            claims = await provider.validate_token(token)
            return provider, claims
        except JWTError:
            # JWKS failed — try userinfo fallback (handles opaque/non-JWKS-validatable tokens)
            try:
                userinfo = await provider.fetch_userinfo(token)
                return provider, userinfo
            except Exception:
                continue
    raise JWTError("Token validation failed for all providers")
```

With this fix:
- Duo access_token → JWKS fails → userinfo succeeds → email + groups returned
- Group-to-role mapping works: user in `caipe-admins` → admin, others → viewer
- Per-user RBAC is fully restored
- PR #1127 can be reverted (the auth header should be forwarded, not stripped)

## Related

- #682 — RBAC for RAG Retrieval Chunk Scopes (depends on per-user identity working)
- #672 — feat: RAG RBAC (implemented the userinfo + group mapping, blocked by this issue)
- PR #1127 — workaround that should be reverted once this is fixed
- PR #815 — fix(rag): prioritize JWT auth over trusted network


subbaksh (changes requested):
> Duo issues access tokens signed with a key whose ID (ce989860..., a 64-char hex SHA-256) is not published in Duo's public JWKS for the configured client. This is expected — access tokens are intended for Duo's own resource servers, not for third-party JWKS validation. The rag-server's attempt to validate them via JWKS is architecturally incorrect for this token type.

I dont think this is true, you can test it using the production duo issuer url + client_id and get the access_token from the UI and validate. The problem might be somewhere else. Lets discuss this more tomorrow

sriaradhyula:
Closing — branch has been rebased onto main and force-pushed to `release/0.5.0`. This PR targeting `release/0.4.0` is superseded.

PR #1201, opened 2026-04-13
fix(supervisor): add get_skills_status() and _rebuild_graph() to DeepAgent

sriaradhyula:
## Summary

- The skills middleware router expects `get_skills_status()` and `_rebuild_graph()` on the MAS instance for FR-016 and FR-012 endpoints
- `PlatformEngineerDeepAgent` was missing both methods, causing `hasattr` checks to fail
- `/internal/supervisor/skills-status` returned all nulls with `mas_registered: false`, making the UI Supervisor skills load panel show Status unavailable
- `/skills/refresh` silently skipped graph rebuild

## Changes

- Added `_skills_loaded_count`, `_skills_merged_at`, `_last_built_catalog_generation` tracking attributes to `__init__`
- Track skills count and timestamp during `_build_graph_async()` after catalog merge
- Added `get_skills_status()` method returning the exact dict shape `router.py` expects, with `sync_status` computed from catalog cache generation
- Added sync `_rebuild_graph()` wrapper that delegates to `_rebuild_graph_async()`, handling the case where called from within a running event loop (FastAPI context)

## Test plan

- [ ] Verify `/internal/supervisor/skills-status` returns real metadata instead of null fallback
- [ ] Verify `/skills/refresh` triggers graph rebuild and returns `supervisor_rebuilt: true`
- [ ] Verify UI Supervisor skills load panel displays skills count, generation, and sync status
- [ ] Verify no regression in existing `get_status()` or `_rebuild_graph_async()` behavior

Assisted-by: Claude:claude-opus-4-6

### DIFF
--- a/ai_platform_engineering/multi_agents/platform_engineer/deep_agent.py
+++ b/ai_platform_engineering/multi_agents/platform_engineer/deep_agent.py
@@ -1100,6 +1100,9 @@ class PlatformEngineerDeepAgent:
         self._graph_lock = threading.RLock()
         self._graph = None
         self._graph_generation = 0
+        self._skills_loaded_count: int = 0
+        self._skills_merged_at: Optional[str] = None
+        self._last_built_catalog_generation: int = 0
         self._initialized = False
         self._subagent_tools: Dict[str, List[str]] = {}
         self._distributed_agents = _get_distributed_agents()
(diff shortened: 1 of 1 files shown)


sriaradhyula:
Closing — changes cherry-picked into prebuild/refactor/skills-gallery-simplification branch.

PR #1229, opened 2026-04-16
feat(auth): add dual auth middleware for shared key + OAuth2 coexistence

artdroz:
## Problem

When `A2A_AUTH_SHARED_KEY` is set in `main.py`, the auth middleware selection uses an if/elif chain that **overrides** OAuth2, breaking the UI's OIDC-based backend calls.

## Solution

Adds a `DualAuthMiddleware` that supports both auth methods simultaneously:

1. **New file:** `ai_platform_engineering/utils/auth/dual_auth_middleware.py`
   - Checks if bearer token matches `A2A_AUTH_SHARED_KEY` → allows immediately (using `hmac.compare_digest` to prevent timing attacks)
   - Otherwise falls back to `verify_token()` from `oauth2_middleware.py` for JWT/JWKS validation
   - Same public_paths and CORS preflight handling as existing middlewares
   - Lazy import of OAuth2 verifier to avoid premature env validation at module load

2. **Modified:** `main.py` auth selection now has a new top-level branch:
   - `if A2A_AUTH_SHARED_KEY and A2A_AUTH_OAUTH2` → DualAuthMiddleware (new)
   - `elif A2A_AUTH_SHARED_KEY` → SharedKeyMiddleware (unchanged)
   - `elif A2A_AUTH_OAUTH2` → OAuth2Middleware (unchanged)
   - `else` → no auth (unchanged)

3. **Tests:** `tests/test_dual_auth_middleware.py` covering both auth paths, rejection cases, public paths, and CORS preflight

## Testing

```bash
uv run pytest tests/test_dual_auth_middleware.py -v
```

### DIFF
--- a/ai_platform_engineering/multi_agents/platform_engineer/protocol_bindings/a2a/main.py
+++ b/ai_platform_engineering/multi_agents/platform_engineer/protocol_bindings/a2a/main.py
@@ -200,7 +200,15 @@ app.mount("/", _skills_api)
 A2A_AUTH_OAUTH2 = os.getenv('A2A_AUTH_OAUTH2', 'false').lower() == 'true'
 A2A_AUTH_SHARED_KEY = os.getenv('A2A_AUTH_SHARED_KEY')
 
-if A2A_AUTH_SHARED_KEY:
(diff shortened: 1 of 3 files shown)


sriaradhyula (commented):
LGTM

PR #1240, opened 2026-04-17
feat(slack-bot): enforce 50-block limit, clean plan cards, keepalive timer, and channel context

subbaksh:
# Description

Fixes the Slack 50-block limit rejection, cleans up plan card UX, adds overthink keepalive timer, and enriches agent context with channel info. Part of the AG-UI migration for release 0.4.0.

Related: 100 (Slack AG-UI migration)

## Changes

- **50-block limit enforcement** — `enforce_block_limit()` truncates blocks and appends a warning before calling `chat.update`, preventing Slack API rejections
- **Clean plan cards** — Plan cards show only todo titles with status badges; no details, output, or tool thoughts (all logged at DEBUG instead)
- **Overthink keepalive timer** — Self-rescheduling `threading.Timer` cycles through keepalive messages every 60s during extended thinking
- **Status rate limiting** — `_STATUS_RATE_LIMIT_SECS = 1.0` prevents flooding Slack API during rapid status updates
- **Confidence marker stripping** — `_strip_confidence_markers()` runs at every exit point
- **Channel context** — Channel topic/purpose fetched via `conversations.info`, cached 1hr, passed in `client_context`
- **GlobalDefaults fix** — `Config.from_env()` now reads `SLACK_INTEGRATION_DM_AGENT_ID` and `SLACK_INTEGRATION_DEFAULT_AGENT_ID`
- **System message filter** — Q&A handler skips events with subtypes (`channel_purpose`, `channel_topic`, etc.)
- **DM agent guard** — DM handler guards against missing `agent_id` from GlobalDefaults

## Type of Change

- [x] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] All new and existing tests pass (168 passing)


ai_platform_engineering/integrations/slack_bot/utils/slack_formatter.py
```diff
@@ -125,6 +119,46 @@ def split_text_into_blocks(text: str, max_length: int = 3000) -> List[str]:
...
+  "elements": [
+    {
+      "type": "mrkdwn",
+      "text": "_Response truncated — exceeded Slack's 50-block limit._",
+    }
+  ],
+}
+
+
+def enforce_block_limit(
```
kevkantes, line 135: This to me is a bit of a red flag. Each block has 3000 char limit, 50 blocks would be 150k chars, which google tells me is 1000-1500 sentences. No response should be anywhere near that big, it seems like something is being sent back incorrectly.

In addition, blocks are just a fallback for streaming failures. However, at this point everything should be using streaming, you shouldn't need to hit blocks

Not saying this safety check isn't completely unnecessary, but definitely not something we should be hitting currently.

ai_platform_engineering/integrations/slack_bot/app.py
```diff
@@ -383,6 +396,10 @@ def handle_dm_message(event, say, client):
       context_message = slack_context.build_thread_context(app, event.get("channel"), thread_ts, message_text, bot_user_id)
 
     agent_id = _get_agent_id_for_dm()
+    if not agent_id:
```
kevkantes, line 399: Why doesn't DMs just use the same agent as the default? Like if no agent is supplied, it uses whatever spins up as default

Part of this migration should also be the removal of each channel having custom prompts and instead having a custom agent_id in the config. 

so currently we have qanda and ai_alerts, the difference being bot logic. Instead, we can probably just have `agent`, `agent.agent_id` with like `agent.bots_enabled` and `agents.bots_enabled.bots_list` (there is bot_list logic in qanda already). 

Maybe long term we support multiple agents per channel

PR #1260, opened 2026-04-20
fix(rag): SDPL-1673 fix RAG loop recursion errors and empty responses

cisco-erilutz:
## Bug story

This PR fixes two interacting bugs that produced either **LangGraph recursion limit crashes** or **empty responses** when the agent searched a RAG knowledge base.

---

### Stage 1 — Original code: `jump_to="end"` fires too early → empty response

The original `DeterministicTaskMiddleware.after_model` detected RAG cap exhaustion and immediately returned `jump_to="end"`, terminating the graph before the LLM had a turn to synthesize its findings into a response.

| Step | What happened |
|------|--------------|
| LLM calls `search` ×5 | Cap fires, middleware injects blocked ToolMessages |
| `jump_to="end"` | Graph exits immediately — LLM never gets a synthesis turn |
| **User sees: empty response** | |

Fixed in PR #1231 (SDPL-1601) by removing `jump_to="end"` entirely. But that introduced the next bug.

---

### Stage 2 — After PR #1231: soft cap message misread as empty result → recursion crash

Without `jump_to="end"`, the middleware injected a ToolMessage when the cap was hit, but the cap message read:

> *"All relevant results have been collected. Do NOT call search again."*

This sounds like a **successful search that found nothing new** — not a hard stop. The LLM interpreted it as "this query came back empty, let me try a different angle" and called search again with a slightly different query. Every subsequent call got the same message, and the LLM kept trying.

From a real prod trace (`53f70a9565f45fc21b43433f0af7b571`), the LLM ran **70 LLM turns over 4.5 minutes**, cycling through 58 query variations on the same 4 topics before hitting the 500-step recursion limit:

| Turn | Time | Tool | Actually ran? | Query |
|------|------|------|:---:|-------|
| 1 | 16:23:01 | search ×5 | ✅ | GlobalProtect access tmp.okta.com, VPN troubleshooting... |
| 2 | 16:23:09 | search, fetch ×2 | ✅ | request tmp okta access new employee |
| 3–13 | 16:23:19–49 | search | ✅ | Various GlobalProtect / okta queries |
| **14** | **16:24:00** | **search** | **❌ cap hit** | tmp okta account not provisioned |
| 15 | 16:24:04 | search | ❌ cap hit | GlobalProtect VPN tunnel gateway selection |
| 16–70 | 16:24:08–16:27:36 | search | ❌ cap hit | *58 more variations on the same 4 topics* |
| 71 | — | — | 💥 recursion limit | |

**None of turns 14–70 returned real data.** Every one received the soft cap message string. The LLM was not getting rewarded — it was misreading the cap message as an empty search result and trying harder.

There were also two compounding bugs that made the cap ineffective:

1. **Per-child-graph counters**: deepagents spawns child graphs with a fresh `thread_id`, so each child had its own independent cap counter starting at 0. With 2 concurrent RAG child graphs, 10 real searches ran instead of 5.
2. **Parallel bypass race condition**: when the LLM emits N parallel `search` calls, LangGraph fans them as concurrent async tasks. All N read `count=0` before any increment lands, bypassing the cap entirely.

---

### Stage 3 — This PR: shared caps, pre-dispatch gate, hard stop messages

**Three fixes working together:**

**1. Pre-dispatch batch gate in `after_model`**
Cap enforcement moves to `after_model`, which runs synchronously before LangGraph fans tool calls out in parallel — eliminating the race condition. `rag_batch_cap_check()` takes an atomic snapshot of remaining budgets under both locks. When the LLM requests more calls than the remaining budget, all calls for that tool type are blocked and the LLM is told exactly how many slots remain.

**2. Shared cap bucket via ContextVar**
`_rag_conversation_id` is set once at stream start and propagates automatically into all child graphs via Python's async context inheritance. All child graphs now share the same cap counter as the parent.

**3. Unambiguous cap messages**
When the budget is fully exhausted:

> *"RAG call limit reached. Do NOT call search or fetch_document again. You MUST now synthesize your final answer from the information already retrieved above."*

When budget remains but the LLM requested too many at once, it's told exactly how many calls it has left:

> *"This RAG call was blocked because more queries were requested than are left under the limit. You still have: 3 search calls remaining. Use those remaining calls to gather any additional information needed, then synthesize your final answer."*

Additionally, when one tool's cap fires inside `_arun`, the message now includes the remaining budget for the *other* tool (e.g. search cap fires → message says how many `fetch_document` calls remain), rather than incorrectly showing both as exhausted.

**Result:** the LLM receives an accurate, unambiguous signal — not a message that looks like an empty search result. In testing, the LLM immediately paused to check available datasources before making 2 targeted searches, then synthesized a real answer.

**Conditional `jump_to="end"` (non-structured response mode only)**
After the LLM receives the synthesis message and calls an exhausted RAG tool a second time, `jump_to="end"` fires to hard-stop the graph. In `USE_STRUCTURED_RESPONSE` mode (the default in prod) this is intentionally omitted — the LLM still needs a turn to call `PlatformEngineerResponse`, and skipping it produces an empty response (the same Stage 1 regression). The hard message is sufficient in practice; `jump_to="end"` is the backstop for non-structured mode.

**Fixed flow (from live trace `8ee28dd403936f692f5b465e09ef3ccf`):**

| Time | LLM call | What happened |
|------|----------|---------------|
| 16:47:04 | `write_todos` (112 tok) | Planned 2 RAG tasks |
| 16:47:08 | `search` ×5 (339 tok) | Cap fires (budget=3), all 5 blocked — "3 remaining, re-strategize" |
| 16:47:15 | `list_datasources_and_entity_types` (54 tok) | LLM checks available datasources before spending budget |
| 16:47:18 | `search` ×2 (192 tok) | Uses 2 of 3 remaining slots with targeted queries |
| 16:47:22–33 | `fetch_document` ×3 | Fetches documents from results |
| 16:47:39 | `PlatformEngineerResponse` (738 tok) | `was_task_successful=True` — real answer |

---

## Summary of changes

- `rag_tools.py`: `rag_batch_cap_check()` / `RagBatchCapResult`, `rag_cap_message()` with accurate remaining counts (including cross-tool remaining budget in `_arun` cap path), `_rag_conversation_id` ContextVar, `get_max_calls()` / `get_call_count_for_thread()` classmethods, unified hard-stop message, fix lock ordering, remove dead `_RagToolCapExhausted` class
- `middleware.py`: pre-dispatch `after_model` gate; `_rag_terminal_response()` omits `jump_to="end"` in structured response mode to prevent empty responses
- `a2a/agent.py`: `set_rag_conversation_id()` at stream start; promote imports to top-level
- `tests/test_middleware_rag_loop_exit.py`: 10 tests covering all cap/termination scenarios

## Test plan

- [x] `pytest tests/test_middleware_rag_loop_exit.py` — 10/10 pass
- [x] `make lint` — clean
- [x] Live trace confirmed: cap fires correctly, LLM re-strategizes with remaining budget, `was_task_successful=True`, 738-token response

Fixes SDPL-1673


ai_platform_engineering/multi_agents/platform_engineer/rag_tools.py
```diff
@@ -212,10 +220,12 @@ async def _arun(self, **kwargs: Any) -> str:
     if capped is not None:
       logger.warning(f"search cap ({self.max_calls}) reached for thread_id={thread_id}")
       _record_rag_cap_hit(thread_id, "search")
-      return (
-        f"[Search complete] You have performed {self.max_calls} searches which is the maximum allowed. "
-        "All relevant results have been collected. Do NOT call search or fetch_document again. "
-        "You MUST now synthesize your final answer using ONLY the information already retrieved above."
+      raise _RagToolCapExhausted(
```
cisco-erilutz: if this is "cap reached but we do return a search" this should probably be a success and not an error

ai_platform_engineering/multi_agents/platform_engineer/rag_tools.py
```diff
@@ -150,13 +159,12 @@ async def _arun(self, document_id: str, thought: str = "", **kwargs: Any) -> str
...
       _record_rag_cap_hit(thread_id, "fetch_document")
-      # Return a normal-looking result so the model doesn't retry with different
-      # document_ids. Raising ToolInvocationError creates is_error=True ToolMessages
-      # which the model treats as "this doc failed, try the next" — causing a loop.
-      return (
-        f"[Document already retrieved] You have fetched {self.max_calls} documents which is the maximum allowed. "
-        "All relevant content has been collected. Do NOT call fetch_document or search again. "
-        "You MUST now synthesize your final answer using ONLY the documents already retrieved above."
+      raise _RagToolCapExhausted(
+        "No more documents available. The knowledge base has been fully searched. "
```
cisco-erilutz: don't imply the knowledge base has been fully searched, we just hit the cap

ai_platform_engineering/multi_agents/platform_engineer/rag_tools.py
```diff
@@ -144,20 +182,24 @@ def from_tool(cls, original: Any, max_calls: int = _DEFAULT_MAX_FETCH_DOCUMENT_C
...
-      # document_ids. Raising ToolInvocationError creates is_error=True ToolMessages
-      # which the model treats as "this doc failed, try the next" — causing a loop.
-      return (
-        f"[Document already retrieved] You have fetched {self.max_calls} documents which is the maximum allowed. "
-        "All relevant content has been collected. Do NOT call fetch_document or search again. "
-        "You MUST now synthesize your final answer using ONLY the documents already retrieved above."
-      )
+      search_used = SearchCapWrapper.get_call_count_for_thread(thread_id)
+      search_remaining = max(0, SearchCapWrapper.get_max_calls() - search_used)
+      return rag_cap_message(search_remaining=search_remaining)
```
cisco-erilutz: in cases like these we should probably pass in the fetch_document limits too right? since otherwise it defaults to zero and we might get the "rag cap reached" message when it should say that there are fetch document calls left.

There might be other places in the code like this too, it seems like we always need to pass in both vars, we should probably make them positional and not kwargs with defaults

ai_platform_engineering/utils/deepagents_custom/middleware.py
```diff
@@ -90,6 +90,28 @@ def _generate_tool_call_id() -> str:
     return f"call_{uuid.uuid4().hex[:24]}"
 
 
+def _rag_terminal_response(tool_messages: list) -> dict:
+    """Build the terminal state update when RAG caps are fully exhausted.
+
+    In USE_STRUCTURED_RESPONSE mode the graph must not jump_to=end because
```
cisco-erilutz, line 96: can we look into this? I thought it was the opposite?

cisco-erilutz: Good question — the logic is: in `USE_STRUCTURED_RESPONSE=True` mode (prod default), after the RAG hard-stop fires the LLM still needs one more turn to call `PlatformEngineerResponse` to produce the actual response. If we return `jump_to="end"` immediately, the graph terminates before that turn and the response assembler finds no AIMessage with text content → empty response. That's the Stage 1 regression (PR #1231).

So in structured mode we omit `jump_to="end"` and let the LLM take its synthesis turn naturally. In plain-text mode `jump_to="end"` is safe because the response assembler reads the last prose AIMessage directly from state — no tool call needed.

ai_platform_engineering/multi_agents/platform_engineer/rag_tools.py
```diff
@@ -26,15 +27,57 @@
...
+    when the stream ends (important in environments that reuse async contexts).
+    """
+    return _rag_conversation_id.set(conversation_id)
+
+
 _DEFAULT_MAX_FETCH_DOCUMENT_CALLS = int(os.getenv("RAG_MAX_FETCH_DOCUMENT_CALLS", "5"))
 _DEFAULT_MAX_SEARCH_CALLS = int(os.getenv("RAG_MAX_SEARCH_CALLS", "5"))
 _STALE_ENTRY_TTL_SECONDS = 300
 
+def rag_cap_message(search_remaining: int = 0, fetch_remaining: int = 0) -> str:
```
cisco-erilutz, line 55: do we need this at all? or could it just be subsumed into rag_cap_message_for_thread?

cisco-erilutz: The middleware `after_model` batch gate still calls `rag_cap_message` directly (line 346) — it already has precise `search_remaining`/`fetch_remaining` values from `RagBatchCapResult` so it doesn't need to re-query thread state. `rag_cap_message_for_thread` is the convenience wrapper for the `_arun` cap paths that don't have pre-computed values. So the split is: `rag_cap_message` is the primitive for callers that already know the counts, `rag_cap_message_for_thread` is for callers that need to look them up. Happy to merge them if you'd prefer, but it would mean `after_model` has to pass a thread_id and ignore its already-computed counts.

sriaradhyula:
@cisco-erilutz @ErikMLutz - close this PR?

PR #1306, opened 2026-04-28
fix(migration): handle stringified artifact dicts and passthrough event types in 0.4.0 a2a migration

kevkantes:
# Description

Fixes two bugs in `scripts/migrations/0.4.0/migrate_messages_to_turns.py` discovered by inspecting the actual dev DocumentDB data before running the 0.4.0 migration.

**Bug 1 — `artifact` field is a stringified Python repr, not a sub-document.**
Legacy A2A event documents stored the `artifact` field as a Python `repr()` string (e.g. `"{'name': 'streaming_result', 'parts': [...]}"`) rather than a proper nested object. The original code called `.get()` directly on the raw value, which throws `AttributeError: 'str' object has no attribute 'get'` for the 29,390+ artifact-type events in the DB and crashes the entire migration for any conversation containing them. Fixed by adding `_parse_artifact()` which safely deserializes via `ast.literal_eval()`.

**Bug 2 — Type mapping missed already-normalized top-level event types.**
The DB contains events with top-level `type` values of `"tool_start"`, `"tool_end"`, `"execution_plan"`, `"task"`, and `"status"` that are already in the correct normalized form. The original code routed everything through `A2A_ARTIFACT_TYPE_MAP` (keyed on `artifact.name`), causing all of these to fall through to `"a2a_raw"`. Fixed by adding `A2A_EVENT_TYPE_PASSTHROUGH` — these types are passed through directly; only `"artifact"`-type events go through the artifact name lookup.

Also includes the `stable_event_id` key to include `raw_type` so passthrough events and artifact events with the same sequence number get distinct deterministic IDs.

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/scripts/migrations/0.4.0/migrate_messages_to_turns.py
+++ b/scripts/migrations/0.4.0/migrate_messages_to_turns.py
@@ -45,6 +45,7 @@ Usage:
 """
 
 import argparse
+import ast
 import hashlib
 import os
 import sys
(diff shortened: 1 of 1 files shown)


subbaksh (approved):
Tyvm!

kevkantes:
> Tyvm!

Np, want me to merge? Or you can feel free. Likely don't need a tag for this

subbaksh:
Yes go for it!

PR #1307, opened 2026-04-28
fix(rag): stop repeated structured-response cap loops

artdroz:
## Summary

- Backports the SDPL-1673 structured-response RAG recursion fix onto current main/0.4.x.
- Tracks when a thread has already received its post-cap synthesis turn.
- If the model ignores that nudge and calls capped RAG tools again, the middleware now forces jump_to=end instead of looping until LangGraph recursion exhaustion.

## Evidence

The existing SDPL-1673 trace analysis on PR #1303 showed 213 LLM generations, 6 real tool executions, and 203 empty search loops before recursion exhaustion. CAIPE dev tracks current platform-apps HEAD / chart 0.4.2, which does not include that second-turn force-exit behavior.

## Test plan

- [x] python3 -m py_compile ai_platform_engineering/multi_agents/platform_engineer/rag_tools.py ai_platform_engineering/utils/deepagents_custom/middleware.py tests/test_middleware_rag_loop_exit.py
- [ ] pytest tests/test_middleware_rag_loop_exit.py -q locally blocked: this host only has Python 3.11, while repo dependencies require Python >=3.13. CI/prebuild should run in the supported environment.

Assisted-by: OpenAI Codex

### DIFF
--- a/ai_platform_engineering/multi_agents/platform_engineer/rag_tools.py
+++ b/ai_platform_engineering/multi_agents/platform_engineer/rag_tools.py
@@ -241,6 +241,7 @@ class SearchCapWrapper(_CapCounterMixin, BaseTool):
 _rag_cap_hit_counts: dict[str, int] = {}
 _rag_capped_tools: dict[str, set[str]] = {}
 _rag_hard_stop_lock = threading.Lock()
+_rag_synthesis_turn_given: set[str] = set()
 
 
 def _record_rag_cap_hit(thread_id: str, tool_name: str = "") -> None:
@@ -267,6 +268,18 @@ def is_rag_hard_stopped(thread_id: str) -> bool:
         return bool(_rag_capped_tools.get(thread_id))
 
 
+def record_synthesis_turn_given(thread_id: str) -> None:
+    """Mark that the model has already received one synthesis turn after cap exhaustion."""
+    with _rag_hard_stop_lock:
+        _rag_synthesis_turn_given.add(thread_id)
+
+
+def was_synthesis_turn_given(thread_id: str) -> bool:
+    """Return True if the model already received a synthesis turn for this thread."""
+    with _rag_hard_stop_lock:
+        return thread_id in _rag_synthesis_turn_given
+
+
 def clear_rag_state(thread_id: str) -> None:
     """Reset RAG cap counters for a thread at the start of a new query.
 
@@ -276,6 +289,7 @@ def clear_rag_state(thread_id: str) -> None:
     with _rag_hard_stop_lock:
         _rag_capped_tools.pop(thread_id, None)
         _rag_cap_hit_counts.pop(thread_id, None)
+        _rag_synthesis_turn_given.discard(thread_id)
     with FetchDocumentCapWrapper._global_lock:
         FetchDocumentCapWrapper._global_counts.pop(thread_id, None)
         FetchDocumentCapWrapper._global_timestamps.pop(thread_id, None)
--- a/ai_platform_engineering/utils/deepagents_custom/middleware.py
+++ b/ai_platform_engineering/utils/deepagents_custom/middleware.py
@@ -90,6 +90,38 @@ def _generate_tool_call_id() -> str:
     return f"call_{uuid.uuid4().hex[:24]}"
 
 
+def _rag_terminal_response(tool_messages: list, thread_id: str = "") -> dict:
+    """Build the terminal state update when RAG caps are fully exhausted."""
+    try:
+        from ai_platform_engineering.multi_agents.platform_engineer.deep_agent import USE_STRUCTURED_RESPONSE  # noqa: PLC0415
+    except ImportError:
+        USE_STRUCTURED_RESPONSE = False
+
+    if USE_STRUCTURED_RESPONSE:
+        try:
+            from ai_platform_engineering.multi_agents.platform_engineer.rag_tools import (  # noqa: PLC0415
+                record_synthesis_turn_given,
+                was_synthesis_turn_given,
+            )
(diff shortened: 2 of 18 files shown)


artdroz:
Validation update:

- Supervisor prebuild image published: ghcr.io/cnoe-io/prebuild/ai-platform-engineering:fix-caipe-rag-recursion-main-3
- Deployed to CAIPE dev via cisco-eti/platform-apps-deployment#374, merged at 0f9bd61d
- After ArgoCD sync, https://caipe.dev.outshift.io/.well-known/agent-card.json reports version 0.4.3-dev.1
- RAG stress query against updated dev completed with a structured response and no GraphRecursionError

Notes:
- I could not reproduce the original recursion error with two synthetic regular-RAG stress prompts before the deploy; both returned successfully on 0.4.2
- The original Slack zip attachment is still inaccessible from this runtime, so this validation uses regular RAG docs only

Assisted-by: OpenAI Codex

artdroz:
[Reviewer] PASS for exact head `d0b7fc4326af74719e806bf71a21f9160a278e0d`.

Scope reviewed: CAIPE recursion fix for repeated structured-response RAG cap loops and regular-RAG behavior. Files checked include `ai_platform_engineering/multi_agents/platform_engineer/rag_tools.py`, `ai_platform_engineering/utils/deepagents_custom/middleware.py`, `tests/test_middleware_rag_loop_exit.py`, `pyproject.toml` / `uv.lock`, and chart version bumps.

Evidence:
- PR #1307 is open, non-draft, mergeable, and currently at exact head `d0b7fc4326af74719e806bf71a21f9160a278e0d`.
- GitHub checks are green for the exact head, including unit tests, Helm chart tests, Ruff/Super Linter, CodeQL, Grype filesystem scan, DCO, pinned dependency check, proprietary-content scan, prebuild supervisor image, sub-agent prebuild images, and prebuild Helm chart publish.
- `rag_tools.py` now tracks `_rag_synthesis_turn_given` per thread and clears it in `clear_rag_state`, so the structured-response escape state does not carry across new queries.
- `middleware.py` preserves the previous non-structured hard stop (`jump_to=end`) while allowing exactly one structured-response synthesis turn after capped RAG calls, then forces `jump_to=end` on a repeated capped-RAG attempt. That directly addresses the structured-response recursion loop without weakening the existing individually-capped RAG tool checks.
- Tests cover the non-structured hard-stop behavior and the new structured-response second-cap exit behavior.
- Deployment validation evidence was reviewed at https://github.com/cnoe-io/ai-platform-engineering/pull/1307#issuecomment-4336921808: prebuild image `ghcr.io/cnoe-io/prebuild/ai-platform-engineering:fix-caipe-rag-recursion-main-3` was deployed through cisco-eti/platform-apps-deployment#374, CAIPE dev reported agent card version `0.4.3-dev.1`, and a regular-RAG stress query completed with a structured response and no `GraphRecursionError`.

Caveats:
- I did not run local pytest because this runtime has Python 3.11 while the repo requires Python >=3.13; I relied on the green GitHub unit/check suite for executable validation.
- I did not inspect the original Slack zip because it is inaccessible from this runtime; validation used regular RAG docs and synthetic stress prompts as recorded by coder.

Verdict: acceptable for the stated CAIPE recursion fix scope.

artdroz:
Attachment follow-up / GREPSTORM thoughts:

I can now read the Slack `grep-bug.rtfd.zip` attachment. The important case described there is slightly different from the original structured-response RAG cap loop this PR targets:

- The RAG cap appears to have fired correctly and injected the synthesis/stop message.
- The model then escaped into remaining tools, specifically `fetch_document` and then `grep`.
- `grep` repeated with the same pattern and empty/stubby results until the graph hit the recursion limit.

My read: this PR is still useful as the narrow hotfix for repeated capped RAG calls after the one structured-response synthesis turn. It may not fully cover the GREPSTORM attachment case if `grep` is outside the capped RAG/tool-budget set, because the model can leave the capped path and loop in another tool.

Follow-up options I would consider, without changing this PR right now:

1. Add `grep` (and possibly `fetch_url` / adjacent retrieval tools) to the same budget/cap family if they are semantically part of RAG retrieval.
2. Prefer a more general per-thread tool-loop limiter that detects repeated identical tool calls with empty/no-op results and forces a terminal response. This would avoid chasing individual tool names.
3. Keep this PR scoped and land it as the immediate structured-response fix, then open a separate GREPSTORM follow-up for generic tool-loop protection.

LangChain middleware candidate:

Arthur pointed at LangChain's `ToolCallLimitMiddleware`: https://reference.langchain.com/python/langchain/agents/middleware/tool_call_limit/ToolCallLimitMiddleware

This looks like the right built-in primitive to evaluate for option 2. It tracks tool-call counts, can apply either to all tools or to a named tool via `tool_name`, supports per-run and per-thread limits, and has an `exit_behavior` including `"end"`. Given this repo is already on `langchain==1.2.15` and `langgraph==1.1.6`, it is worth testing before adding a custom loop limiter.

Suggested follow-up shape:

- Use a broad run/thread cap for total retrieval/tool calls to prevent unbounded tool churn.
- Add a tighter tool-specific cap for `grep` if GREPSTORM is the known failure mode.
- Prefer `exit_behavior="end"` or an equivalent terminal response path so the graph stops cleanly rather than raising a user-visible recursion failure.
- Add regression coverage where the model repeatedly calls `grep` with the same query and empty/no-op results.
- Verify it composes with the existing custom middleware and structured-response path; if it does not compose cleanly, use the same semantics in our custom middleware.

No code changes made here; this is just the newly available attachment context and middleware candidate added to the PR.

Assisted-by: OpenAI Codex

sriaradhyula:
Debugging PR. Can be closed

PR #1319, opened 2026-04-28
feat(supervisor): add ToolCallLimitMiddleware and ModelCallLimitMiddleware with configurable env vars

sriaradhyula:
## Summary

- Adds `ToolCallLimitMiddleware` to cap total tool invocations per run — prevents agents from hammering external tools in a loop
- Adds `ModelCallLimitMiddleware` to cap total LLM inference calls per run — prevents infinite reasoning loops where the model calls itself repeatedly
- Adds opt-in `SummarizationMiddleware` to compress conversation history when it approaches token/message thresholds — prevents context overflow on long-running tasks; uses the same model as the supervisor (via `LLMFactory().get_llm()`)
- All three limits are individually toggleable and configurable via env vars / Helm configmap — no code change needed at deploy time
- Adds E2E test for `ToolCallLimitMiddleware` using `LLMToolEmulator` + `_ToolAwareFakeModel`

### New env vars (Helm values: `supervisor-agent`)

| Env var | Default | Description |
|---------|---------|-------------|
| `ENABLE_TOOL_CALL_LIMIT_MIDDLEWARE` | `"true"` | Toggle tool call cap |
| `TOOL_CALL_LIMIT` | `"500"` | Max tool calls per run |
| `TOOL_CALL_LIMIT_EXIT_BEHAVIOR` | `"continue"` | `continue` / `end` / `error` |
| `ENABLE_MODEL_CALL_LIMIT_MIDDLEWARE` | `"true"` | Toggle LLM call cap |
| `MODEL_CALL_LIMIT` | `"200"` | Max LLM inference calls per run |
| `MODEL_CALL_LIMIT_EXIT_BEHAVIOR` | `"end"` | `end` / `error` / `continue` |
| `ENABLE_SUMMARIZATION_MIDDLEWARE` | `"false"` | Toggle summarization (opt-in) |
| `SUMMARIZATION_TRIGGER_TOKENS` | `"100000"` | Token count that triggers summarization (~50% of 200K window) |
| `SUMMARIZATION_TRIGGER_MESSAGES` | `"50"` | Message count that triggers summarization |
| `SUMMARIZATION_KEEP_MESSAGES` | `"20"` | Recent messages to keep after summarization |

**Why `end` for model call limit vs `continue` for tool call limit?** Hitting the LLM call limit is a stronger signal of an infinite reasoning loop; ending gracefully is safer. Tool call limit overflow can sometimes recover if the model summarizes what it knows.

**Why `false` for summarization by default?** Summarization requires an LLM call per trigger and silently drops older message content. Enable explicitly when sessions are known to run long.

**Why `100000` token trigger?** Claude has a 200K context window; triggering at 50% avoids unnecessary summarization on healthy sessions while still catching runs that are genuinely at risk of overflow.

## Test plan

- [X] `test_tool_call_limit_e2e.py` — `continue` behavior (limit-exceeded ToolMessage injected) and `end` behavior (agent stops at limit)
- [X] Run: `uv run --group unittest pytest ai_platform_engineering/multi_agents/platform_engineer/tests/test_tool_call_limit_e2e.py -v`
- [ ] Enable `ENABLE_SUMMARIZATION_MIDDLEWARE=true` and run a long session exceeding 100K tokens — verify older messages are summarized and the session continues


artdroz (dismissed):
LGTM

PR #1339, opened 2026-05-05
fix(mcp): remove trailing slash from default HTTP MCP path

dasein:
## Summary

The streamable-http MCP client does not follow redirects. FastMCP (via
Starlette) mounts at `/mcp` and issues a **307 Temporary Redirect** for
requests to `/mcp/`. With the old default path of `"/mcp/"`, every
HTTP-transport agent failed to load tools on startup.

The Confluence agent was the first visible casualty because it uses
`ghcr.io/sooperset/mcp-atlassian`, a third-party FastMCP server that
strictly rejects the trailing slash. CNOE-built MCP servers (Jira,
VictorOps) happen to tolerate it. Slack and Webex were unaffected
because they already pass `path="/mcp"` explicitly.

The same latent bug affects the GitHub and GitLab agents, which call
`resolve_mcp_url`/`build_http_mcp_config` without an explicit path and
would hit the same 307 if deployed against any strict FastMCP server.

## Changes

- `resolve_mcp_url`: default `path` `"/mcp/"` → `"/mcp"`
- `build_http_mcp_config`: default `path` `"/mcp/"` → `"/mcp"`

## Test plan

- [ ] Confluence agent connects successfully (no 307 on tool load)
- [ ] Jira / VictorOps agents unaffected (they tolerate both paths)
- [ ] Slack / Webex unaffected (explicit `path="/mcp"` unchanged)

### DIFF
--- a/ai_platform_engineering/utils/mcp_config.py
+++ b/ai_platform_engineering/utils/mcp_config.py
@@ -65,12 +65,12 @@ def resolve_mcp_url(
     *,
     default_host: str = _DEFAULT_HOST,
     default_port: str = _DEFAULT_PORT,
-    path: str = "/mcp/",
+    path: str = "/mcp",
 ) -> str:
     """Build the full MCP HTTP URL for *agent_name*.
 
     Combines :func:`resolve_mcp_host` and :func:`resolve_mcp_port` with
-    the given *path* (defaults to ``/mcp/``).
+    the given *path* (defaults to ``/mcp``).
     """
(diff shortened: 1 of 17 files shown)


sriaradhyula (approved):

sriaradhyula (approved):
Fix is correct and well-explained. Removing the trailing slash from the default MCP path eliminates the 307 redirect that the streamable-http client doesn't follow. Clean, minimal change.

PR #1354, opened 2026-05-06
feat(skills): add migration-guide skill for helm chart upgrades

sriaradhyula:
## Summary

- Adds `skills/migration-guide/SKILL.md` — a skill that generates a structured upgrade runbook between two chart versions
- Fetches default values for both versions via `helm show values`, diffs them, and classifies changes as Breaking / New required / Deprecated / Default changed
- Produces before/after YAML snippets, a copy-paste upgrade runbook, and cross-references the user's existing `values.yaml` to flag only relevant changes
- Works in coding-agent (writes to `docs/migration/<from>-to-<to>.md`) and chat-only contexts

## Motivation

Upgrading `values.yaml` between chart versions is error-prone — keys get renamed, tags replaced by global flags, defaults change silently. This skill automates discovery and documentation of those changes. Example: 0.4.5 → 0.4.8 introduced `global.skillScanner.enabled` replacing the unsupported `tags.skill-scanner`.

## Test plan

- [ ] Run the skill with `from=0.4.5`, `to=0.4.8` and verify the generated guide covers the skillScanner breaking change
- [ ] Verify chat-only mode renders fenced code blocks instead of writing to disk

### DIFF
--- a/skills/migration-guide/SKILL.md
+++ b/skills/migration-guide/SKILL.md
@@ -1,0 +1,279 @@
+---
+name: migration-guide
+description: >
+  Generate a migration guide between two ai-platform-engineering chart releases.
+  Diffs the Helm values schema, highlights breaking changes, renamed keys, new
+  required fields, and deprecated options. Produces a step-by-step upgrade
+  runbook with before/after YAML snippets. Use when a user is upgrading their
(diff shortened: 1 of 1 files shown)


sriaradhyula:
Superseded by #1356 which rewrites this as a generic release-docs skill covering both release notes and migration guides.

PR #1363, opened 2026-05-07
feat(a2a): outbound OAuth2 client_credentials auth via AuthInterceptor

tmttodd:
## Summary

Adds end-to-end m2m authentication for the supervisor → remote-agent A2A leg, fully backward-compatible.

Today, [`OAuth2Middleware`](../blob/main/ai_platform_engineering/utils/auth/oauth2_middleware.py) validates inbound JWTs but the supervisor's [`A2ARemoteAgentConnectTool`](../blob/main/ai_platform_engineering/utils/a2a_common/a2a_remote_agent_connect.py#L144) constructs an `A2AClient` with **no auth wiring**. Setting `A2A_AUTH_OAUTH2=true` on a remote agent therefore 401-rejects every supervisor → agent call. The placeholder [`user_jwt = "TBD_USER_JWT"`](../blob/main/ai_platform_engineering/utils/a2a_common/base_langgraph_agent.py#L370) confirms this was unfinished.

This change wires up `a2a-sdk`'s built-in `AuthInterceptor` only when the remote agent advertises it.

### 1. `a2a_server.py` — env-conditional security scheme

When `A2A_AUTH_OAUTH2=true` AND `TOKEN_ENDPOINT` is set, the AgentCard publishes an `oauth2` security scheme with the client_credentials flow. When env isn't set, `security=[{"public": []}]` is preserved exactly.

Misconfigured agents (auth on, no token endpoint) fall back to public with a warning instead of crash-looping.

### 2. `oauth2_client_credentials_service.py` (new)

`CredentialService` implementation that mints + caches client_credentials access tokens. Refreshes at 80% of TTL, keeps stale-but-not-expired tokens through transient IdP outages, serializes refreshes via `asyncio.Lock`. Uses the existing `OAUTH2_CLIENT_ID` / `OAUTH2_CLIENT_SECRET` / `TOKEN_ENDPOINT` env vars already documented for the inbound middleware — **no new config**.

### 3. `a2a_remote_agent_connect.py` — wire the interceptor

`connect()` inspects the fetched `AgentCard` for an `oauth2` security requirement and, if present, attaches `AuthInterceptor(OAuth2ClientCredentialsService())` to the `A2AClient`. Cards advertising `{\"public\": []}` continue with no interceptor — **zero behavior change for agents that didn't opt in**.

## Backward compatibility

- Default env (`A2A_AUTH_OAUTH2` unset or false): identical to upstream behavior.
- Existing `test_agent_card_security_public` still passes.
- No new dependencies (`httpx`, `jwt`, `a2a-sdk` are all already pinned).
- No new env vars.

## Tested

New `TestA2AServerOAuth2SecurityScheme` cases cover the env matrix:
- off (default)
- on without TOKEN_ENDPOINT (falls back to public + warning)
- on with full config (advertises oauth2 scheme)
- token URL is surfaced correctly on the scheme

End-to-end: verified on a Keycloak realm. AgentCard advertises `oauth2`; supervisor mints a client_credentials token via `OAuth2ClientCredentialsService`; remote agent validates the JWT via the existing `OAuth2Middleware`; supervisor logs `Outbound A2A auth enabled for {agent}` and successful tool calls flow through.

## Naming

`OAuth2ClientCredentialsService` (not Keycloak-specific) so any RFC6749 §4.4 IdP works — Keycloak, Okta, Auth0, AWS Cognito, Azure AD, etc.

## Related

A separate PR will follow up to harden the [inbound `cid` claim check](../blob/main/ai_platform_engineering/utils/auth/oauth2_middleware.py#L121-L124) (which currently silently accepts tokens missing the client-identity claim). That hardening is gated behind an opt-in flag for backward compatibility.

### DIFF
--- a/ai_platform_engineering/utils/a2a_common/oauth2_client_credentials_service.py
+++ b/ai_platform_engineering/utils/a2a_common/oauth2_client_credentials_service.py
@@ -24,6 +24,7 @@ import asyncio
 import logging
 import os
 import time
+from urllib.parse import urlparse
 
 import httpx
 import jwt
@@ -32,6 +33,61 @@ from a2a.client.middleware import ClientCallContext
 
 logger = logging.getLogger(__name__)
 
+# Default token TTL (seconds) when the issued token is opaque (not a JWT)
+# or its `exp` claim cannot be decoded. Conservative: short enough that a
+# stale token won't outlive a typical IdP rotation, long enough to avoid
+# minting on every request.
+_DEFAULT_OPAQUE_TOKEN_TTL = 60.0
+
+# Hostnames considered acceptable for plaintext (http://) token endpoints.
+# Loopback only — anything else triggers a warning because client_secret
+# would travel in plaintext over the wire.
+_PLAINTEXT_OK_HOSTS = frozenset({'localhost', '127.0.0.1', '::1'})
+
+_TOKEN_ENDPOINT_VALIDATED = False
+
+
+def _validate_token_endpoint_scheme(token_endpoint: str) -> None:
+    """Warn (once per process) if TOKEN_ENDPOINT uses plaintext over a non-loopback host.
+
+    Refuses to escalate to an exception so misconfigured operators see
+    auth failures (clear) rather than crash loops (confusing). The warning
+    surfaces in startup logs.
+    """
+    global _TOKEN_ENDPOINT_VALIDATED
+    if _TOKEN_ENDPOINT_VALIDATED:
+        return
+    _TOKEN_ENDPOINT_VALIDATED = True
+    parsed = urlparse(token_endpoint)
+    if parsed.scheme == 'https':
+        return
+    if parsed.scheme == 'http' and parsed.hostname in _PLAINTEXT_OK_HOSTS:
+        logger.info(
+            "TOKEN_ENDPOINT uses plaintext http:// to a loopback host (%s) — "
+            "acceptable for local development only.",
+            parsed.hostname,
+        )
+        return
+    logger.warning(
+        "TOKEN_ENDPOINT scheme is %r (not https). Client secret will travel "
+        "in plaintext over the wire on each token mint. Use https:// in any "
+        "non-loopback environment.",
+        parsed.scheme or '(missing)',
+    )
+
+
+def _scrub_secret(text: str, secret: str) -> str:
+    """Redact `secret` from `text`. Cheap belt-and-suspenders for log lines.
+
+    `httpx` doesn't include request bodies in default exception reprs, but
+    some IdPs echo `client_secret` back in their error responses (e.g.
+    "invalid_client_secret: <secret>"). Scrubbing here ensures the secret
+    never leaks via the credential service's own error path.
+    """
+    if not secret or not text:
+        return text
+    return text.replace(secret, '***REDACTED***')
+
 
 class OAuth2ClientCredentialsService(CredentialService):
     """Mints + caches OAuth2 client_credentials access tokens.
@@ -39,6 +95,10 @@ class OAuth2ClientCredentialsService(CredentialService):
     Hard-coded to scheme name 'oauth2' to match what `A2AServer`'s patched
     `_build_security_for_card()` advertises. Agents using a different
     scheme name should subclass and override `SCHEME_NAME`.
+
+    One instance per `AuthInterceptor` — token cache is per-instance, not
+    per-process. Multiple instances minting against the same IdP/client is
+    inefficient but not unsafe.
     """
 
     SCHEME_NAME = 'oauth2'
@@ -57,12 +117,15 @@ class OAuth2ClientCredentialsService(CredentialService):
         """Return a valid access token for the named scheme, or None."""
         if security_scheme_name != self.SCHEME_NAME:
             return None
-        now = time.time()
         async with self._lock:
+            now = time.time()  # capture inside lock to avoid TOCTOU
             if not self._token or now >= self._refresh_after:
                 try:
                     await self._refresh()
                 except Exception as exc:  # noqa: BLE001
+                    # Stale-on-error: serve the existing token until its
+                    # actual exp if a refresh fails (transient IdP outage).
+                    now = time.time()
                     if self._token and now < self._token_exp:
                         logger.warning(
                             "OAuth2 token refresh failed; using stale token "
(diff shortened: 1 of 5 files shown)


tmttodd:
## Self-review (security audit)

After opening, I went back and did a formal Auditor pass on this PR. Findings + fixes are in [`00d850b`](https://github.com/cnoe-io/ai-platform-engineering/pull/1363/commits/00d850b). Documenting the full audit log here for maintainer transparency.

### HIGH

| # | Finding | Fix |
|---|---------|-----|
| 1 | TOKEN_ENDPOINT scheme not validated. Plaintext `http://` to a non-loopback host leaks `client_secret` on each token mint. | Warn (not reject) on `http://` to non-loopback. Loopback (`localhost`/`127.0.0.1`/`::1`) logs at INFO for local-dev. Applied in both `oauth2_client_credentials_service._validate_token_endpoint_scheme` (mint time) and `a2a_server._build_security_for_card` (when advertising the scheme). |
| 2 | `httpx.HTTPStatusError` can include the request body in its string repr. Some IdPs echo `client_secret` back in error responses. | Catch `HTTPStatusError`, scrub the secret via new `_scrub_secret` helper, re-raise with redacted message. Belt-and-suspenders — httpx default text doesn't include the body, but operator-side wrappers might. |

### MEDIUM

| # | Finding | Fix |
|---|---------|-----|
| 3 | TOCTOU on `now = time.time()` outside the `asyncio.Lock`. | Capture `now` inside the lock so cache decisions and refresh timestamps see the same monotonic snapshot. |
| 4 | Token from IdP isn't always a JWT. Previous code did `jwt.decode(...)` and would crash on opaque tokens. | Catch `jwt.PyJWTError`, fall back to a conservative default TTL (60s, configurable via `_DEFAULT_OPAQUE_TOKEN_TTL`). |
| 5 | `httpx.AsyncClient(timeout=10)` didn't honor proxy / CA-bundle env (HTTPS_PROXY, REQUESTS_CA_BUNDLE). Common in corporate networks. | Add `trust_env=True`. |
| 6 | When the remote agent advertises oauth2 but local interceptor wiring fails (bad import, missing dep), the previous code logged a warning and proceeded with `interceptors=[]` — silently sending unauthenticated requests that the remote will (correctly) 401-reject, producing a confusing error chain. | Drop the try/except. Wiring failure now raises at startup with a clear stack trace. |

### LOW

| # | Finding | Fix |
|---|---------|-----|
| 7 | `verify_signature=False` on the JWT decode wasn't commented. | Added comment explaining the rationale (we just minted the token from our own IdP; signature verification belongs at the receiving agent, not the minting client). |
| 8 | `scopes={}` on the advertised security scheme wasn't commented. | Added comment: this implementation does not use scope-based authz; authorization happens via `OAUTH2_CLIENT_ID` allowlist in the inbound middleware. |

### New tests

Added `tests/test_oauth2_client_credentials_service.py` with 9 tests covering:
- SCHEME_NAME filter (returns None for non-matching scheme)
- First call mints + caches; subsequent within-TTL call reuses cache
- Stale-on-error: refresh failure with valid token serves stale until actual exp
- Hard-fail when refresh fails AND no usable token
- Opaque-token (non-JWT) fallback to default TTL
- Secret scrubbing in `_scrub_secret`
- TOKEN_ENDPOINT scheme validation: https → silent, http loopback → INFO, http non-loopback → WARNING

### What I did NOT change

- Per-instance `asyncio.Lock` (one cache per service instance). Multiple instances minting against the same IdP/client is inefficient but not unsafe; out of scope for this PR.
- Soft fall-back to public when `A2A_AUTH_OAUTH2=true` but `TOKEN_ENDPOINT` is missing. Considered hard-failing at startup; chose to keep the soft fall-back so misconfigured agents surface the warning rather than crash-loop. The inbound middleware still enforces.

Open to maintainer feedback on any of the above; happy to amend or split further.

tmttodd:
## CI re-trigger after fixes

Force-pushed amended commits ([01ef4ee](https://github.com/cnoe-io/ai-platform-engineering/pull/1363/commits/01ef4ee)). Two real failures fixed:

| Check | Cause | Fix |
|-------|-------|-----|
| `linter` | `F401` unused `asyncio` import in new `test_oauth2_client_credentials_service.py` | Removed |
| `conventional-commits` | Used `fixup(...)` on the audit-fix commit; not a valid CC type | Renamed to `fix(...)` |

The other two failing checks are **fork-PR runner limitations**, not code issues — flagging for maintainer awareness:

- **`check-proprietary-content`** — actually passed (`✅ No proprietary content detected`) but the comment-post step `403`d on `Resource not accessible by integration`. Fork PRs run with `pull_request` trigger, which has read-only `issues` permission. Workflow would need either `pull_request_target` or skipping the comment step on fork PRs to avoid the false-failure.

- **`Bump version files on PR`** — failed with `Input required and not supplied: token`. Same fork-PR root cause: `secrets.GITHUB_TOKEN` isn't available to write back to the fork branch.

Both are routine for fork contributions. Re-running with `pull_request_target` permissions or having a maintainer push the same branch from a same-repo branch would clear them.

sriaradhyula:
@subbaksh @cisco-erilutz @kevkantes - Can you guys approve this?

kevkantes (approved):

kevkantes:
> @subbaksh @cisco-erilutz @kevkantes - Can you guys approve this?

Rubber stamped, let me know if you wanted an actual review

ai_platform_engineering/utils/a2a_common/oauth2_client_credentials_service.py
```diff
@@ -0,0 +1,204 @@
...
+                            exc,
+                        )
+                        raise
+            return self._token
+
+    async def _refresh(self) -> None:
+        client_id = os.environ['OAUTH2_CLIENT_ID']
+        client_secret = os.environ['OAUTH2_CLIENT_SECRET']
+        token_endpoint = os.environ['TOKEN_ENDPOINT']
+        _validate_token_endpoint_scheme(token_endpoint)
```
sriaradhyula, line 148: Please make this fail closed before any credential mint happens. `_validate_token_endpoint_scheme()` currently only warns for non-loopback `http://`, and then `_refresh()` still posts `client_secret` to that endpoint. For a confidential-client flow, warning-only means a production misconfig leaks the client secret on every token refresh. I’d reject non-HTTPS endpoints by default, keep loopback HTTP for local dev, and require an explicit dev override if there is a real need for anything else.

sriaradhyula (changes requested):
Thanks for wiring the outbound A2A OAuth2 path; the AuthInterceptor integration itself lines up with the a2a-sdk model. I’m requesting changes on two security points before this lands:

1. The new client-credentials minting path must fail closed for non-loopback plaintext `TOKEN_ENDPOINT` values. Right now the code warns but still sends `client_secret` to the configured URL.

2. This PR’s advertised security model depends on the inbound `OAUTH2_CLIENT_ID` allowlist, but the current middleware accepts tokens that omit `cid`. Since the new AgentCard says the remote agent is OAuth2-protected and the PR comments say authorization is via the client-id allowlist, we should harden that dependency in this PR or before relying on the new outbound auth flow. Please require one configured client identity claim (`azp`, `client_id`, or `cid`) and reject tokens that have none.

One additional design note: because the advertised scheme has `scopes={}` and the credential service mints one env-backed token, deployments need distinct audience/client boundaries per remote agent, or a compromised peer can replay the supervisor’s token to other agents until expiry. That can be handled through config/docs if it is intentional, but it should be explicit.

sriaradhyula:
To make the maintainer expectation explicit: the outbound OAuth2 client-credentials path should either be fixed to fail closed, or removed from this PR.

Specifically, please do one of these before merge:

1. Reject non-loopback plaintext `TOKEN_ENDPOINT` values before any token minting request sends `client_secret`, while still allowing loopback HTTP for local development if needed.
2. Remove/defer the outbound credential-minting and OAuth2 AgentCard advertisement pieces from this PR until the transport safety behavior is fixed.

Warning-only is not sufficient for this path because the configured endpoint receives the confidential client secret on every refresh.

PR #1398, opened 2026-05-12
fix(skills): remove install pagination and retitle to Advanced install

sriaradhyula:
## Summary

- Removes `page=1&page_size=20` from the catalog URL built by `buildCatalogUrl` — the API returns all results when `page` is omitted, so bulk install no longer caps at 20 skills
- Removes the now-unused **Page size** input from the advanced catalog URL builder UI
- Renames the try-gateway page heading from _"How to use Skills API Gateway with coding agents"_ to **"Advanced install"** and rewrites the description to direct first-time users to Quick Install, reserving Steps 1–3 for custom/scripted flows

## Test plan

- [x] `TrySkillsGateway` test suite (26 tests) passes
- [ ] Verify quick-install URL no longer contains `page_size=20` in the dialog
- [ ] Verify the page heading reads "Advanced install" on `grid.outshift.io`

### DIFF
--- a/ui/src/components/skills/TrySkillsGateway.tsx
+++ b/ui/src/components/skills/TrySkillsGateway.tsx
@@ -157,7 +157,6 @@ export function TrySkillsGateway() {
   const [queryRepo, setQueryRepo] = useState("");
   const [queryTags, setQueryTags] = useState("");
   const [queryVisibility, setQueryVisibility] = useState("");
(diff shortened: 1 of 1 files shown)


sriaradhyula:
Changes landed directly on main via cherry-pick (a430f0780). Title rename was dropped per feedback.

PR #1411, opened 2026-05-13
feat(dynamic-agents): add curl builtin tool for PUT/POST/PATCH/DELETE support

cisco-erilutz:
## Summary

- The `fetch_url` builtin tool for dynamic agents only supports GET requests; agents that need to call write APIs (PUT, POST, PATCH, DELETE) had no way to do so
- The supervisor/multi-agent stack already has a `curl` tool (added in PR #1242) but it was never wired into the dynamic agents builtin tools system
- This PR adds `curl` as a first-class builtin tool for dynamic agents, matching the same domain ACL and https-only guard pattern already used by `fetch_url`

## Changes

- **`models.py`**: `CurlToolConfig` model with `enabled`/`allowed_domains` fields; `BuiltinToolsConfig.curl` field
- **`builtin_tools.py`**: `create_curl_tool()` factory with domain ACL (same pattern as `create_fetch_url_tool`), https-only enforcement, `strip_html` support, and exposed via `get_builtin_tool_definitions()` for UI discovery
- **`agent_runtime.py`**: wire `curl` into `_build_builtin_tools()` alongside `fetch_url`
- **`test_builtin_tools.py`**: 4 new tests — http-blocking, domain ACL, successful execution, tool definition discovery

## Configuration

Enable for an agent via `builtin_tools.curl.enabled: true` (disabled by default, same as `fetch_url`). Optionally scope to specific domains via `allowed_domains`.

## Test plan

- [ ] `pytest ai_platform_engineering/dynamic_agents/tests/test_builtin_tools.py` — all 5 tests pass
- [ ] Agent with `curl` enabled can execute PUT/POST requests
- [ ] Agent with `curl` disabled does not have the tool available
- [ ] Domain ACL blocks requests to non-allowed domains
- [ ] `http://` URLs are rejected with a clear error

### DIFF
--- a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/models.py
+++ b/ai_platform_engineering/dynamic_agents/src/dynamic_agents/models.py
@@ -203,6 +203,20 @@ class FetchUrlToolConfig(BaseModel):
     )
 
 
+class CurlToolConfig(BaseModel):
+    """Configuration for the curl built-in tool."""
+
+    enabled: bool = Field(False, description="Whether the tool is enabled")
+    allowed_domains: str = Field(
+        default="*",
+        description=(
+            "Comma-separated domain patterns. "
+            "Use * for all, *.domain.com for subdomains, or exact domain. "
+            "Empty string blocks all domains."
+        ),
+    )
+
+
 class CurrentDatetimeToolConfig(BaseModel):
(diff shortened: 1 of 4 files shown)


cisco-erilutz:
Superseded by a prebuild/ branch PR for Docker image testing.

PR #1419, opened 2026-05-13
fix(slack): use is_overthink_message context var to always respond to thread @mentions

kevkantes:
# Description

When an agent has `overthink: enabled: true` and `listen: all`, it would silently drop responses to direct @mentions in threads — because the skip-marker check ran regardless of whether the message was a top-level channel message or a thread @mention. Overthink filtering is only appropriate for ambient channel noise, not for explicit user requests in a thread.

**Changes:**

- `app.py`: Injects `is_overthink_message: not is_thread_reply` into `client_context` for every `handle_mention` call. Top-level messages get `true`; thread @mentions get `false`.
- `utils/ai.py`: Gates `_check_overthink_skip` on `is_overthink_message` being `true`, so skip markers are never evaluated for thread @mentions. Updates `OVERTHINK_BOILERPLATE` to a Jinja2 conditional: when `is_overthink_message` is `false`, the rendered prompt tells the LLM to always respond directly and never emit `[DEFER]`/`[LOW_CONFIDENCE]`; when `true`, the full STEP 1/2/3 filter is rendered as before.

Agents with custom `[DEFER]` rules in their system prompt (not using the boilerplate) can gate them with `{% if client_context.is_overthink_message %}` to get the same behaviour.

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


dasein (approved):
LGTM

PR #1444, opened 2026-05-16
chore(release): prepare 0.5.1 release branch

sriaradhyula:
## Summary

- Squashes the approved 0.5.0 release payload onto the latest `main` baseline as `release/0.5.1`.
- Preserves RBAC/OpenFGA, Keycloak, Slack, MCP auth, UI admin, docs, CI, and chart changes while resolving merge drift intentionally.
- Normalizes release metadata to `0.5.1-rc.1` and regenerates affected UI and Python lockfiles.

## Test plan

- [x] `git diff --check`
- [x] no unresolved merge conflicts or conflict markers
- [x] `npm test -- --runTestsByPath 'src/app/(app)/admin/__tests__/admin-page.test.tsx'`


charts/ai-platform-engineering/Chart.yaml
```diff
@@ -179,30 +179,47 @@ dependencies:
...
-    version: 0.4.12-dev.11 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.5.1-rc.4 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     condition: global.skillScanner.enabled
   # Slack Bot Integration (client, not an agent)
   - name: slack-bot
-    version: 0.4.12-dev.11 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.5.1-rc.4 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     tags:
       - slack-bot
+  # Keycloak Identity Provider for CAIPE RBAC/OIDC
```
suwhang-cisco: For agentgateway, I previous used existing upstream helm chart like this: https://github.com/cnoe-io/ai-platform-engineering/pull/1145/changes#diff-48101687e353f0bbc72b9851a06ba0dc9b995364dcb8c89853e48d238e7c1c58R204-R224. Could we also do this here or is there a reason we need customisation?

Similarly for keycloak, if there is an official helm chart it'd be good to use that as well (although not sure this exists)

sriaradhyula: Checked this against the current chart shape. For `agentgateway`, the earlier PR used the upstream controller/CRD chart path; this branch is intentionally using a local subchart because it deploys the standalone static-config proxy with CAIPE OpenFGA ext_authz wiring. Moving to the upstream controller chart would be a larger migration because we need to preserve the static MCP routing and authz bridge contract for this release.

For `keycloak`, I do not see an official chart that covers our CAIPE-specific bootstrap invariants. The local chart owns realm import, token-exchange/OBO setup, IdP/auth-flow reconciliation, branded login theme packaging, and ExternalSecret wiring. We can revisit an upstream/base chart after 0.5.1, but swapping it in here would risk changing the release auth path.

I added comments in `charts/ai-platform-engineering/Chart.yaml` documenting why both local subcharts are intentional for this release.

subbaksh:
## Dynamic agents auth feedback

exclusively inside `ai_platform_engineering/dynamic_agents/`

**Feel free to feed them into your coding agent**

### 1. CRUD Routes Are Unused

**File:** `dynamic_agents/routes/agents.py`

The DA service exposes full CRUD endpoints (`POST /agents`, `PATCH /agents/{id}`, `DELETE /agents/{id}`). These duplicate what the Next.js gateway already does at `ui/src/app/api/dynamic-agents/route.ts`.

The Next.js gateway:
- Talks directly to MongoDB for all agent create/update/delete operations
- Calls `requireResourcePermission(session, { type: "agent", id, action: "write" })` via OpenFGA
- Manages OpenFGA relationship tuples when agent tool bindings change (`reconcileAgentRelationships`)
- Validates owner-team membership before allowing creates

The DA CRUD routes:
- Also talk to MongoDB
- Use `require_admin` which checks an `is_admin` flag from the `X-User-Context` header
- Have no OpenFGA integration for write operations
- Have no team-ownership validation

Since the gateway goes direct to MongoDB, no client should be hitting the DA CRUD routes. They appear to be dead code that creates ambiguity about which service is the source of truth for agent configuration.

**Recommendation:** Confirm that nothing calls these endpoints (check access logs or remove and observe). Then remove them to maintain the clean contract that DA is a runtime reader.

---

### 2. Two Auth Models Coexisting

This PR introduces `JwtAuthMiddleware` which validates Bearer tokens via JWKS. This directly contradicts the existing `auth/auth.py` module whose docstring states DA never validates JWTs.

The result is two authentication paths running simultaneously in the same service:

| Path | Mechanism | Used by |
|------|-----------|---------|
| `X-User-Context` header | Base64-encoded JSON injected by the gateway, trusted without verification | CRUD routes (admin operations) |
| Bearer JWT + JWKS validation | Full cryptographic signature verification against Keycloak | Chat/invoke routes |

The middleware includes a "lenience" mode that lets requests through without a Bearer token, so the legacy `X-User-Context` path continues to function. Both paths are live indefinitely.

**Why this is a problem:**

First, the privilege model is inverted. The highest-privilege operations (create, update, delete agents) rely on the weaker trust-the-header mechanism. The lowest-privilege operation (chatting with an agent) gets full cryptographic validation plus external PDP calls. If the goal is defense-in-depth, the admin operations should have the stronger auth, not the weaker one.

Second, the two paths can disagree. The gateway might inject `X-User-Context` with `is_admin: true`, but DA's OpenFGA check on the chat path might deny the same user (because tuples are out of sync, or the user's email vs subject ID resolve differently). There is no single source of truth for "is this user authorized?"

Third, two auth models means two sets of failure modes, two code paths to audit for security issues, and cognitive overhead for developers who need to understand which path applies to which route.

**Recommendation:** Converge on one model. The gateway-trusted model is simpler and already works for the entire service. If Bearer validation is the desired end-state, migrate all routes to it and remove `X-User-Context` support entirely. The current state of having both is the worst of both worlds — the complexity of cryptographic validation with the trust assumptions of gateway headers still baked in.

---

### 3. OpenFGA Check Duplicates the Gateway

**File:** `dynamic_agents/auth/openfga_authz.py` (~300 lines)

This file re-implements the same OpenFGA permission check that the Next.js gateway already performed before forwarding the request to DA. Specifically:

- `_get_openfga_store_id()` — resolves the store by name via `GET /stores` (identical to the RAG service's version)
- `_check_agent_use()` — `POST /stores/{id}/check` with tuple `user:X can_use agent:Y`
- JWT payload decoding without verification (base64 decode of the middle segment) to extract subject and email

The gateway already evaluates this exact permission. In the `GET /api/dynamic-agents` handler, it calls `filterResourcesByPermission(session, items, { type: "agent", action: "use" })`. Only agents the user can access are returned to the UI. When the user then initiates a chat, the gateway has already confirmed they have `can_use` on that agent.

The DA re-check adds:
- An additional network round-trip to OpenFGA on every chat request (~5-50ms depending on network)
- A new failure mode: if OpenFGA is momentarily unreachable, DA returns 503 to the user even though the gateway already approved the request 200ms earlier
- ~300 lines of authorization code to maintain in Python, duplicating logic that already exists in TypeScript in the gateway

**JWT signature validation is worthwhile defense-in-depth.** Verifying that the token was issued by Keycloak is a cheap local operation (~1ms with cached JWKS keys) and catches forged headers or gateway bypass. That is a different thing from re-asking the PDP whether the user has permission — which is what this file does.

**Recommendation:** Remove `openfga_authz.py`. Keep JWT signature validation as defense-in-depth (it's cheap and local). If additional defense-in-depth beyond JWT validation is needed, consider having the gateway include a signed claim (e.g., a short-lived, HMAC-signed `X-Authz-Decision` header) that DA can verify locally without a network call.

---

### 4. Audit + OTEL Coupling in AuthZ

**File:** `dynamic_agents/auth/openfga_authz.py`

The `require_agent_use_permission()` function handles four distinct concerns in a single call:

1. **Input validation** — checks agent ID format against a regex
2. **Authorization** — calls OpenFGA to check the `can_use` relation
3. **Audit logging** — writes a structured event to MongoDB
4. **Distributed tracing** — exports an OTEL span to the traces collector

The audit piece brings significant infrastructure with it:
- A module-level `MongoClient` initialized lazily with a thread lock
- Five index creation calls (`ts`, `type+ts`, `subject_hash+ts`, `source+ts`, `correlation_id`)
- Privacy-aware subject hashing (SHA-256 with configurable salt)
- A dedicated `_persist_openfga_rebac_audit()` function

The OTEL piece manually constructs the OTLP JSON payload and POSTs it via httpx, rather than using the `opentelemetry-sdk` which would provide this through auto-instrumentation.

**Why this coupling is problematic:**

- You cannot test the authorization logic without mocking MongoDB and an OTEL collector endpoint.
- MongoDB write latency (even "best-effort") is in the request's hot path.
- The gateway is the actual decision-maker — having DA emit a second audit record per action means two records that may show different outcomes (gateway allowed, DA denied, or vice versa).
- Changing audit format/destination later requires modifying authorization code.

**Recommendation:** Separate concerns cleanly. The authz function should return a boolean decision. A decorator or middleware should handle audit logging based on that decision. Tracing should come from OpenTelemetry SDK instrumentation, not hand-rolled span construction.

---

### 5. OBO Token Exchange

**File:** `dynamic_agents/auth/obo_exchange.py` (~130 lines)

This is a well-motivated addition. DA calls downstream MCP servers (via agentgateway) that need to know the *user's* identity — not that the DA service account is calling, but that "user alice@company.com is requesting Jira tickets." RFC 8693 token exchange is the correct pattern for this, and DA (as the caller) is the right place to perform it.

The implementation:
- POSTs to Keycloak's token endpoint with `grant_type=urn:ietf:params:oauth:grant-type:token-exchange`
- Caches results in-memory keyed by `(user_sub, target_audience)` with expiry
- Returns `None` if OBO is not configured, falling back to forwarding the original token (graceful for local dev)

**Opportunities to simplify:**

The `JwtAuthMiddleware` already decoded and validated the JWT, storing it in a ContextVar. Yet `obo_exchange.py` decodes the JWT *again* (unverified base64) to extract `sub` for cache keying. It should read from the ContextVar instead.

The manual cache implementation with expiry arithmetic duplicates what `cachetools.TTLCache` already provides (and `utils/auth/keycloak_authz.py` already imports this library).

With these changes, the file goes from ~130 lines to approximately 40 — same functionality, less code to maintain.

**Recommendation:** Keep this module. Simplify by reading claims from the ContextVar and using the shared TTL cache pattern.

subbaksh:
## RAG auth feedback

exclusively inside `ai_platform_engineering/knowledge_bases`

**Feel free to feed them into your coding agent**

### 1. RAG Implements a Full Policy Engine That Doesn't Belong Here
`server/rbac.py` (~500 lines) contains:
- Role hierarchy (ANONYMOUS → READONLY → INGESTONLY → ADMIN) with numeric ranking
- Keycloak realm-role mapping (`admin`, `admin_user`, `kb_admin`, `team_member`, `chat_user`, `denied`)
- Trusted network detection (CIDR range matching, `X-Forwarded-For` parsing, `X-Trust-Token`)
- Client-credentials token detection heuristics
- Per-KB realm-role parsing via regex (`kb_reader:X`, `kb_ingestor:X`, `kb_admin:X`)
- OpenFGA knowledge-base check and list-objects calls
- MongoDB `team_kb_ownership` queries (no connection pooling, no caching)
- Query filter injection (mutates request to restrict datasource_ids)
- Datasource binding validation
This is an authorization service embedded inside a vector search application. RAG's job is to store and retrieve vectors — not to understand Keycloak role semantics, query MongoDB for team ownership, or call OpenFGA.

### 2. OpenFGA Is Called Twice Per UI Request
The gateway calls OpenFGA to check `user:X can_read knowledge_base:Y` before forwarding. RAG then calls OpenFGA again with the same tuple. Two network hops to the same PDP for the same answer. They can also disagree due to timing or configuration drift.

### 3. MongoDB Team-Ownership in the Hot Path
Every authenticated query triggers a MongoDB lookup to `team_kb_ownership` (no caching, new motor client per call). If MongoDB is slow, vector search is slow — even though the vector DB is healthy. This couples RAG's availability to a database it shouldn't need for query serving.

### 4. Role-Mapping Logic Is Duplicated
RAG re-implements Keycloak-to-role mapping in Python. The gateway does the same in TypeScript. If a new role is introduced, both must be updated. They can drift.

### 5. Doc-Level ACL Conflates Data Classification with Access Control
`doc_acl.py` tags individual documents in Milvus with ACL tags (e.g., `["team:security"]`) and filters at query time. While this sounds fine in theory, it has a fundamental flaw: it bakes access control decisions into the data layer.

**The problem:** When teams change (user joins/leaves, team renamed, reorg), you need to re-tag potentially millions of documents in Milvus. That's a bulk reindex operation for what should be an instant access-control change.

**The simpler model:** Use datasources as the access boundary. Control access at the datasource level via OpenFGA:
```
knowledge_base:platform-docs-public   → everyone can read
knowledge_base:security-incidents     → only team:security can read
```
- Alice joins security team → add one OpenFGA tuple → immediate access
- Alice leaves → remove tuple → instant revocation
- No documents touched. No reindex. No migration scripts.

If finer granularity is needed, create more datasources — that's a one-time ingest-time decision, not a runtime authorization concern. Datasource filtering is something RAG already does and is cheap (it's a metadata field on every vector).
`doc_acl.py` should be removed in favor of datasource-level RBAC via OpenFGA.

---
## The Simplification: Datasources Are Just Resources
The fundamental problem is that RAG treats knowledge-bases as a special case requiring bespoke logic. They are not special. They are resources — the same way agents are resources.
### Current Model (500 lines of Python)
```
Request → validate JWT → parse Keycloak realm roles → map to role hierarchy
  → detect client-credentials → check trusted network CIDRs → parse kb_reader:X roles
  → query MongoDB team_kb_ownership → call OpenFGA list-objects → intersect results
  → inject datasource_id filter → run vector search
```
### Proposed Model (same pattern as agents)
```
# Actors
user:alice
service:github-ingestor
# Resources (same type system as agents)
knowledge_base:platform-docs
# Relations (OpenFGA handles inheritance natively)
team:platform-eng    member     user:alice
team:platform-eng    reader     knowledge_base:platform-docs
service:gh-ingestor  can_ingest knowledge_base:platform-docs
organization:caipe   admin      user:bob              # → all KBs
```
OpenFGA resolves inheritance automatically: alice is in team → team has reader on KB → alice can read KB. No MongoDB query. No role parsing. No CIDR matching.
### What RAG's Auth Becomes
```python
# For UI requests (gateway already checked and constrained datasource_ids):
#   Validate JWT. Trust the gateway's constraints. Run the search.
# For direct-client requests (ingestors, agents):
#   Validate JWT. One OpenFGA check. Done.
async def require_kb_access(subject: str, kb_id: str, action: str) -> bool:
    return await openfga.check(
        user=f"user:{subject}", relation=f"can_{action}", object=f"knowledge_base:{kb_id}"
    )
```
The entire 500-line `rbac.py` collapses to:
1. JWT validation (defense-in-depth, keep it)
2. One OpenFGA check per KB access for direct clients
3. Trust gateway-constrained requests for the UI flow
4. Doc-level ACL filtering (keep `doc_acl.py`, it's domain logic)

### Why This Works

- **Ingestors** become `service:X` actors with `can_ingest` relations — modeled in OpenFGA, not detected via client-credentials heuristics
- **Team-scoped access** is team→KB relations in OpenFGA — no MongoDB collection needed
- **Role hierarchy** is implicit in the relation model (org admin implies all relations)
- **Trusted network** becomes unnecessary — ingestors authenticate with service-account tokens that have explicit relations
- **Datasource filtering replaces doc_acl** — access boundaries are datasources, controlled by OpenFGA tuples (instant to change). No per-document tagging, no reindexing when teams change.
- **The gateway** continues doing what it already does: resolve permissions, constrain `datasource_id` in the request body, forward to RAG
- **RAG stays a vector search service** — validates JWT, trusts the constrained request, runs the query

subbaksh:
## Slack bot auth feedback

exclusively inside `ai_platform_engineering/integrations/slack_bot/`

**Feel free to feed them into your coding agent**

### 1. Identity Linking Belongs in the BFF
`identity_linker.py` handles Slack → Keycloak user resolution:
- Queries Keycloak Admin API for `slack_user_id` attribute
- Fetches Slack profile email via Slack Web API
- Creates Keycloak users (JIT provisioning)
- Generates HMAC-signed linking URLs
The bot holds Keycloak Admin API credentials to do this. This is a significant privilege for a chat bot.

**Should be:** The bot sends `{slack_user_id, slack_email}` to a BFF endpoint. The BFF (which already has Keycloak Admin access for team/user management) resolves or creates the user and returns the Keycloak user ID. The bot never touches Keycloak directly.

### 2. Channel-to-Team Resolution Belongs in the BFF
`channel_team_resolver.py` (~200 lines) queries MongoDB:
```
channel_id → channel_team_mappings → team ObjectId → teams collection → slug
```
Then verifies team membership (MongoDB + Keycloak Admin API fallback).
The bot maintains its own MongoDB client, TTL caches, and membership resolution logic.

**Should be:** The bot sends `{channel_id, keycloak_user_id}` to a BFF endpoint. The BFF (which already owns team data, channel mappings, and OpenFGA) returns `{team_slug, allowed: true/false}`. The bot never queries MongoDB.

### 3. OBO Exchange Could Stay or Move
`obo_exchange.py` mints user-scoped tokens via Keycloak RFC 8693. It's well-designed with proper security invariants (verifies `active_team` claim matches request, fails hard on mismatch).

This one is debatable:
- **Keep in bot:** The bot needs the token immediately to call AgentGateway. A BFF round-trip adds latency to every message.
- **Move to BFF:** The BFF returns the OBO token as part of the identity resolution response. Cleaner separation but adds one more hop.
Either way is acceptable. If it stays in the bot, it should be the *only* Keycloak interaction the bot performs (not Admin API, not UMA).

### 4. Keycloak UMA Check Should Be Removed
`keycloak_authz.py` performs a Keycloak UMA ticket grant check before handlers execute. This is a second PDP (alongside OpenFGA via the BFF). The bot shouldn't be calling authorization endpoints directly.

**Should be:** Authorization is the BFF's concern. The bot's identity-resolution call to the BFF already implies "is this user allowed?" — if the BFF returns an OBO token, the user is authorized. If not, they aren't. No separate PDP call needed from the bot.

### 5. Channel ReBAC Delegation Is the Right Pattern
`slack_rebac.py` calls the BFF for channel access decisions:
```python
path = f"/api/admin/slack/channels/{workspace_id}/{channel_id}/access-check"
```
This is correct — the bot delegates to the BFF rather than implementing OpenFGA checks locally. **This is the pattern that all other auth concerns should follow.**

### 6. The Bot Holds Too Many Credentials
Currently the bot needs:
- Slack Bot Token (necessary)
- Slack Signing Secret (necessary)
- Keycloak Admin API credentials (should not need)
- Keycloak Bot Client ID + Secret for OBO (acceptable if OBO stays in bot)
- MongoDB URI (should not need)
- BFF/UI URL (necessary for delegation)

A thin bot should only need: Slack tokens + Keycloak OBO client (if OBO stays) + BFF URL.

sriaradhyula:
Dynamic Agents auth feedback is tracked in #1454.

Scope: remove or quarantine stale DA CRUD/auth surfaces, clarify the Bearer-token and `X-User-Context` transition, and separate the DA authorization decision path from audit/tracing concerns where feasible.

sriaradhyula:
Slack/Webex runtime ReBAC route feedback is tracked in #1455.

Scope: add bot/runtime-safe BFF access-check endpoints so Slack and Webex runtime checks can delegate to the BFF/OpenFGA logic without depending on admin UI route guards.

sriaradhyula:
Slack thin-bot architecture feedback is tracked in #1456.

Scope: move Slack identity linking and channel/team resolution behind BFF endpoints so the bot no longer needs Keycloak Admin or MongoDB access for the runtime message path.

sriaradhyula:
RAG authorization simplification feedback is tracked in #1457.

Scope: slim RAG toward a thin PEP around OpenFGA, remove duplicate role/team policy logic from query hot paths where safe, and keep direct-client authorization explicit.

sriaradhyula:
The cross-cutting RBAC transition documentation follow-up is tracked in #1458.

Scope: document which dual gates are intentional defense-in-depth for this release, which paths are transitional, and the retirement conditions for legacy auth surfaces.

PR #1449, opened 2026-05-18
fix(chat): add configurable conversation history persistence to /invoke endpoint

cisco-erilutz:
## Summary

- \`POST /api/v1/chat/invoke\` was using \`cache.ephemeral()\`, which creates a throw-away \`MemorySaver\` checkpointer per request, so the agent had no memory of prior turns even when the caller reused the same \`conversation_id\`
- Adds \`INVOKE_PERSIST_HISTORY\` env var (default: \`false\`) to control persistence: when \`false\` (default), each \`/invoke\` call remains ephemeral with no MongoDB writes; when \`true\`, the shared MongoDB-backed runtime cache is used, enabling multi-turn conversation history
- Uses \`AsyncExitStack\` so the ephemeral/persistent branch only controls how the runtime is acquired — the handler body is unified and runs the same code path either way
- Add \`apply_config_override\` call that was missing from \`chat_invoke\` (parity with streaming endpoints)
- Catch \`RuntimeCapacityError\` explicitly and return 503 instead of falling through to 500 (parity with streaming endpoints)
- Helm chart \`values.yaml\` updated with \`INVOKE_PERSIST_HISTORY: "false"\` and documentation of the trade-off

## Test plan

- [ ] Multi-turn \`/invoke\` conversation retains context across calls with the same \`conversation_id\` when \`INVOKE_PERSIST_HISTORY=true\`
- [ ] Default behaviour (\`INVOKE_PERSIST_HISTORY=false\`) is fully ephemeral — no MongoDB writes, no history across calls
- [ ] Single-turn \`/invoke\` calls continue to work as before under both settings
- [ ] HITL interrupt path (tool approval / user input form) still returns HTTP 400 with correct payload
- [ ] Streaming endpoint (\`/stream\`) unaffected

## Validation

Deployed two successive prebuild images to a dev environment with \`INVOKE_PERSIST_HISTORY=true\` and ran the following checks:

**Multi-turn conversation history**

Issued two \`/invoke\` calls with the same \`conversation_id\`:
- Turn 1: introduced a piece of information
- Turn 2: asked the agent to recall it

Before the fix, turn 2 had no memory of turn 1. After the fix, the agent correctly recalled the information from the prior turn — confirming that the MongoDB-backed runtime persists conversation state across calls when the env var is enabled.

**Smoke tests**

Ran the full suite of agent smoke tests against the patched deployment. All \`POST /api/v1/chat/invoke\` calls returned \`200 OK\` with no errors in the pod logs. Pod logs confirmed \`persist_history=True\` was being read from the env var correctly.

**No regressions observed** — streaming endpoint, HITL interrupt path, and all other agent invocations continued to function normally throughout testing.

### DIFF
--- a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/config.py
+++ b/ai_platform_engineering/dynamic_agents/src/dynamic_agents/config.py
@@ -61,6 +61,13 @@ class Settings(BaseSettings):
     # Maximum allowed TTL (0 = no cap, infinite allowed)
     max_fs_ttl_seconds: int = 0
 
+    # /invoke endpoint persistence
+    # When False (default), each /invoke call uses an ephemeral in-memory runtime that is
+    # discarded after the request — no MongoDB writes, no conversation history across calls.
+    # Set to True to use the shared MongoDB-backed runtime cache, enabling multi-turn
+    # conversation history via /invoke at the cost of additional MongoDB load.
+    invoke_persist_history: bool = False
+
     # Runtime
     agent_runtime_ttl_seconds: int = 60  # 60s inactivity TTL for agent runtimes
     # Max concurrent cached runtimes. Each costs ~15-20MB (with shared clients).
--- a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/routes/chat.py
+++ b/ai_platform_engineering/dynamic_agents/src/dynamic_agents/routes/chat.py
@@ -1,6 +1,7 @@
 """Chat endpoint for Dynamic Agents with SSE streaming."""
 
 import logging
+from contextlib import AsyncExitStack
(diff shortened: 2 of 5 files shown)


subbaksh:
This was intentionally kept ephemeral, otherwise its very easy to DOS mongodb with a 3-line script (by generating random conv ids) but if there is a use case for it then sure. Just be careful esp. if you're providing this as a service :)

cisco-erilutz:
> This was intentionally kept ephemeral, otherwise its very easy to DOS mongodb with a 3-line script (by generating random conv ids) but if there is a use case for it then sure. Just be careful esp. if you're providing this as a service :)

updated to feature flag this, default false, so that mongodb is an option but not the default one

sriaradhyula (approved):

PR #1461, opened 2026-05-19
fix(slack-bot): overthink skip missing in _route_to_agent; pre-render boilerplate

cisco-erilutz:
## Summary

- `_route_to_agent` never set `is_overthink_message` in `client_context`, causing the `[DEFER]` skip-check (gated on this flag in `ai.py:885`) to never fire for auto-routed channel messages — `[DEFER]` was stripped to an empty string and a footer-only message was posted to Slack
- `OVERTHINK_BOILERPLATE` was inserted as a raw Jinja string into `client_context`; the backend's `_render_system_prompt` only does one render pass, so the `{% if client_context.is_overthink_message %}` conditional inside the boilerplate was emitted as literal text to the LLM instead of being resolved

## Changes

- Add `"is_overthink_message": True` to `_route_to_agent`'s `client_context` (`app.py:534`) — all messages routed here are root-level channel posts, never thread replies
- Pre-render `OVERTHINK_BOILERPLATE` through a `SandboxedEnvironment` in both `handle_mention` and `_route_to_agent` before inserting into `client_context`, so the conditional resolves on the bot side where `is_overthink_message` is already known

## Test plan

- [ ] Deploy to staging and post a code-review request to an overthink-enabled channel without `@mention` — bot should silently skip (no message posted)
- [ ] Post a non-deferred question to an overthink-enabled channel — bot should respond normally
- [ ] `@mention` the bot in a thread — `is_overthink_message=False` path should still render the direct-response branch of the boilerplate correctly
- [ ] Verify no regressions in `make test` / `make lint`

### DIFF
--- a/ai_platform_engineering/integrations/slack_bot/app.py
+++ b/ai_platform_engineering/integrations/slack_bot/app.py
@@ -534,7 +534,7 @@ def _route_to_agent(event, say, client, channel_config, agent_match, is_bot, bot
       "channel_topic": channel_info.get("topic", ""),
       "channel_purpose": channel_info.get("purpose", ""),
       "overthink": bool(overthink and overthink.enabled),
-      "is_overthink_message": True,
+      "is_overthink_message": bool(overthink and overthink.enabled),
       "timestamp": thread_ts,
     }
     if overthink and overthink.enabled:
--- a/ai_platform_engineering/integrations/slack_bot/app.py
+++ b/ai_platform_engineering/integrations/slack_bot/app.py
@@ -339,8 +339,6 @@ def handle_mention(event, say, client):
         agent_match = next((a for a in channel_config.agents if a.agent_id == owner_id), None)
         agent_id = owner_id
 
-    overthink = agent_match.users.overthink if agent_match and agent_match.users else None
-
     # Build thread context: full on first interaction, delta on follow-ups
     context_message = message_text
     if event.get("thread_ts"):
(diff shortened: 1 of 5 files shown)


ai_platform_engineering/integrations/slack_bot/app.py
```diff
@@ -533,10 +534,11 @@ def _route_to_agent(event, say, client, channel_config, agent_match, is_bot, bot
       "channel_topic": channel_info.get("topic", ""),
       "channel_purpose": channel_info.get("purpose", ""),
       "overthink": bool(overthink and overthink.enabled),
+      "is_overthink_message": True,
```
cisco-erilutz: should this be `overthink` based on the variable above?

cisco-erilutz: Good catch — changed to `bool(overthink and overthink.enabled)` to mirror the `"overthink"` field directly above it.

kevkantes (dismissed):

ai_platform_engineering/integrations/slack_bot/app.py
cisco-erilutz: we should just omit the overthink_config here, and also remove the otherthink var, because we don't do overthink on app_mention responses

cisco-erilutz: Good call — removed the `overthink` variable entirely from `handle_mention`, pass `overthink_config=None` to `_call_ai`, and set `"overthink": False` in `client_context`. Also dropped the `followup_prompt` injection since that was only relevant to the overthink path.

cisco-erilutz:
Superseded by #1466 (prebuild branch for testing)

PR #1466, opened 2026-05-19
fix(slack-bot): overthink skip logic missing in _route_to_agent; pre-render boilerplate

cisco-erilutz:
## Summary

- `_route_to_agent` never set `is_overthink_message` in `client_context`, causing the `[DEFER]` skip-check to never fire for auto-routed channel messages — `[DEFER]` was stripped to empty and a footer-only message was posted to Slack
- `OVERTHINK_BOILERPLATE` now set directly at each call site (no Jinja) — `""` for `@mention` responses, the full boilerplate for overthink channel messages
- `handle_mention` no longer passes `overthink_config` — app_mentions never use overthink filtering
- `_post_final_response` now suppresses the post if text is empty after stripping confidence markers

## Test plan

- [ ] Deploy prebuild image and post a code-review request to an overthink-enabled channel without `@mention` — bot should silently skip (no message posted)
- [ ] Post a non-deferred question to an overthink-enabled channel — bot should respond normally
- [ ] `@mention` the bot in a thread — should respond directly with no overthink filtering
- [ ] `make lint && make test` pass

Supersedes #1461

### DIFF
--- a/ai_platform_engineering/integrations/slack_bot/app.py
+++ b/ai_platform_engineering/integrations/slack_bot/app.py
@@ -534,7 +534,7 @@ def _route_to_agent(event, say, client, channel_config, agent_match, is_bot, bot
       "channel_topic": channel_info.get("topic", ""),
       "channel_purpose": channel_info.get("purpose", ""),
       "overthink": bool(overthink and overthink.enabled),
-      "is_overthink_message": True,
+      "is_overthink_message": bool(overthink and overthink.enabled),
       "timestamp": thread_ts,
     }
     if overthink and overthink.enabled:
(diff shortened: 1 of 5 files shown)


kevkantes (dismissed):

sriaradhyula (dismissed):

ai_platform_engineering/integrations/slack_bot/app.py
```diff
@@ -367,11 +367,9 @@ def handle_mention(event, say, client):
       "channel_topic": channel_info.get("topic", ""),
       "channel_purpose": channel_info.get("purpose", ""),
       "humble_followup": is_humble_followup,
-      "overthink": bool(overthink and overthink.enabled),
-      "is_overthink_message": not is_thread_reply,
+      "overthink": False,
```
kevkantes, line 370: why is this hardcoded to false? Shouldn't this change based on the channel/message/scenario?

cisco-erilutz: This is in the @mention handler (expand to line 271 above)

kevkantes: ah yup. This may be worth a comment, but not blocking

kevkantes (approved):

PR #1488, opened 2026-05-21
fix(rag): handle list[dict] return type in _truncate_output

cisco-erilutz:
## Summary

- `_truncate_output` in `rag_tools.py` checked `isinstance(result, str)` before applying the length cap, but MCP tools (via `langchain_mcp_adapters`) return `list[dict]` content blocks, not plain strings
- The guard was always `False`, so the `RAG_MAX_OUTPUT_CHARS` truncation never fired — RAG tool results were returned at full size regardless of the configured limit
- Fix: serialize non-`str` results with `json.dumps` before the length check, so truncation applies to all return types

## Changes

- `ai_platform_engineering/multi_agents/platform_engineer/rag_tools.py`: add `import json`; update `_truncate_output` to handle `Any` input, serializing non-str values before truncating
- `tests/test_rag_tools_hard_stop.py`: add 3 regression tests covering `list[dict]` truncation path for both `SearchCapWrapper` and `FetchDocumentCapWrapper`

## Test plan

- [ ] `PYTHONPATH=. uv run pytest tests/test_rag_tools_hard_stop.py::TestOutputTruncation -v` — all 6 tests pass (3 existing str tests + 3 new list[dict] tests)
- [ ] Existing truncation tests still pass (no regression)

### DIFF
--- a/ai_platform_engineering/multi_agents/platform_engineer/rag_tools.py
+++ b/ai_platform_engineering/multi_agents/platform_engineer/rag_tools.py
@@ -18,6 +18,7 @@ ToolInvocationError caused the model to retry with different arguments, which
 defeated the cap.
 """
 
+import json
 import logging
 import os
 import threading
@@ -85,9 +86,23 @@ class _CapCounterMixin:
       return None
 
   @staticmethod
-  def _truncate_output(result: str, tool_name: str, max_chars: int = _DEFAULT_MAX_OUTPUT_CHARS) -> str:
-    """Truncate tool output to prevent context window overflow."""
-    if isinstance(result, str) and len(result) > max_chars:
+  # assisted-by Codex Codex-sonnet-4-6
+  def _truncate_output(result: Any, tool_name: str, max_chars: int = _DEFAULT_MAX_OUTPUT_CHARS) -> str:
+    """Truncate tool output to prevent context window overflow.
+
+    MCP tools (via langchain_mcp_adapters) return list[dict] content blocks
+    rather than plain str. Normalize to str before applying the length cap.
+    """
+    if isinstance(result, list):
+      text_parts = [
+        item["text"]
+        for item in result
+        if isinstance(item, dict) and isinstance(item.get("text"), str)
+      ]
+      result = "\n\n".join(text_parts) if text_parts else json.dumps(result, default=str)
+    elif not isinstance(result, str):
+      result = json.dumps(result, default=str)
+    if len(result) > max_chars:
       logger.info(f"{tool_name} output truncated: {len(result)} -> {max_chars} chars")
       return result[:max_chars] + f"\n\n[Output truncated — {len(result) - max_chars} chars omitted. Use the information above to answer.]"
     return result
--- a/tests/test_rag_tools_hard_stop.py
+++ b/tests/test_rag_tools_hard_stop.py
@@ -440,3 +440,42 @@ class TestOutputTruncation:
             result = await wrapper._arun(query="small query")
         assert result == small_result
         assert "truncated" not in result
+
+    @pytest.mark.asyncio
+    async def test_search_truncates_list_dict_result(self):
+        """SearchCapWrapper truncates list[dict] MCP content blocks (the real return type)."""
+        from ai_platform_engineering.multi_agents.platform_engineer import rag_tools as m
+        # Simulate MCP tool returning list[dict] content blocks with large text
+        big_text = "x" * (m._DEFAULT_MAX_OUTPUT_CHARS + 5000)
+        list_result = [{"type": "text", "text": big_text}]
+        wrapper, original = _make_search_wrapper(max_calls=5)
+        original.arun = AsyncMock(return_value=list_result)
+        with _patch_thread("thread-list-trunc"):
+            result = await wrapper._arun(query="big query")
+        assert isinstance(result, str)
+        assert len(result) <= m._DEFAULT_MAX_OUTPUT_CHARS + len("\n\n[Output truncated — XXXXX chars omitted. Use the information above to answer.]")
+        assert "truncated" in result.lower()
+
+    @pytest.mark.asyncio
+    async def test_fetch_truncates_list_dict_result(self):
+        """FetchDocumentCapWrapper truncates list[dict] MCP content blocks."""
+        from ai_platform_engineering.multi_agents.platform_engineer import rag_tools as m
+        big_text = "y" * (m._DEFAULT_MAX_OUTPUT_CHARS + 3000)
+        list_result = [{"type": "text", "text": big_text}]
+        wrapper, original = _make_fetch_wrapper(max_calls=5)
+        original.arun = AsyncMock(return_value=list_result)
+        with _patch_thread("thread-fetch-list-trunc"):
+            result = await wrapper._arun(document_id="doc-big")
+        assert isinstance(result, str)
+        assert "truncated" in result.lower()
+
+    @pytest.mark.asyncio
+    async def test_search_list_dict_under_limit_not_truncated(self):
+        """Small list[dict] results are serialized and returned without truncation marker."""
+        list_result = [{"type": "text", "text": "small content"}]
+        wrapper, original = _make_search_wrapper(max_calls=5)
+        original.arun = AsyncMock(return_value=list_result)
+        with _patch_thread("thread-list-no-trunc"):
+            result = await wrapper._arun(query="small query")
+        assert isinstance(result, str)
+        assert "truncated" not in result
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "ai-platform-engineering"
-version = "0.4.15-dev.2"
+version = "0.4.15-dev.3"
 description = "Multi-Agent Collaboration & Workflow Automation - AI agents collaborating across tools and teams to deliver high quality results"
 authors = [
   {name = "CNOE Contributors"}
--- a/uv.lock
+++ b/uv.lock
@@ -67,7 +67,7 @@ wheels = [
 
 [[package]]
 name = "ai-platform-engineering"
-version = "0.4.15.dev2"
+version = "0.4.15.dev3"
 source = { virtual = "." }
 dependencies = [
     { name = "a2a-sdk" },


ai_platform_engineering/multi_agents/platform_engineer/rag_tools.py
```diff
@@ -85,9 +86,16 @@ def _check_and_increment(self, thread_id: str, max_calls: int) -> int | None:
...
-    if isinstance(result, str) and len(result) > max_chars:
+  def _truncate_output(result: Any, tool_name: str, max_chars: int = _DEFAULT_MAX_OUTPUT_CHARS) -> str:
+    """Truncate tool output to prevent context window overflow.
+
+    MCP tools (via langchain_mcp_adapters) return list[dict] content blocks
+    rather than plain str. Serialize to str before applying the length cap so
+    the truncation actually fires regardless of return type.
+    """
+    if not isinstance(result, str):
+      result = json.dumps(result)
```
sriaradhyula: This fixes the immediate bug, but I’m worried about making `json.dumps()` the primary path for every non-string result. For MCP content blocks this changes the return shape from `list[dict]` into literal JSON text, and when the cap fires the returned value can be sliced mid-object/mid-string. I’d rather special-case content blocks / RAG-shaped objects and truncate known text-bearing fields (`text`, `text_content`, `page_content`, `content`) while preserving metadata and block structure, using serialization only as the unknown-shape fallback. Priority: P1. Confidence: high for the contract-change risk, medium on the exact downstream impact without an end-to-end trace.

tests/test_rag_tools_hard_stop.py
```diff
@@ -440,3 +440,43 @@ async def test_search_output_not_truncated_when_under_limit(self):
             result = await wrapper._arun(query="small query")
         assert result == small_result
         assert "truncated" not in result
+
+    @pytest.mark.asyncio
+    async def test_search_truncates_list_dict_result(self):
+        """SearchCapWrapper truncates list[dict] MCP content blocks (the real return type)."""
+        from ai_platform_engineering.multi_agents.platform_engineer import rag_tools as m
+        import json
```
sriaradhyula: This import is unused and `ruff check --select F401 --stdin-filename tests/test_rag_tools_hard_stop.py -` flags it as `F401`. Please remove it so the Python quality gate stays green. Priority: P2. Confidence: high.

sriaradhyula (commented):
Thanks for tracking this down. I agree with the diagnosis: the existing cap was checking for `str`, while the MCP adapter path returns content blocks / structured objects, so the guard never fired in the real RAG path.

My recommendation is to treat this PR as a good emergency brake, but tighten it before calling it the durable fix. The main issue is that serializing every non-string result changes the tool-output contract and can produce sliced JSON when truncation happens. I’d prefer content-block-aware/object-aware truncation that preserves useful metadata and truncates known large text fields.

Process note: the commits also need DCO sign-off, and because the PR says this was generated with Claude Code, please add the repo-required `assisted-by Codex Codex-sonnet-4-6` attribution comment in the touched code path.

ai_platform_engineering/multi_agents/platform_engineer/rag_tools.py
```diff
@@ -85,9 +86,16 @@ def _check_and_increment(self, thread_id: str, max_calls: int) -> int | None:
...
+    if len(result) > max_chars:
       logger.info(f"{tool_name} output truncated: {len(result)} -> {max_chars} chars")
       return result[:max_chars] + f"\n\n[Output truncated — {len(result) - max_chars} chars omitted. Use the information above to answer.]"
     return result
```
sriaradhyula, line 108: Suggested shape-preserving implementation:

```suggestion
  def _truncate_output(result: Any, tool_name: str, max_chars: int = _DEFAULT_MAX_OUTPUT_CHARS) -> Any:
    """Truncate text-bearing tool output while preserving structured result shape."""
    text_keys = {"text", "text_content", "page_content", "content"}
    remaining = max_chars
    omitted_chars = 0

    def _safe_len(value: Any) -> int:
      try:
        return len(json.dumps(value, default=str))
      except TypeError:
        return len(str(value))

    def _marker(omitted: int) -> str:
      return f"\n\n[Output truncated — {omitted} chars omitted. Use the information above to answer.]"

    def _truncate_text(value: str) -> str:
      nonlocal remaining, omitted_chars
      if len(value) <= remaining:
        remaining -= len(value)
        return value

      kept_chars = max(remaining, 0)
      omitted = len(value) - kept_chars
      omitted_chars += omitted
      remaining = 0
      if kept_chars == 0:
        return _marker(omitted).lstrip()
      return value[:kept_chars] + _marker(omitted)

    def _truncate_value(value: Any) -> Any:
      nonlocal omitted_chars
      if isinstance(value, str):
        return _truncate_text(value)
      if isinstance(value, list):
        truncated = []
        for item in value:
          if remaining <= 0:
            omitted_chars += _safe_len(value[len(truncated):])
            break
          truncated.append(_truncate_value(item))
        return truncated
      if isinstance(value, dict):
        return {
          key: _truncate_text(item)
          if key in text_keys and isinstance(item, str)
          else _truncate_value(item)
          if isinstance(item, (dict, list))
          else item
          for key, item in value.items()
        }
      return value

    if isinstance(result, str):
      truncated = _truncate_text(result)
    elif isinstance(result, (dict, list)):
      if _safe_len(result) <= max_chars:
        return result
      truncated = _truncate_value(result)
    else:
      try:
        serialized = json.dumps(result, default=str)
      except TypeError:
        serialized = str(result)
      truncated = _truncate_text(serialized)

    if omitted_chars:
      logger.info(f"{tool_name} output truncated: omitted {omitted_chars} chars")
    return truncated
```

tests/test_rag_tools_hard_stop.py
sriaradhyula: Suggested fix:

```suggestion
```

tests/test_rag_tools_hard_stop.py
```diff
@@ -440,3 +440,43 @@ async def test_search_output_not_truncated_when_under_limit(self):
...
+        # Simulate MCP tool returning list[dict] content blocks with large text
+        big_text = "x" * (m._DEFAULT_MAX_OUTPUT_CHARS + 5000)
+        list_result = [{"type": "text", "text": big_text}]
+        wrapper, original = _make_search_wrapper(max_calls=5)
+        original.arun = AsyncMock(return_value=list_result)
+        with _patch_thread("thread-list-trunc"):
+            result = await wrapper._arun(query="big query")
+        assert isinstance(result, str)
+        assert len(result) <= m._DEFAULT_MAX_OUTPUT_CHARS + len("\n\n[Output truncated — XXXXX chars omitted. Use the information above to answer.]")
+        assert "truncated" in result.lower()
```
sriaradhyula, line 457: If we preserve MCP content-block shape, this test should assert the nested text was truncated instead of asserting the whole result became a string.

```suggestion
        assert isinstance(result, list)
        assert result[0]["type"] == "text"
        assert len(result[0]["text"]) < len(big_text)
        assert "truncated" in result[0]["text"].lower()
```

tests/test_rag_tools_hard_stop.py
```diff
@@ -440,3 +440,43 @@ async def test_search_output_not_truncated_when_under_limit(self):
...
+        """FetchDocumentCapWrapper truncates list[dict] MCP content blocks."""
+        from ai_platform_engineering.multi_agents.platform_engineer import rag_tools as m
+        big_text = "y" * (m._DEFAULT_MAX_OUTPUT_CHARS + 3000)
+        list_result = [{"type": "text", "text": big_text}]
+        wrapper, original = _make_fetch_wrapper(max_calls=5)
+        original.arun = AsyncMock(return_value=list_result)
+        with _patch_thread("thread-fetch-list-trunc"):
+            result = await wrapper._arun(document_id="doc-big")
+        assert isinstance(result, str)
+        assert "truncated" in result.lower()
```
sriaradhyula, line 470: Same here: keep the content block and assert truncation on the text payload.

```suggestion
        assert isinstance(result, list)
        assert result[0]["type"] == "text"
        assert len(result[0]["text"]) < len(big_text)
        assert "truncated" in result[0]["text"].lower()
```

tests/test_rag_tools_hard_stop.py
```diff
@@ -440,3 +440,43 @@ async def test_search_output_not_truncated_when_under_limit(self):
...
+    @pytest.mark.asyncio
+    async def test_search_list_dict_under_limit_not_truncated(self):
+        """Small list[dict] results are serialized and returned without truncation marker."""
+        list_result = [{"type": "text", "text": "small content"}]
+        wrapper, original = _make_search_wrapper(max_calls=5)
+        original.arun = AsyncMock(return_value=list_result)
+        with _patch_thread("thread-list-no-trunc"):
+            result = await wrapper._arun(query="small query")
+        assert isinstance(result, str)
+        assert "truncated" not in result
```
sriaradhyula, line 481: For under-limit content blocks, preserving the original object is the behavior I’d want to lock in.

```suggestion
        assert result == list_result
        assert "truncated" not in result[0]["text"]
```

sriaradhyula (commented):
Adding concrete inline suggestions for the behavior I described above. The main suggested implementation keeps the current hotfix intent, but preserves structured MCP/RAG result shapes and only uses JSON serialization as the unknown-shape fallback.

ai_platform_engineering/multi_agents/platform_engineer/rag_tools.py
sriaradhyula, line 108: I’d make the MCP content-block path explicit here and keep JSON serialization as the fallback. This avoids feeding the model raw JSON wrappers in the common `list[{"type": "text", "text": ...}]` case, while still enforcing the cap for unknown structured outputs.

```suggestion
  def _truncate_output(result: Any, tool_name: str, max_chars: int = _DEFAULT_MAX_OUTPUT_CHARS) -> str:
    """Truncate tool output to prevent context window overflow."""
    if isinstance(result, list):
      text_blocks = [
        item["text"]
        for item in result
        if isinstance(item, dict) and isinstance(item.get("text"), str)
      ]
      result = "\n\n".join(text_blocks) if text_blocks else json.dumps(result, default=str)
    elif not isinstance(result, str):
      result = json.dumps(result, default=str)

    if len(result) > max_chars:
      logger.info(f"{tool_name} output truncated: {len(result)} -> {max_chars} chars")
      return result[:max_chars] + f"\n\n[Output truncated — {len(result) - max_chars} chars omitted. Use the information above to answer.]"
    return result
```

sriaradhyula (commented):
Adding one focused suggestion for `_truncate_output` that avoids JSON as the default path for MCP content blocks.

ai_platform_engineering/multi_agents/platform_engineer/rag_tools.py
cisco-erilutz: fixing per our convo!

sriaradhyula (approved):

PR #1492, opened 2026-05-22
fix(ui): default interrupt_on to empty object instead of undefined in seed config

kevkantes:
# Description

Changes the default value of `interrupt_on` in the agent seed config from `undefined` to `{}` (empty object). This ensures downstream code that expects an object type for `interrupt_on` doesn't receive `undefined`, preventing potential runtime errors or unexpected behavior when seeding agents.

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/mongo.py
+++ b/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/mongo.py
@@ -21,6 +21,17 @@ from dynamic_agents.models import (
 logger = logging.getLogger(__name__)
 
 
+def _strip_nulls(doc: dict) -> dict:
+    """Strip None values from a MongoDB document before pydantic construction.
+
+    Pydantic only applies default_factory when a key is absent — an explicit
+    None passed as a kwarg bypasses the default. Stripping nulls here lets
+    fields like interrupt_on recover their default when stored as null in
+    older documents.
+    """
+    return {k: v for k, v in doc.items() if v is not None}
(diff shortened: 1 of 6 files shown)


subbaksh (dismissed):

ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/mongo.py
```diff
@@ -21,6 +21,17 @@
 logger = logging.getLogger(__name__)
 
 
+def _strip_nulls(doc: dict) -> dict:
```
kevkantes, line 24: Is this a one time thing we can remove after it runs? Not sure if this would be considered defensive or a hack lol

cisco-erilutz: no this is a defensive thing. missing values get stored as nulls in Mongo, and this strips those nulls out so that pydantic can default them again. This applies to all fields going forward

The root cause was that the interrupt_on was defaulted to null in the database, but was included as `None` when it got loaded, and an explicit None doesn't trigger the pydantic default factories and None isn't allowed so it crashed everything

This would happen again on any new field added that needs a default factory and doesn't allow none

cisco-erilutz:
Closing in favour of a prebuild-branch PR to get a test image.

PR #1502, opened 2026-05-22
feat(dynamic-agents): add sliding-window 5xx error rate check to /readyz

cisco-erilutz:
## Summary

- Adds a thread-safe sliding-window (last 20 requests) to `PrometheusHTTPMiddleware` that tracks whether each response was a 5xx/error
- Updates `/readyz` to return 503 if ≥50% of the window are errors (with a minimum floor of 5 samples so startup is not penalised)
- Updates the `dynamic-agents` Helm chart default readiness probe to point at `/readyz` instead of `/healthz`

## Motivation

Previously `/readyz` only checked MongoDB connectivity. This meant pods that were technically "up" but returning 500s on real traffic would still pass readiness, causing Kubernetes to cut over traffic to broken pods during a rollout.

With `maxUnavailable: 0`, a pod that fails readiness is never promoted to serve traffic — old pods stay up and the rollout stalls until the issue is resolved or rolled back.

## Test plan

- [ ] Start the dynamic-agents service locally, send >5 requests that return 5xx — confirm `GET /readyz` returns 503
- [ ] Verify that non-5xx traffic (or fewer than 5 requests total) does not fail readiness
- [ ] Verify `/healthz` still returns 200 regardless of error rate (liveness probe unaffected)
- [ ] Deploy to dev environment and confirm rolling update stalls when a bad image is deployed

### DIFF
--- a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/metrics/http_middleware.py
+++ b/ai_platform_engineering/dynamic_agents/src/dynamic_agents/metrics/http_middleware.py
@@ -5,6 +5,8 @@ No JWT parsing — auth is handled by the UI gateway in 0.4.0.
 
 import logging
 import time
+from collections import deque
+from threading import Lock
 
 from starlette.middleware.base import BaseHTTPMiddleware
 from starlette.requests import Request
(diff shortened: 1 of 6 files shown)


cisco-erilutz:
Closing — the readiness-probe approach can't reliably detect 500s before the rollout commits. Pivoting to Flagger for proper blue/green traffic shifting with Prometheus-gated promotion.

PR #1577, opened 2026-05-27
feat(infra): compose hot/prod split + login-time team auto-create + Hello-World seed + chunked transactional OpenFGA sync

sriaradhyula:
## Summary

Four foundational changes that grew up around the local-dev Compose stack and the OpenFGA identity-sync reconciler. None of them depend on each other — bundled together because they all moved through dev/test on the same morning and they all touch local-dev / first-boot ergonomics.

### Compose: split `caipe-ui` into hot and prod-parity sibling services (`95d6b58`)

Splits the single `caipe-ui` service into:
- `caipe-ui` (default): `next dev` with hot-reload + bind-mounted `./ui/src`. Used by `make caipe-ui-hot`.
- `caipe-ui-prod` (profile `caipe-ui-prod`): `next build` + `next start`, no bind mount. Used by `make caipe-ui-prod`.

Both publish on `:3000`. Keycloak's `caipe-ui` OIDC client only allow-lists `localhost:3000/*` as a redirect, so the targets stop and restart the sibling before switching modes.

### feat(rbac): opt-in login-time team auto-creation (`179c5ad`)

Adds `IDENTITY_SYNC_LOGIN_AUTO_CREATE_TEAMS` env flag. When `true`, the OIDC claim reconciler will create missing CAIPE teams on first login (matching the user's IdP groups) instead of waiting for the periodic sync. Defaults to `false` — opt-in only — so existing deployments see no behaviour change.

### feat(seed): bootstrap a Hello-World agent and auto-create-teams sync rule on fresh install (`80c4d41`)

Adds `ui/src/lib/seed-config.ts` which, on a fresh deployment (empty `dynamic_agents` and `identity_sync_rules` collections), seeds:
1. A `Hello World` Dynamic Agent so first-time admins have something to click.
2. An `auto-create-teams-from-groups` IdP sync rule so OIDC groups become CAIPE teams automatically.

Both seeds are idempotent and skip when the collections already have rows.

### fix(rbac): chunk OpenFGA writes and make identity-group-sync apply transactional (`60b97d7`)

Two correctness fixes that surfaced when an org admin had 800+ users to sync:
- **Chunking**: `openfga.ts` now splits `write` / `delete` calls into 50-tuple chunks. OpenFGA HTTP responses started rejecting batches over 100.
- **Transactional apply**: `identity-group-sync-reconciler.ts` now collects the full plan, then applies writes-then-deletes as one logical transaction so a mid-sync failure leaves OpenFGA in a consistent state.

## Test plan

- [x] `npx jest --testPathPatterns "auth-config|oidc-claim-reconciler|seed-config-default-agent|sync-reconciler|openfga-chunking"` — **93 passed**.
- [ ] Manual: `make caipe-ui-hot` then `make caipe-ui-prod`, confirm both publish on `:3000` and Keycloak login still works.
- [ ] Manual: set `IDENTITY_SYNC_LOGIN_AUTO_CREATE_TEAMS=true`, log in as a new IdP user, confirm new CAIPE teams appear in the Admin UI.
- [ ] Manual: wipe the local Mongo, restart the UI, confirm a `Hello World` agent and an `auto-create-teams-from-groups` sync rule are seeded.

Assisted-by: Claude:claude-opus-4-7

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -240,12 +240,12 @@ caipe-ui-hot: ## Run CAIPE UI in Docker with hot reload (next dev + bind-mounted
 	@echo "Starting CAIPE UI in hot-reload mode (next dev)..."
 	@echo "  - Edits in ui/src trigger sub-second rebuild via next dev"
 	@echo "  - public/ asset changes still need: make caipe-ui-hot (rebuilds image)"
+	@# Bring down the prod-parity sibling first — both bind host port 3000 and
+	@# Keycloak's caipe-ui client only allow-lists localhost:3000/* as a redirect
+	@# URI (see deploy/keycloak/realm-config.json), so the two services are
+	@# mutually exclusive at runtime.
+	@docker compose -f docker-compose.dev.yaml --profile caipe-ui-prod rm -sf caipe-ui-prod 2>/dev/null || true
 	$(DOCKER_COMPOSE_BUILD_ENV) \
-	CAIPE_UI_MODE=hot \
-	CAIPE_UI_BUILD_TARGET=dev \
-	CAIPE_UI_NODE_ENV=development \
-	CAIPE_UI_COMMAND="npm run dev" \
-	CAIPE_UI_IMAGE_SUFFIX=-dev \
 	docker compose -f docker-compose.dev.yaml --profile caipe-ui up -d --build caipe-ui
 	@echo ""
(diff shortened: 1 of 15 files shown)


sriaradhyula:
Superseded by #1583 — bundled into a single combined PR per maintainer decision (single-PR strategy).

PR #1579, opened 2026-05-27
fix(operators): surface LLM config + slack RBAC denials, add PDP coverage docs, compose alias

sriaradhyula:
## Summary

A bundle of operator-facing fixes and docs that surfaced during the canonical-team-membership / RBAC work. Grouped together because they're small, independent, and ship cleaner as one PR than four.

### `fix(dynamic-agents)`: surface LLM config errors instead of generic chat failure (`b83dc6d`)

When an agent had no usable LLM config (no `model_id`, no `provider`, or provider missing required API key), the chat endpoint silently emitted `GENERIC_AGENT_ERROR` ("Something went wrong"), so operators had no hint that the actual fix was `MODEL_ID` / `OPENAI_API_KEY` / similar.

- `chat.py`: catch the new `LlmConfigError` and re-raise with a typed `LLM_CONFIG_ERROR` code that the UI can map to a "fix your LLM config" toast.
- `services/llm_clients.py`: validate config up front and raise `LlmConfigError` with the missing field; new fallback path for `provider="custom"` agents that lack an explicit `base_url`.
- 11 new tests in `test_chat_llm_config_error.py` + `test_llm_clients_fallback.py`.

### `docs(rbac)`: add PDP coverage audit + fine-grained-rbac spec (`9f2ed86`)

Documents the gap discovered while debugging "Generic user sees wrong default agent in Admin UI": the BFF's legacy `withAuth(...)` helper silently routes every `/api/admin/*` GET to `admin_ui#view` (admins-only), bypassing each route handler's own fine-grained PDP check. The new `pdp-coverage-audit.md` enumerates every `/api/admin/*` route, what `withAuth` resolves it to, what the handler then checks, and the resulting coverage gap. The follow-up spec (`2026-05-27-fine-grained-rbac-for-withauth-routes/`) proposes per-route allow-list overrides — independent of this PR.

### `build(compose)`: add `caipe-ui` network alias to `caipe-ui-prod` (`7753069`)

The `caipe-ui-prod` profile (added in PR A #1577) was reachable from the browser on host `:3000` but unreachable from sibling containers (slack-bot, supervisor) because those services resolve `http://caipe-ui:3000` by container name. Adds a `caipe-ui` network alias to the prod sibling so sibling-to-sibling traffic resolves to whichever sibling is currently running.

### `fix(slack-bot)`: surface RBAC denials and stop silent retries (`11ac0c8`)

The Slack bot's global RBAC middleware short-circuited every deny / silence / dedupe path with a bare `return`. Bolt-python logged `"middleware skipped calling next() without responding"` (HTTP 500 → Slack retries up to 3×), so a single denied event became 3 noisy retries and operators got zero feedback.

- Replace bare `return` with `ack()` so bolt knows the event is handled.
- For denied events, post an ephemeral message to the channel with the deny reason ("You don't have access to this agent in this channel.") so users get fast, actionable feedback.
- Setup-mode middleware now stops handlers BEFORE they respond.
- 9 tests pass (3 existing + 6 new in `test_slack_route_no_silent_failure.py`).

## Test plan

- [x] `cd ai_platform_engineering/dynamic_agents && uv run pytest tests/test_chat_llm_config_error.py tests/test_llm_clients_fallback.py` — **11 passed**.
- [x] `cd ai_platform_engineering/integrations/slack_bot && uv run pytest tests/test_slack_route_no_silent_failure.py` — **9 passed**.
- [ ] Manual: create a Dynamic Agent with no `model_id`, send a chat message, confirm the UI shows the typed LLM-config error toast (not the generic "Something went wrong").
- [ ] Manual: bring up `caipe-ui-prod` profile, confirm slack-bot can reach `http://caipe-ui:3000` from its container.
- [ ] Manual: send a Slack message to an agent the user lacks access to, confirm ONE ephemeral deny message (not 3 retries) and bot logs show `"deny"` not 500.

## Dependencies

- Stacked on top of #1577 (compose split) because `caipe-ui-prod` only exists once #1577 lands. Will rebase cleanly onto `main` after #1577 merges.

Assisted-by: Claude:claude-opus-4-7


sriaradhyula:
Superseded by #1583 — bundled into a single combined PR per maintainer decision (single-PR strategy).

PR #1662, opened 2026-06-01
feat(setup): persist Keycloak + OpenFGA on a shared Postgres

sriaradhyula:
## Summary

By default `setup-caipe.sh` left **Keycloak on embedded H2** and **OpenFGA on an in-memory store**, so every pod restart wiped all identity, realm, and authorization state — making RBAC unusable beyond a throwaway demo.

This deploys a single shared `bitnami/postgresql` instance (`caipe-postgres`) that backs both, with one role + database per consumer (`keycloak`, `openfga`, optional `litellm`).

- **Shared Postgres**: generated passwords persisted in the `caipe-postgres-credentials` Secret (idempotent reuse on re-runs); `initdb` script creates a login role + owned DB per consumer; consumer-facing `caipe-keycloak-db` / `caipe-openfga-db` Secrets emitted each run.
- **Keycloak**: `KC_DB` / `KC_DB_URL` / `KC_DB_USERNAME` via `keycloak.env`, and a new **`keycloak.db.passwordSecret`** chart hook so `KC_DB_PASSWORD` is injected via `secretKeyRef` (never rendered into values/release data).
- **OpenFGA**: `datastore.engine=postgres` + `datastore.uriSecretRef` (verified on both the deployment and the `migrate` Job).
- Deployed automatically when the RBAC runtime (or `--litellm-db`) is enabled; **`--no-shared-postgres`** restores the previous ephemeral H2/in-memory behavior.
- RBAC `architecture.md` updated per the living-documentation rule.

## Test plan

- [x] `bash -n setup-caipe.sh` clean
- [x] `helm template` of the keycloak subchart renders `KC_DB_PASSWORD` via `secretKeyRef` + plaintext `KC_DB*` env
- [x] `helm template` of the openfga subchart renders `engine=postgres` + `uriSecretRef` on deployment **and** migrate Job
- [x] Generated RBAC values YAML valid in both populated and `--no-shared-postgres` (empty-block) paths
- [ ] Fresh `./setup-caipe.sh` install on a kind cluster; restart Keycloak + OpenFGA pods and confirm realm/users/tuples survive

PR #1663, opened 2026-06-01
feat(setup): opt-in unified LiteLLM front for chat + embeddings

sriaradhyula:
## Summary

Adds an **opt-in** `--litellm` mode to `setup-caipe.sh` that routes the chat model — and OpenAI/Azure embeddings — through the in-cluster LiteLLM proxy. CAIPE agents then talk to a single OpenAI-compatible endpoint, and upstream provider credentials live only in the proxy instead of being spread across the agent `llm-secret`.

- **`_finalize_litellm_mode`** captures the real chat/embeddings providers, then repoints the working `LLM_PROVIDER`/`EMBEDDINGS_PROVIDER` at the proxy. It **self-disables (with a warning)** on unsupported providers so a bare `--litellm` can never break an otherwise-valid install.
  - Chat: `anthropic-claude`, `openai`, `aws-bedrock`, `azure-openai`.
  - Embeddings: `openai`/`azure-openai` are fronted by the proxy; other embeddings providers (bedrock/cohere/huggingface/voyage) keep their native path so RAG still works.
- **`deploy_litellm`** now builds the proxy `model_list` from the real provider, requires a routing `master_key`, and reads upstream creds from a dedicated `litellm-upstream-secret`. The agent-facing `llm-secret` only ever points at the proxy. **vLLM mode is unchanged.**
- **`--litellm-db`** opts into a persistent LiteLLM database on the shared Postgres (from the base PR) for virtual keys / spend tracking; **DB-less is the default**.

> Stacked on #1662 (shared Postgres). Review/merge that first — the `--litellm-db` path reuses its `caipe-litellm-db` Secret + `litellm` database.

## Test plan

- [x] `bash -n setup-caipe.sh` clean
- [x] Rendered LiteLLM manifests parse as valid YAML in both unified mode (with `envFrom` + `general_settings`) and legacy vLLM mode (empty blocks); inner `config.yaml` parses with `caipe-chat`/`caipe-embeddings` model_list and `master_key`
- [ ] Live `./setup-caipe.sh --litellm` with an Anthropic key: confirm agents resolve `caipe-chat` through the proxy and complete a chat turn
- [ ] Live `--litellm` with OpenAI embeddings + `--rag`: confirm RAG embeds via the proxy
- [ ] `--litellm-db`: confirm the proxy connects to the `litellm` database and persists virtual keys

## Notes / scope

- Opt-in, default **OFF** — existing installs are unaffected.
- v1 embeddings routing covers OpenAI/Azure; Bedrock/Cohere/HF embeddings stay on their native path (follow-up).

### DIFF
--- a/docs/docs/getting-started/kind/setup.md
+++ b/docs/docs/getting-started/kind/setup.md
@@ -118,6 +118,9 @@ Credentials are read from `~/.config/claude.txt` and `~/.config/openai.txt` when
 | `--rag` | Deploy the RAG stack (knowledge base, embeddings) |
 | `--graph-rag` | Deploy Graph RAG (Neo4j + ontology agent; implies `--rag`) |
 | `--tracing` | Deploy Langfuse and enable tracing |
+| `--no-shared-postgres` | Skip the shared Postgres; use Keycloak embedded H2 + OpenFGA in-memory (ephemeral — RBAC state is lost on every pod restart). Persistent Postgres is the default for RBAC installs. |
+| `--litellm` | Route chat (and OpenAI/Azure embeddings) through an in-cluster LiteLLM proxy so agents use one OpenAI-compatible endpoint and upstream provider keys stay in the proxy. Supports anthropic/openai/aws-bedrock/azure-openai. |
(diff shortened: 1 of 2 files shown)


sriaradhyula:
Superseded by #1664 (both commits consolidated into one branch).

PR #1668, opened 2026-06-01
fix: validate LiteLLM MCP runtime config

MarwaOuledElHadjAli:
## Summary
- render the prod LiteLLM API URL by default when `litellmMcp.config.LITELLM_API_URL` is left empty
- add Helm fail-fast validation so `litellmMcp.enabled=true` requires a LiteLLM API token source
- document the required token source and rendered URL default

## Why
The LiteLLM MCP server imports its LiteLLM client at startup and requires both `LITELLM_API_URL` and a token (`LITELLM_API_KEY`, `LITELLM_TOKEN`, or `LITELLM_API_TOKEN`). The chart previously allowed rendering a pod with a blank URL and no secret source, which leads to startup failures in Argo-managed environments.

## Validation
- `git diff --check`
- `helm template` against a temporary minimal chart covering the LiteLLM MCP templates:
  - existing secret renders a ConfigMap with `https://litellm.prod.outshift.ai`
  - explicit URL override is preserved
  - missing token source fails fast
  - `secret.create=true` without a token key fails fast
  - external secret with `LITELLM_API_KEY` renders successfully

Note: the full umbrella chart render was not run because this local checkout is missing Helm dependency charts (`slim`, `slim-control-plane`, `rag-stack`).

### DIFF
--- a/charts/ai-platform-engineering/templates/litellm-mcp-configmap.yaml
+++ b/charts/ai-platform-engineering/templates/litellm-mcp-configmap.yaml
@@ -1,4 +1,5 @@
 {{- if .Values.litellmMcp.enabled }}
+{{- $config := .Values.litellmMcp.config | default dict }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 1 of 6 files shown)


MarwaOuledElHadjAli:
Superseded by deployment-values fix: https://github.com/cisco-eti/platform-apps-deployment/pull/526

The prod values already define `litellmMcp.config.LITELLM_API_URL`; the likely runtime issue is the shared `llm-secret` also injecting `LITELLM_API_URL` via `envFrom` and overriding the ConfigMap value. The deployment PR switches LiteLLM MCP to a dedicated ExternalSecret containing only `LITELLM_API_KEY`, so the endpoint remains controlled by the prod values file.

PR #1669, opened 2026-06-01
fix(keycloak): add view-realm and manage-realm to caipe-platform service account roles

kevkantes:
## Summary

- Adds `view-realm` and `manage-realm` to the `desired_roles` list in `init-idp.sh` for the `caipe-platform` service account
- Adds the same roles to `realm-config.json` so fresh Keycloak imports include them before init-idp runs

## Problem

On a fresh install (or any Keycloak pod restart, since H2 is ephemeral), the `caipe-platform` client credentials token cannot:
- Call `GET /users-management-permissions` → 403 (`view-realm` required)
- Perform any reconcile mutations → 403 (`manage-realm` required)

This causes the Keycloak admin invariant panel to show failures and reconcile-all to silently do nothing.

Previous deployments only appeared to work because accumulated H2 state from older init-idp runs had granted broader roles that persisted across pod restarts. Any fresh deploy or realm rename (which forces a restart) hits this immediately.

## Testing

Verified on a fresh Keycloak deploy with a renamed realm (`forge` instead of `caipe`) — invariant checks pass and reconcile-all succeeds after this change.

### DIFF
--- a/charts/ai-platform-engineering/charts/keycloak/realm-config.json
+++ b/charts/ai-platform-engineering/charts/keycloak/realm-config.json
@@ -66,7 +66,7 @@
     {
       "clientId": "caipe-ui",
       "name": "CAIPE Admin UI",
-      "description": "Next.js frontend — confidential OIDC client for Admin UI (NextAuth server-side flow)",
+      "description": "Next.js frontend \u2014 confidential OIDC client for Admin UI (NextAuth server-side flow)",
       "enabled": true,
       "publicClient": false,
       "secret": "caipe-ui-dev-secret",
@@ -138,7 +138,7 @@
     {
       "clientId": "caipe-platform",
       "name": "CAIPE Platform Resource Server",
-      "description": "Confidential service client — identity and token-exchange only; CAIPE authorization is enforced in OpenFGA.",
+      "description": "Confidential service client \u2014 identity and token-exchange only; CAIPE authorization is enforced in OpenFGA.",
       "enabled": true,
       "publicClient": false,
       "secret": "caipe-platform-dev-secret",
@@ -163,7 +163,7 @@
     {
       "clientId": "caipe-slack-bot",
       "name": "CAIPE Slack Bot",
-      "description": "Confidential client — service account with token-exchange permission for OBO (FR-018, FR-021)",
+      "description": "Confidential client \u2014 service account with token-exchange permission for OBO (FR-018, FR-021)",
       "enabled": true,
       "publicClient": false,
       "secret": "caipe-slack-bot-dev-secret",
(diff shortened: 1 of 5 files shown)


charts/ai-platform-engineering/charts/keycloak/scripts/init-idp.sh
```diff
@@ -150,7 +150,7 @@ seed_spec104_main || echo "[init-idp] [spec-104] seed had errors (see above)"
 # BEFORE the early-exit below so dev/CI stacks that don't wire IDP_ISSUER still
 # get the correct role mapping.
 _ensure_caipe_platform_user_roles() {
-  local desired_roles="view-users query-users manage-users query-clients view-clients manage-clients view-authorization manage-authorization"
+  local desired_roles="view-realm manage-realm view-users query-users manage-users query-clients view-clients manage-clients view-authorization manage-authorization"
```
sriaradhyula: `view-realm` is enough to fix the reported 403: `init-idp.sh` already performs the privileged `PUT /users-management-permissions` with the bootstrap superadmin token (lines ~1913-1916), so by the time the BFF runs the realm feature is enabled and it only needs to *read* it — the `enableUsersManagementPermissions` 403→GET fallback succeeds with `view-realm`.

`manage-realm` grants the `caipe-platform` service account full realm administration (token lifespans, realm settings, delete realm). That's a meaningful privilege bump it doesn't need here. Suggest dropping it unless we explicitly want to support BYO / externally-managed Keycloak where `init-idp.sh` never runs (in which case keep it and note that's why).

```suggestion
  local desired_roles="view-realm view-users query-users manage-users query-clients view-clients manage-clients view-authorization manage-authorization"
```

sriaradhyula (commented):
Diagnosis and fix look right — this is exactly the `readUsersManagementPermissionsAfterForbiddenEnable: 403` path (the BFF, as the `caipe-platform` service account, calls `/users-management-permissions`: GET needs `view-realm`, PUT needs `manage-realm`). One scoping suggestion inline: `view-realm` alone should be enough, and `manage-realm` is probably over-privileged for the chart-managed path. Details on the two lines below.

charts/ai-platform-engineering/charts/keycloak/realm-config.json
```diff
@@ -722,6 +722,8 @@
       "realmRoles": [],
       "clientRoles": {
         "realm-management": [
+          "manage-realm",
+          "view-realm",
```
sriaradhyula: Same reasoning as `init-idp.sh` — keep `view-realm`, drop `manage-realm` (over-privileged for the chart-managed path; the realm feature is enabled by `init-idp.sh` with the superadmin token).

```suggestion
          "view-realm",
```

charts/ai-platform-engineering/charts/keycloak/scripts/init-idp.sh
kevkantes: done

charts/ai-platform-engineering/charts/keycloak/realm-config.json
kevkantes: done

PR #1672, opened 2026-06-01
fix(slack-bot): stop intermediate agent narration leaking into final response

cisco-erilutz:
## Summary

Stacked on top of #1609 (the Anthropic Bedrock client switch). That switch surfaced a latent bug in the Slack bot's response reconstruction: intermediate agent narration ("Now let me search…", "I notice the pagination is returning the same data…") leaks into the final Slack message body, in front of the real answer.

## Root cause

The Slack bot rebuilds the final answer from the AG-UI SSE event stream (`integrations/slack_bot/utils/ai.py`), not from the `AIMessage` directly. The `ChatBedrockConverse` → `ChatAnthropicBedrock` change altered **which** intermediate text reaches the bot as `TEXT_MESSAGE_CONTENT` deltas — the Anthropic client emits per-step narration tokens that the converse path did not surface. That exposed two flaws:

1. `TEXT_MESSAGE_CONTENT` arriving **while a tool was active** (stream already open in todo/plan mode) was streamed live via `appendStream`, so every narration segment was written straight into the message body.
2. `accumulated_text` (which was never reset during a run) was used as a final-answer fallback, so the entire concatenated transcript could surface at finalization and on the `RUN_ERROR` recovery path.

## Fix

- Never stream text to the message body mid-run.
- Text emitted **while a tool is active** is treated as discardable narration — it only feeds the typing indicator.
- Text emitted **when no tool is active** is buffered and cleared on the next `TOOL_CALL_START`, so only the text after the last tool call (the real answer) is rendered at finalization.
- Removed the `accumulated_text` buffer entirely; finalization and error recovery now use the single answer buffer.

## Tests

- Added a regression test for narration emitted **while a tool is active** (the previous between-tools test didn't cover that path).
- Full `integrations/slack_bot` suite passes (179 tests).

### DIFF
--- a/ai_platform_engineering/integrations/slack_bot/tests/test_ai_plan_streaming.py
+++ b/ai_platform_engineering/integrations/slack_bot/tests/test_ai_plan_streaming.py
@@ -500,6 +500,68 @@ class TestThinkingBuffer:
     assert "Searching..." not in combined
     assert "Analyzing data..." not in combined
 
+  def test_narration_while_tool_active_does_not_leak(self):
+    """Regression (thinking-text leak): with the stream already open (todo mode),
+    narration that arrives *while a tool is still active* must not be streamed
+    into the message body.
+
+    This reproduces the ChatAnthropicBedrock regression: the LLM client emits
+    inter-step text deltas that arrive between TOOL_CALL_START and the matching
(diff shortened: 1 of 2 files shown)


cisco-erilutz:
Superseded by #1675 (moved to a prebuild/ branch so prebuild slack-bot images get published for testing). Same commit, same base.

PR #1705, opened 2026-06-03
fix(escalation): use fresh conversation for VictorOps on-call lookup

cisco-erilutz:
## Summary

- The VictorOps on-call lookup was reusing the channel thread's `conversation_id` when calling `stream_chat`, causing the VictorOps agent to inherit the full channel thread history as context
- Long-lived threads can accumulate 150K+ tokens of unrelated conversation, causing the model to reason incorrectly and return a wrong on-call email
- Fix: generate a fresh `uuid4` conversation inside `_ping_victorops_oncall` so each lookup starts clean with no prior context

## Changes

`ai_platform_engineering/integrations/slack_bot/utils/escalation.py`
- Import `uuid`
- Remove `conversation_id` param from `_ping_victorops_oncall` signature
- Generate `str(uuid.uuid4())` as the `conversation_id` for `stream_chat` inside `_ping_victorops_oncall`
- Update docstring on `execute_escalation` to clarify that `conversation_id` is used for feedback/metadata tracking only, not forwarded to the VictorOps query

## Test plan

- [ ] Trigger "Get help" on a thread with long history and verify the on-call email is correct
- [ ] Trigger "Get help" on a fresh thread and verify the on-call email is correct
- [ ] Verify feedback/metadata tracking still works (conversation still marked as escalated)

### DIFF
--- a/ai_platform_engineering/integrations/slack_bot/utils/escalation.py
+++ b/ai_platform_engineering/integrations/slack_bot/utils/escalation.py
@@ -10,6 +10,7 @@ Supports three configurable actions (all fire if enabled):
 """
 
 import re
+import uuid
 
 from loguru import logger
 
@@ -26,7 +27,6 @@ def execute_escalation(
   user_id: str,
   escalation_config: EscalationConfig,
   agent_id: str = "",
-  conversation_id: str = "",
 ):
   """Run all configured escalation actions.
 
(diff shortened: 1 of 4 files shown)


kevkantes (approved):

ai_platform_engineering/integrations/slack_bot/utils/escalation.py
cisco-erilutz: is this used? should this be removed entirely?

kevkantes (approved):

PR #1711, opened 2026-06-03
feat(rbac): OpenFGA authz, team-invoke + org-wide sharing for custom MCP tools

sriaradhyula:
## Summary

Custom MCP tool creation from the UI always failed with **403 Forbidden**, even for org admins and owner-team members. The RAG server gates `POST/PUT/DELETE /v1/mcp/custom-tools` on `require_role(Role.ADMIN)`, but `rbac.py` assigns **every human user `Role.READONLY`** with no elevation path — so no human can ever create/update/delete a custom MCP tool. The unified shareable-resource RBAC work added the `mcp_tool` OpenFGA type, ownership/sharing UI, and BFF tuple-writing, but left these server endpoints on the legacy coarse-admin gate.

Observed failure:

```
[rag.server.rbac] WARNING - Access denied for <user>: required admin, has readonly
POST /v1/mcp/custom-tools 403 Forbidden
```

This migrates the three custom-tool write endpoints to **OpenFGA-based authorization**, matching the documented model (`rag/README.md`: *"tool authorization comes from OpenFGA relationships"*) and the BFF/UI ownership semantics already shipped in this feature branch.

## Changes

- **`server/rbac.py`** — new fail-closed helpers:
  - `authorize_mcp_tool_create(user, owner_team_slug)` — allow if org admin **or** member of the requested owner team (`team:<slug>#can_use`).
  - `authorize_mcp_tool_manage(user, tool_id)` — allow if org admin **or** `mcp_tool:<tool_id>#can_manage` (owner, owner-team admin, org admin).
  - Both return **403** when unauthorized and **503** when the OpenFGA PDP is unavailable / not configured / the caller has no stable subject.
  - Coarse-`ADMIN` service principals (admin client-credentials tokens, `CAIPE_UNSAFE_RBAC_BYPASS`) are still permitted first, so existing automation is not regressed.
- **`server/restapi.py`** — `POST/PUT/DELETE /v1/mcp/custom-tools` now depend on `require_authenticated_user` and call the new authz helpers (replacing `require_role(Role.ADMIN)`).
- **`tests/test_mcp_tool_authz.py`** — 13 unit tests: coarse-admin bypass, org-admin, owner-team-member, `can_manage`, denial (403), and fail-closed paths (PDP error / not configured / no subject).

## Why this is the right layer

The RAG server is a real, directly-reachable enforcement point, so authorization must live here (not only in the BFF). On create the tool has no tuples yet, so we mirror the BFF first-set rule (org-admin or owner-team membership); on update/delete the per-resource `mcp_tool#can_manage` relation exists and is authoritative.

## Test plan

- [x] `uv run ruff check` on changed files — clean
- [x] `uv run pytest tests/test_mcp_tool_authz.py tests/test_openfga_team_rebac.py` — 24 passed
- [x] Rebuilt `rag-server` image, recreated container, verified startup + `/health` 200
- [ ] Create a custom MCP tool from the UI as an owner-team member → **201** and the tool appears in the list
- [ ] Non-member without org admin → **403**
- [ ] Update/delete by a non-owner non-admin → **403**

> Pre-existing (not introduced here): several server tests (`test_doc_acl.py`, `test_role_mapping.py`, `test_keycloak_oidc_auth.py`) fail on this branch because `UserContext` now requires a `groups` field they don't pass; `test_e2e.py` needs a live server.

### DIFF
--- a/ai_platform_engineering/knowledge_bases/rag/common/src/common/models/rag.py
+++ b/ai_platform_engineering/knowledge_bases/rag/common/src/common/models/rag.py
@@ -317,6 +317,10 @@ class MCPToolConfig(OwnedResourceMixin, BaseModel):
   description: str = Field(default="", description="Tool description shown to the LLM agent")
   parallel_searches: List[ParallelSearch] = Field(default_factory=lambda: [ParallelSearch(label="results")], description="One or more parallel sub-searches. Response is a dict keyed by label.")
   allow_runtime_filters: bool = Field(default=False, description="If True, expose a 'filters' parameter so the LLM can pass extra filters per-call.")
+  shared_with_org: bool = Field(
+    default=False,
+    description="If True, every organization member may call/use this tool (in addition to the owner and shared teams). The OpenFGA projection grants organization#member reader/user/caller.",
+  )
   enabled: bool = True
   created_at: int = Field(default=0, description="Unix timestamp of creation")
   updated_at: int = Field(default=0, description="Unix timestamp of last update")
--- a/ai_platform_engineering/knowledge_bases/rag/server/src/server/rbac.py
+++ b/ai_platform_engineering/knowledge_bases/rag/server/src/server/rbac.py
@@ -538,6 +538,97 @@ async def _openfga_check_org_admin(user_context: UserContext) -> bool:
   return await _openfga_check_object(user_context, "can_manage", "organization", _caipe_org_key())
 
 
+# ============================================================================
+# Custom MCP tool authorization (spec 2026-06-03-unified-shareable-resource-rbac)
+# ============================================================================
+#
+# Custom MCP tool management (POST/PUT/DELETE /v1/mcp/custom-tools) is a
+# group-owned, shareable resource. Authorization for human callers is resolved
+# through OpenFGA on the `mcp_tool` type — NOT the legacy coarse `require_role`
+# gate, which can never elevate a human above READONLY (see `rbac.py` role
+# assignment and `rag/README.md`: "tool authorization comes from OpenFGA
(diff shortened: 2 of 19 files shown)


sriaradhyula:
### Follow-up commits (same MCP-tool authz theme)

- **feat(rbac): org-wide and team-invoke sharing for custom MCP tools** (`2d33ae3f5`)
  - Fixes the latent bug where shared/owner team members got `reader` + `user` but **not** `caller`, so `can_call` denied them and KB-search tools couldn't actually be invoked. Member grants now write `caller`; the `mcp_tool_grants_backfill_v1` migration backfills existing rows.
  - Adds first-class **org-wide sharing**: new `shared_with_org` flag on `MCPToolConfig` (Python + TS), reconciler emits/revokes `organization#member` reader/user/caller, OpenFGA model + chart `authorization-model.json` allow `organization#member` as a subject for `mcp_tool`, BFF route threads + persists the flag, and a "Share with the whole organization" toggle in the tool editor. Makes the previously hand-applied `caipe_kb` org grant durable.
  - UI polish: MCP tools view shows canonical datasource **names** (id in tooltip) instead of raw `src_...` ids.
- **feat(rbac): grant org-admin to super-admins team members** (`2b149abf4`) — bootstraps `team:super-admins#admin -> admin -> organization:<key>`.
- **test(rag): make RAG server e2e tests opt-in via RAG_E2E** (`333c990cd`) — keeps the default `pytest` run green.

All new/affected UI tests pass (79 in the rbac + rag-route suites). `caipe_kb` itself is a runtime tool (no config file), so its `shared_with_org` flag becomes durable once it's re-saved with the toggle on after the next `caipe-ui-prod` rebuild.

PR #1708, opened 2026-06-03
feat(rbac): unify group-based access control across agents, RAG datasources, and MCP tools

kevkantes:
# Description

Makes the agent **owner-team + share-with-teams** access-control pattern canonical and reusable, then brings RAG datasources and custom MCP tools to full parity. Config remains the source of truth; OpenFGA is the derived projection. Implements spec `2026-06-03-unified-shareable-resource-rbac` (US1–US6).

**Shared module (US1).** A reconciler core (`buildShareableResourceTupleDiff` / `reconcileShareableResource` + the `buildTeamGrantTuples` primitive), a route helper (`handleShareableResourceWrite`), a Pydantic `OwnedResourceMixin`, and a `<TeamOwnershipFields>` React component. The agent and knowledge_base reconcilers are refactored into thin adapters over the core — existing suites pass unchanged.

**Creator/owner split (US2).** Adds an audit-only `creator: [user]` relation to `agent`, `knowledge_base`, `data_source`, and `mcp_tool` (referenced by **no** `can_*`). A shareable-type drift test guards that invariant and authored↔chart model parity.

**RAG datasource inheritance (US4).** `data_source` gains `parent_kb` and `can_read`/`can_ingest`/`can_manage` inherit from the knowledge_base (the model's first tuple-to-userset). **Retires the PR #1703 `mirrorKnowledgeBaseDiffToDataSource` mirror** — the sharing PUT and team KB-assignment route now write the single inheritance edge instead of duplicating per-team tuples. Fixes the prior "see-but-not-search" gap.

**RAG + MCP persistence and parity (US5/US6).** `OwnedResourceMixin` on `DataSourceInfo` and `MCPToolConfig`; the datasource sharing GET returns the real owner/creator from config; owner/share UI via `<TeamOwnershipFields>` in `KbSharingPanel` and the MCP `ToolFormDialog`. MCP custom tools now reconcile owner/shared/creator on POST/PUT, **clean up all `mcp_tool:<id>` grants on DELETE** (no orphan tuples), and enforce **`can_call` on `/v1/mcp/invoke`** (user + agent principals; built-ins exempt; org-admin bypass).

**Ownership transfer (US3).** Owner team is immutable on edit but transferable via the editor, guarded by `canTransferResourceOwnership` (owner-team admin **or** org admin), with a not-a-member confirmation; the old team's grants are revoked and the `creator` tuple is preserved.

[...]

PR #1727, opened 2026-06-04
feat(skills): share skills with teams via member-accessible teams picker

sriaradhyula:
## Summary

Lets generic org **members** author and **team-share** skills in the Skill Builder. Previously the coarse `withAuth` gate for `/api/skills/configs` (POST/PUT/DELETE) and `/api/catalog-api-keys` fell through `organizationRelationFor()` to the admin-only `can_manage` relation, so `skill#configure` / `skill#delete` returned `403 "You do not have permission to perform this action."` for non-admins — and there was no UI to pick teams to share with.

The backend sharing machinery (`grantSkillsToTeams`, `shared_with_teams`, `team:<slug>#member user skill:<id>` grants) already shipped in #1526; this wires up the **member gate** and the **teams picker** so the capability is actually reachable.

## Changes

- **`api-middleware.ts`** — map every `skill` capability (`view` / `invoke` / `configure` / `delete`) to the member-level `can_use` (`member or admin`), not `can_manage`. Per-skill mutation/deletion of an *existing* skill is still ownership-constrained per-resource inside the route handlers via `requireResourcePermission({ type: "skill", action })`; the org gate only asserts the Skill Builder exists for the caller at all.
- **`OverviewTab.tsx`** — add a member-accessible teams picker. `useShareableTeams()` loads the caller's shareable teams from `GET /api/dynamic-agents/teams` (the same endpoint the RAG KB / MCP / Dynamic-Agent editors use — members get their own teams, org admins get all) and renders `TeamMultiPicker` when `visibility = team`, so the save writes the `team:<slug>#member user skill:<id>` grants.
- **Tests** — member `can_use` gate (`api-middleware.test.ts`) and the teams picker (`OverviewTab.test.tsx`).
- **Docs** — RBAC architecture reference: "Skill authoring is a member self-service surface".

### Automated

- [ ] `npx jest src/lib/__tests__/api-middleware.test.ts src/components/skills/workspace/__tests__/OverviewTab.test.tsx`
- [ ] `tsc --noEmit` clean

### Manual test cases

| ID | Title | Preconditions | Steps | Expected result |
|----|--------|---------------|-------|-----------------|
| TC-1727-01 | Team picker visible for team visibility | Logged in | 1. Create/edit skill → visibility **Team** | **TeamMultiPicker** shown on Overview |
| TC-1727-02 | Team picker hidden for private | Skill private/non-team | 1. Open Overview | No team picker (or disabled) |
| TC-1727-03 | Member sees only own teams | Non-admin member on Team A only | 1. Open team picker | Team A listed; not all org teams |
| TC-1727-04 | Org admin sees broader list | Org admin | 1. Open team picker | More teams than a single-member user |
| TC-1727-05 | Save with teams | Team visibility | 1. Select Team A + Team B → Save | Success; reload shows same teams selected |
| TC-1727-06 | Shared team member sees skill | TC-1727-05; user on Team A | 1. Log in as Team A member → Skills list | Skill visible per catalog rules |

- [ ] TC-1727-01
- [ ] TC-1727-02
- [ ] TC-1727-03
- [ ] TC-1727-04
- [ ] TC-1727-05
- [ ] TC-1727-06

Assisted-by: Claude:claude-opus-4.8

### DIFF
--- a/charts/ai-platform-engineering/Chart.yaml
+++ b/charts/ai-platform-engineering/Chart.yaml
@@ -1,20 +1,20 @@
 apiVersion: v2
 # pyproject.toml version
-appVersion: 0.5.8-dev.1 # Do NOT modify. It will be updated automatically using the release workflow
+appVersion: 0.5.8-dev.2 # Do NOT modify. It will be updated automatically using the release workflow
 name: ai-platform-engineering
 description: Parent chart to deploy multiple agent subcharts as different platform agents
 sources:
   - https://github.com/cnoe-io/ai-platform-engineering/charts
 # Chart version for ai-platform-engineering parent chart
-version: 0.5.8-dev.1 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+version: 0.5.8-dev.2 # Do NOT bump this. It will be updated automatically using the PR or release workflow
 dependencies:
   # AI Platform Engineer Multi-Agent
   - name: supervisor-agent
     # Chart version for agent subchart
-    version: 0.5.8-dev.1 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.5.8-dev.2 # Do NOT bump this. It will be updated automatically using the PR or release workflow
   # Single agent chart used multiple times with different aliases
   - name: agent
-    version: 0.5.8-dev.1 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.5.8-dev.2 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     alias: agent-argocd
     tags:
       - agent-argocd
@@ -24,7 +24,7 @@ dependencies:
       - child: agentExports.data
         parent: global.enabledSubAgents.argocd
   - name: agent
-    version: 0.5.8-dev.1 # Do NOT bump this. It will be updated automatically using the PR or release workflow
+    version: 0.5.8-dev.2 # Do NOT bump this. It will be updated automatically using the PR or release workflow
     alias: agent-aws
     tags:
       - agent-aws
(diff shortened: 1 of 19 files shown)


kevkantes (changes requested):
You should be able to use the shared modules here: https://github.com/cnoe-io/ai-platform-engineering/pull/1726

This should standardize the RBAC flow between all resources

sriaradhyula:
> You should be able to use the shared modules here: #1726
> 
> This should standardize the RBAC flow between all resources

Done. Refactored to use shared RBAC modules

sriaradhyula:
Thanks @kevkantes for pointing at #1726 — agreed on standardizing RBAC where the resource model matches.

**Why this PR is closed:** After rebase onto `main`, branch `prebuild/feat/skills-team-share-picker` has no diff. The functional work landed elsewhere:

- **Member gate + team sharing (backend):** #1729 — skills create/update use `reconcileShareableResource` / shared shareable-resource helpers (same direction as #1726), plus catalog API key HMAC and related tests/docs.
- **Team picker (UI):** Skill Builder Overview already uses visibility + `TeamMultiPicker` fed by `GET /api/dynamic-agents/teams` (members see own teams; org admins see a broader list). That matches the **skill** model (user-owned, private/team/global visibility, grant teams via `shared_with_teams` / FGA member grants) — not the **owner-team** model from #1708.

**Why not `TeamOwnershipFields` here:** That component is for resources with a single **owner team**, transfer semantics, and `owner_team_slug` on the document. Skills do not have an owner team today; forcing that component would be a product/schema change (Mongo + FGA + transfer UX), not a straight reuse of the shared reconciler.

**#1726 takeaway for skills:** Backend is aligned with the shared shareable-resource path. UI parity with agents/KB/MCP is intentionally different until we decide skills should be owner-team resources.

**Optional follow-ups (not blocking this thread):** tracked in a follow-up issue (link will be added in a reply once created).

- Extract `useShareableTeams` → `ui/src/hooks/use-shareable-teams.ts` (OverviewTab, KbSharingPanel, DynamicAgentEditor, MCPToolsView, WorkflowCanvas, etc.)
- Or a slim “effective grants” callout on team visibility without full ownership fields

If product wants skills to look like agents (owner team, transfer, `TeamOwnershipFields`), that should be a separate spec/PR — not required to close this review thread.

(Supersedes the earlier short “Done” note — this is the full rationale for close vs #1726.)

sriaradhyula:
Follow-up tracking (optional polish + owner-team skills spec): #1755

PR #1749, opened 2026-06-04
feat(slack): add channel delete and editable agent swap in channel route editor

kevkantes:
# Description

Two related improvements to the Slack channel admin panel (`/admin?cat=integrations&tab=slack` → Configured channels).

**1. Editable Dynamic Agent in the route editor (agent swap)**

Previously the Dynamic Agent picker was disabled when editing an existing route (`Boolean(editingRoute)`), so changing which agent a channel routes to required deleting the route and recreating it — re-checking every setting (priority, listen modes, overthink, escalation) by hand.

The picker is now editable during edit. The existing PUT reconcile already handles the swap cleanly: it deletes the old agent's OpenFGA `can_use` tuple, revokes the old Mongo route doc, and upserts the new agent's doc with the kept settings (settings travel in the request body, so they transfer with no orphaned tuple/metadata). Agents that already have their own route in the channel are excluded from the picker to preserve the one-route-per-agent invariant (Mongo upserts on `agent_id`).

**2. Delete channel (hard delete)**

There was no way to offboard a channel after onboarding. Added a Danger zone "Delete channel" action with a confirmation dialog, backed by a new `DELETE` handler that purges all channel state:

- OpenFGA tuples in both directions — channel→resource grants (channel as subject) and team→channel visibility (channel as object), enumerated in a single read pass.
- Mongo: `slack_channel_agent_routes`, `slack_channel_grants`, and `channel_team_mappings`.

OpenFGA is cleared before Mongo so a failure aborts before any DB mutation, keeping the delete safely re-runnable. A new `deleteExactOpenFgaTuples()` primitive deletes the exact enumerated tuple keys, bypassing the per-tuple `/check` filter in `writeOpenFgaTuples` (that filter cannot resolve a userset such as `team:<slug>#member` as the `user`, so the team-visibility deletes would otherwise be silently dropped and orphaned).

### Changed files
- `ui/src/components/admin/rebac/slack/SlackConfiguredChannelDetail.tsx` — editable agent picker + Danger zone delete UI/confirm dialog
- `ui/src/app/api/admin/slack/channels/[workspaceId]/[channelId]/route.ts` — new channel-root `DELETE` handler
- `ui/src/lib/rbac/openfga.ts` — `deleteExactOpenFgaTuples()`
- `ui/src/lib/rbac/slack-channel-route-store.ts` — `deleteSlackChannelAgentRoutes()`
- `ui/src/lib/rbac/slack-channel-grant-store.ts` — `deleteSlackChannelGrants()`

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

N/A — no chart changes.

## Checklist


sriaradhyula (dismissed):

ui/src/app/api/admin/slack/channels/[workspaceId]/[channelId]/route.ts
```diff
@@ -0,0 +1,96 @@
...
+  const matches: OpenFgaTupleKey[] = [];
+  let continuationToken: string | undefined;
+  do {
+    const result = await readOpenFgaTuples({
+      pageSize: 100,
+      ...(continuationToken ? { continuationToken } : {}),
+    });
+    for (const tuple of result.tuples) {
+      if (tuple.key.user === channelRef || tuple.key.object === channelRef) {
+        matches.push(tuple.key);
```
Comment (sriaradhyula) [inline ui/src/app/api/admin/slack/channels/[workspaceId]/[channelId]/route.ts]: **[Major] Full-store OpenFGA scan on every channel delete**

`readOpenFgaTuples()` is called without a filter tuple, so it reads **all** tuples in the store and filters client-side — O(total_tuples). At scale this can hit read timeouts.

OpenFGA's `Read` API accepts a partial `tuple_key` for server-side filtering. Consider issuing two targeted reads instead:
- `{ tuple: { object: channelRef } }` — tuples where channel is the object
- `{ tuple: { user: channelRef } }` — tuples where channel is the user

Then union the results. This reduces the scan to O(channel_tuples).

ui/src/lib/rbac/openfga.ts
```diff
@@ -508,6 +508,46 @@ export async function writeOpenFgaTuples(diff: TeamResourceTupleDiff): Promise<O
...
+  }
+  const storeId = await getOpenFgaStoreId();
+  let total = 0;
+  for (const chunk of chunkOpenFgaDiff({ writes: [], deletes: unique })) {
+    await postOpenFgaWriteChunk(baseUrl, storeId, chunk);
+    total += chunk.deletes.length;
+  }
+  return { enabled: true, writes: 0, deletes: total };
+}
+
```
sriaradhyula, line 593: **[Major] No compensation on partial chunk failure**

`writeOpenFgaTuples` has `compensateAppliedChunks` to roll back already-applied chunks if a later chunk fails. This function has no such compensation — if chunk 2 of 3 fails, chunk 1 is already deleted with no rollback.

The delete is re-runnable (re-reads surviving tuples on retry), so this isn't a hard correctness bug, but it's inconsistent with the existing pattern and leaves the store in a partial state until retry. Consider adding the same compensation loop for consistency.

ui/src/components/admin/rebac/slack/SlackConfiguredChannelDetail.tsx
```diff
@@ -454,6 +464,23 @@ export function SlackConfiguredChannelDetail({
...
+      });
+      if (!res.ok) throw new Error(await res.text());
+      setChannelDeleteOpen(false);
+      toast(`Removed ${selected.item_name || selected.item_id} from CAIPE.`, "success");
+      await onRefresh();
+    } catch (err) {
+      toast(err instanceof Error ? err.message : "Failed to delete Slack channel", "error");
+    } finally {
+      setLoading(false);
+    }
```
sriaradhyula, line 486: **[Minor] Detail panel may remain visible after channel deletion**

`onRefresh()` reloads the channel list but doesn't clear the selected channel. If the parent keeps the deleted channel selected until the list fully reloads, the user briefly sees a detail panel for a non-existent channel. Consider calling a `onChannelDeleted()` / `onDeselect()` callback (or resetting selection inside `onRefresh`) to navigate away immediately on successful delete.

ui/src/components/admin/rebac/slack/SlackConfiguredChannelDetail.tsx
```diff
@@ -515,6 +552,28 @@ export function SlackConfiguredChannelDetail({
...
+            <p>The Slack bot stops responding here once its route cache expires. Re-onboard the channel from the Onboard tab to set it up again.</p>
+          </div>
+          <DialogFooter>
+            <Button type="button" variant="outline" onClick={() => setChannelDeleteOpen(false)} disabled={loading}>Cancel</Button>
+            <Button type="button" variant="destructive" onClick={() => void deleteChannelConfirmed()} disabled={loading}>{loading ? "Deleting..." : "Delete channel"}</Button>
+          </DialogFooter>
+        </DialogContent>
+      </Dialog>
     </div>
   );
```
sriaradhyula, line 583: **[Minor] Shared `loading` state could cross-contaminate two dialogs**

Both the route-delete dialog and the channel-delete dialog read from the same `loading` / `setLoading` state. If both are open simultaneously (unlikely but possible in a race), the loading state from one operation could unblock the other dialog's confirm button. Consider a dedicated `isDeletingChannel` local state for the channel-delete flow.

ui/src/app/api/admin/slack/channels/[workspaceId]/[channelId]/route.ts
Comment (kevkantes) [reply ui/src/app/api/admin/slack/channels/[workspaceId]/[channelId]/route.ts]: Fixed in c87411309. Replaced the full-store paginated scan with two server-side filtered reads — `readOpenFgaTuples({ tuple: { user: channelRef } })` and `{ tuple: { object: channelRef } }` — run in parallel, then unioned and deduped. This matches the existing filtered-read pattern (e.g. `webex-space-openfga.ts`, `rebac-graph.ts`, `openfga-owned-resources.ts`) and scopes the scan to O(channel_tuples).

ui/src/lib/rbac/openfga.ts
kevkantes: Fixed in c87411309. Extracted the chunk-apply-with-compensation loop out of `writeOpenFgaTuples` into a shared private `applyDiffWithCompensation(baseUrl, storeId, diff)`; both `writeOpenFgaTuples` (post-filter) and `deleteExactOpenFgaTuples` now route through it, so the delete path gets the same partial-failure rollback. No duplicated logic.

ui/src/components/admin/rebac/slack/SlackConfiguredChannelDetail.tsx
kevkantes: Fixed in c87411309. Added an `onDeselect` callback threaded through the adapter context (`ConnectorAdminPanel` passes `() => setSelectedKey("")`); the channel-delete handler now calls it on success before `onRefresh()`, so the detail panel closes immediately instead of lingering on a now-deleted row. `onDeselect` is required on the context, but Webex doesn't use `configuredDetailExtra`, so only the Slack adapter is affected.

ui/src/components/admin/rebac/slack/SlackConfiguredChannelDetail.tsx
kevkantes: Left as-is, intentionally. The two confirmation dialogs are Radix `Dialog`s, which are modal — only one can be open at a time, so they can't actually race over the shared `loading` flag. Sharing `loading`/`setLoading` here is also consistent with the existing `deleteRouteConfirmed` flow in this same component; introducing a dedicated `isDeletingChannel` state just for this path would make it the odd one out. Happy to revisit if you'd still prefer the dedicated state.

sriaradhyula (approved):

PR #1751, opened 2026-06-05
feat(workflows): RBAC visibility, run access, team slugs, and unsaved UX

sriaradhyula:
## Summary

- **Workflow RBAC**: Global/team/private visibility with OpenFGA `task#read` / `task#use`; list and detail use the same view rules; runs API uses `withAuth` and returns 403 when access is denied.
- **Team sharing**: Normalize `shared_with_teams` to team slugs (not Mongo `_id`) in API, seed, and UI (`TeamMultiPicker`).
- **Non-admin access**: Map `GET/POST /api/workflow-configs` to `dynamic_agent#view` / `manage` in BFF middleware so non-admins can list/run shared workflows.
- **Execution auth**: Forward user JWT to dynamic agents for workflow runs; add `workflow_execution_authz` checks on the DA side.
- **Editor UX**: Show an **Unsaved** toolbar badge and in-app discard dialog when leaving the editor (Exit, nav links, sidebar workflow/run switches).
- **Agent access on save**: Before saving team/global workflows, check OpenFGA for agent access gaps per shared team; show `WorkflowAgentAccessModal` to bulk-grant `team#member → user → agent` tuples via admin OpenFGA API, then persist.

## Test plan

- [x] `pytest tests/test_workflow_execution_authz.py` (dynamic agents)
- [x] `npm test` — `workflow-config-rebac`, `workflow-runs-rbac`, `task-workflow-rbac`, `api-middleware`, `unsaved-changes-store`
- [ ] As non-admin (e.g. CU Generic): see global/team workflows in sidebar; runs list matches openable runs; denied run returns 403
- [ ] Edit workflow name → **Unsaved** badge; Exit / switch workflow → discard dialog
- [ ] Team workflow: share via team picker (slugs); member can run/list
- [ ] Save team workflow with agent teams cannot use → access modal lists gaps; **Grant access and save** grants tuples and saves

## Notes

- Does not include unrelated admin migration-status / schema-bootstrap changes left on local `main`.
- Global workflows: modal detects `(all users)` agent gaps but does not auto-grant (team slugs only in grant loop).

### DIFF
--- a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/auth/workflow_execution_authz.py
+++ b/ai_platform_engineering/dynamic_agents/src/dynamic_agents/auth/workflow_execution_authz.py
@@ -1,0 +1,101 @@
+"""Authorize agent use when executing a workflow step (delegated from workflow config)."""
+
+from __future__ import annotations
+
+import logging
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from dynamic_agents.models import UserContext
+    from dynamic_agents.services.mongo import MongoDBService
+
+logger = logging.getLogger(__name__)
+
+
+def _normalize_email(value: str | None) -> str:
+    return (value or "").strip().lower()
+
+
+def _normalize_team_slug(value: str) -> str:
+    return value.strip().lower()
+
+
+def _agent_in_workflow_steps(workflow_doc: dict[str, Any], agent_id: str) -> bool:
+    for step in workflow_doc.get("steps") or []:
+        if not isinstance(step, dict):
+            continue
+        if step.get("type") == "step" and step.get("agent_id") == agent_id:
+            return True
+    return False
+
+
+def _user_team_slugs(mongo: MongoDBService, user_email: str) -> set[str]:
+    db = getattr(mongo, "_db", None)
+    if db is None:
+        return set()
+    try:
+        rows = db["teams"].find(
+            {"members.user_id": user_email},
+            {"slug": 1},
+        )
+        return {
+            _normalize_team_slug(row["slug"])
+            for row in rows
+            if isinstance(row.get("slug"), str) and row["slug"].strip()
+        }
+    except Exception as exc:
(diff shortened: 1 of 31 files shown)


ui/src/lib/server/workflow-da-auth.ts
```diff
@@ -0,0 +1,34 @@
...
+  } else if (accessToken) {
+    headers.Authorization = `Bearer ${accessToken}`;
+  }
+
+  headers["X-User-Context"] = Buffer.from(
+    JSON.stringify({
+      email: user.email,
+      name: user.name ?? null,
+    }),
+  ).toString("base64");
```
subbaksh, line 31: Any reason why we still do this? I thought we migrated to using jwt tokens and letting DA validate it, instead of sending the user/email in the header and DA trusting it?

ai_platform_engineering/dynamic_agents/src/dynamic_agents/auth/workflow_execution_authz.py
```diff
@@ -0,0 +1,101 @@
+"""Authorize agent use when executing a workflow step (delegated from workflow config)."""
+
```
subbaksh, line 2: Workflows are purely a UI server concept. The UI invokes agents one step at a time.  DA usually has no clue whether its a workflow or a user chatting.

For invocation - the DA server just has tool (with workflow_id param) that it calls to invoke a workflow. The RBAC for invocation should be in UI server no?

ui/src/lib/server/workflow-da-auth.ts
sriaradhyula: @subbaksh 

X-User-Context is still required because chat/workflow handlers in Dynamic Agents depend on get_user_context(), which decodes that header today for getting email for adding it to the context. 

I added a follow-up task to clean this up - https://github.com/cnoe-io/ai-platform-engineering/issues/1753

sriaradhyula:
Superseded by #1772, which re-implements workflow RBAC on the Centralized Authorization Service (CAS). Closing in favor of the CAS-based approach.

PR #1767, opened 2026-06-05
fix(workflows): require agent#use AND workflow access for step execution

sriaradhyula:
Companion fix for #1751.

## Problem

The fallback in `require_agent_use_permission` allowed workflow membership to **escalate privilege**: a user without `dynamic_agent#use` could gain access to an agent by being a member of a team that owns a workflow containing that agent as a step.

```
# Before — OR logic: workflow bypasses direct agent check
if not allowed and can_use_agent_via_workflow(...):
    allowed = True
```

## Fix

Workflow context is **additional validation**, not an escalation path. When `workflow_config_id` is present both conditions must be true:

1. **`dynamic_agent#use`** — user has direct OpenFGA permission on the agent
2. **`can_use_agent_via_workflow`** — agent is a step in the referenced workflow AND the user can run that workflow (visibility check)

```python
# After — AND logic: workflow tightens, not loosens, the check
if workflow_config_id and mongo and user:
    if allowed and not can_use_agent_via_workflow(...):
        allowed = False
        deny_reason = "DENY_WORKFLOW_ACCESS"
```

A distinct audit reason code `DENY_WORKFLOW_ACCESS` makes it easy to distinguish in audit logs from a plain missing-capability denial.

## Tests

Added 3 new unit tests covering:
- Agent not in workflow steps → denied
- Missing/unknown workflow_config_id → denied
- Private workflow owner vs non-owner

All 5 tests pass.

### DIFF
--- a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/auth/openfga_authz.py
+++ b/ai_platform_engineering/dynamic_agents/src/dynamic_agents/auth/openfga_authz.py
@@ -421,14 +421,15 @@ async def require_agent_use_permission(
             },
         )
 
-    if (
-        not allowed
-        and workflow_config_id
-        and mongo is not None
-        and user is not None
-        and can_use_agent_via_workflow(agent_id, workflow_config_id, user, mongo)
-    ):
-        allowed = True
+    # When a workflow_config_id is present the call is a workflow step execution.
+    # Require both: the user must have direct dynamic_agent#use AND the agent must
+    # be a step in a workflow the user can run. Workflow context is additional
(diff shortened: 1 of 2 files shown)


sriaradhyula:
Closing — the OR delegation is intentional. Addressing the access gap at the UI layer instead: warn on share when target users lack dynamic_agent#use on workflow step agents (separate PR).

sriaradhyula:
Closing this stale companion PR. The workflow-agent OR delegation was intentionally kept, and the access gap was handled through the later workflow/CAS and UI guardrail work instead of changing this runtime check to AND semantics.

PR #1784, opened 2026-06-08
feat(slack): Run As identity (user / service account) + unlinked-user fallback for Slack routing

cisco-erilutz:
## Summary

Builds on the Service Accounts feature (PR #1780) to add a per-route **"Run as"** identity for Slack routing and an **unlinked-user fallback**.

- **Per-route "Run as"** — each Slack channel→agent route runs as the **User** (default) or a **Service Account**. Effective permissions are the intersection of the chosen identity's permissions and the agent's permissions.
- **Platform "unlinked" service account** — a single bootstrapped SA (owned by the `super-admins` team) used as the fallback identity for callers with no linked enterprise identity. Platform admins manage its base scopes via an admin-gated modal.
- **IdP-broker-gated unlinked fallback** — when a federated IdP broker is configured, a caller with no linked federated identity (e.g. a just-in-time shell user created from Slack) runs as the unlinked SA until they link their enterprise identity. With no broker configured, JIT users run as themselves (no downgrade).
- **SA runtime token** is minted via Keycloak token-exchange (impersonation of the SA's service-account user) — no stored secrets; reuses the existing OBO plumbing.

> **Naming:** "anonymous" was renamed to **"unlinked"** throughout. These users *are* authenticated in Slack — they simply haven't linked their enterprise/IdP identity yet — so "anonymous" was misleading in security/IAM terms (per review discussion).

## How it works

`User` routes run as the caller when they have a verified federated identity, else the unlinked SA. `Service Account` routes always run as the chosen SA (works for both human and bot messages). Identity is namespaced consistently as `service_account:<sub>` across the BFF, dynamic-agents, and the AgentGateway authz bridge via the canonical `preferred_username` rule.

## Identity correctness & security

- **Cross-team SA guard** — a route can only be set to "Run as" a service account owned by the channel's owning team; the server validates this on write (403 otherwise), not just the UI.
- Apply the route "Run as" identity in `handle_mention` (not just ambient messages), so an `@mention` on a service-account route binds the SA token before the conversation `can_use` check.
- Grant the SA `writer` on conversations it creates — `owner_id` stays the human's email for auditability — plus a startup reconcile that backfills missing grants.
- Thread `isServiceAccount` through `requireConversationWriteAccess` so SA callers graph as `service_account:<sub>` on `conversation#write`.
- Broker-availability check uses a last-known-good value on Keycloak error (no silent privilege change during a transient outage).
- The slack-bot no longer talks to MongoDB directly — it resolves the unlinked SA via a BFF endpoint (`GET /api/integrations/unlinked-service-account`).

## Protected service accounts

A "protected" SA (the unlinked SA) cannot be revoked or moved to another team. The UI greys out the control with a shield icon, and the revoke route returns 403 as defense in depth.

## UX

- Agent-access / channel-grant denials are **ephemeral**, distinguish **"you"** vs **"the \<service account\>"**, name the **owning team** to contact, and post at the channel root for top-level messages (in-thread only for genuine replies).
- The **"Run as"** selector sits at the top of the route editor with inline permission copy; **Priority** is the top-level field. The service-account picker is a searchable dropdown scoped to the owning team, with explicit loading/empty/error states.
- Slack-bot config import defaults "Run as" to **User**, so importing legacy YAML never produces service-account routes.

## Testing

- **UI/BFF (Jest):** "Run as" types/store/route/UI, unlinked-SA bootstrap + resolver, the admin modal, the org-admin scope-management bypass, the cross-team SA guard, the conversation writer grant + reconcile, the `conversation#write` SA namespacing, the protected-SA revoke guard, and the searchable SA picker.
- **Slack bot (pytest):** SA token minting, the unlinked-SA resolver (via BFF), the "Run as" dispatch decision table, the federation/broker gate, and the broker-gated fallback.
- `tsc` clean; Ruff clean.

## Notes

- This branch is stacked on `2026-06-05-service-accounts-impl` (PR #1780) and targets that branch as its base. It should merge after #1780.
- The unlinked-SA "grantable" scope set currently reflects the acting admin's own grants; a full-platform grantable variant for org-admins is tracked as a follow-up.

### DIFF
--- a/ai_platform_engineering/integrations/slack_bot/app.py
+++ b/ai_platform_engineering/integrations/slack_bot/app.py
@@ -7,6 +7,7 @@ A Slack bot that communicates with dynamic agents via AG-UI protocol.
 Supports @mention queries, Q&A mode, AI alert processing, HITL forms, and feedback.
 """
 
+import asyncio
 import os
 import re
 import sys
@@ -51,6 +52,8 @@ from utils.slack_runtime_policy import (  # noqa: E402
     should_post_route_miss_notice,
     should_process_slack_payload,
 )
+from utils.dispatch_identity import apply_execution_identity  # noqa: E402
+from utils.unlinked_fallback import apply_unlinked_fallback  # noqa: E402
 from utils.slack_admin_api import start_slack_admin_api_server  # noqa: E402
 
 app = App(token=os.environ.get("SLACK_INTEGRATION_BOT_TOKEN", os.environ.get("SLACK_BOT_TOKEN", "")))
@@ -67,7 +70,6 @@ def _msg_link(channel_id: str, ts: str) -> str:
 RBAC_ENABLED = os.environ.get("SLACK_RBAC_ENABLED", "false").lower() == "true"
 
 if RBAC_ENABLED:
-    import asyncio
     from utils.identity_linker import (
         resolve_slack_user,
         generate_linking_url,
@@ -81,14 +83,21 @@ if RBAC_ENABLED:
         is_dm_channel,
     )
     from utils.slack_channel_auto_assign import get_slack_channel_auto_assigner
-    from utils.obo_exchange import impersonate_user, OboExchangeError
+    from utils.obo_exchange import impersonate_user, impersonate_service_account, OboExchangeError
     from utils.slack_rebac import get_slack_channel_rebac_evaluator
     from utils.user_messages import TEAM_SESSION_UNAVAILABLE_MESSAGE
+    from utils.service_account_resolver import get_unlinked_service_account_sub
+    from utils.keycloak_admin import user_is_federated, realm_has_enabled_idp_broker
 
     async def _rbac_enrich_context(body, slack_user_id, context, *, require_mapping: bool = True):
         """Resolve identity and enrich Bolt context.
 
         Returns 'unlinked', ('deny', message), or 'ok'.
+        - 'unlinked': no Keycloak user could be resolved (JIT off/failed/no email),
+          OR user resolved but has no live IdP link AND the realm has an enabled
+          broker → route as the unlinked SA.
+        - ('deny', message): channel has no team mapping (hard reject).
+        - 'ok': fully linked user; OBO token minted and stored in context.
         Stores team/workspace context for downstream OpenFGA channel checks.
         Channel→agent routing is now relationship-based: the selected Slack
         agent is authorized later against the channel's ReBAC grants.
@@ -184,6 +193,19 @@ if RBAC_ENABLED:
                     channel_id, team_resolution.team_name, team_resolution.team_slug, keycloak_user_id,
                 )
 
+        # Unlinked-fallback gate (anonymous-and-obo-routing):
+        # A JIT-from-Slack user (empty federatedIdentities) should run as the
+        # unlinked SA when the realm has an enabled IdP broker — broker
+        # presence means "real users authenticate through SSO; JIT shells are
+        # unverified placeholders until they link".  When there is NO broker,
+        # JIT-via-Slack IS the legitimate user base and they run as themselves.
+        if await realm_has_enabled_idp_broker() and not await user_is_federated(keycloak_user_id):
+            logger.info(
+                "User %s not IdP-linked (broker active) — routing as unlinked",
+                keycloak_user_id,
+            )
+            return "unlinked"
+
         try:
             obo = await impersonate_user(keycloak_user_id)
             context["obo_token"] = obo.access_token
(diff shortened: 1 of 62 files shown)


ai_platform_engineering/integrations/slack_bot/utils/config_models.py
```diff
@@ -67,11 +67,43 @@ def _require_listen_when_enabled(self):
     return self
 
 
+SlackRouteExecutionMode = Literal["obo_user", "service_account"]
```
cisco-erilutz: lets do `execution mode` -> Run As so that it matches the UI

cisco-erilutz: that will probably need to get changed in a few places

ai_platform_engineering/integrations/slack_bot/utils/dispatch_identity.py
```diff
@@ -0,0 +1,174 @@
...
+                "Please try again shortly or contact your admin."
+            ),
+        )
+        return False
+    finally:
+        if sa_loop is not None:
+            sa_loop.close()
+
+
+def _send_error_notice(
```
cisco-erilutz, line 174: this seems like it shouldn't be in this file, and be in a more appropriate file instead

ai_platform_engineering/integrations/slack_bot/utils/service_account_resolver.py
```diff
@@ -0,0 +1,130 @@
+# Copyright 2025 CNOE Contributors
+# SPDX-License-Identifier: Apache-2.0
+"""Resolve the platform anonymous service account from MongoDB.
```
cisco-erilutz: what do we need to resolve the service account for here? We DONT want the slack bot directly talking to the backend DB I think...

cisco-erilutz: is this even used anywhere?

ai_platform_engineering/integrations/slack_bot/utils/slack_agent_routes.py
```diff
@@ -471,6 +472,24 @@ def _route_to_agent_binding(route: dict[str, Any]) -> AgentBinding | None:
         return None
 
     try:
+        # Build execution_identity from the Mongo route doc when present.
+        # Omitted / None / unexpected values all fall back to the default
+        # ExecutionIdentity (mode="obo_user") — preserving backward compat.
+        exec_id_raw = route.get("execution_identity")
```
cisco-erilutz, line 478: al;so rebrand execution identity and obo -> Run as user vs service account

ai_platform_engineering/integrations/slack_bot/utils/config_models.py
cisco-erilutz: Renamed to `SlackRouteRunAsMode` / "Run as" across the Python slack-bot (config_models, dispatch_identity, slack_agent_routes, slack_admin_api, app). Kept the persisted `execution_identity` field + `obo_user` wire value stable for backward-compat. (A backward-compat `SlackRouteExecutionMode` alias remains so nothing breaks.)

ai_platform_engineering/integrations/slack_bot/utils/dispatch_identity.py
cisco-erilutz: Moved the implementation to `utils/user_messages.py` (`send_error_notice`). `dispatch_identity.py` now just imports + delegates via a thin wrapper, so the messaging logic no longer lives here.

ai_platform_engineering/integrations/slack_bot/utils/service_account_resolver.py
cisco-erilutz: Removed the direct MongoDB access. The bot now resolves the unlinked SA via a new BFF endpoint `GET /api/integrations/unlinked-service-account` (authenticated with the bot's service-account token), matching the existing BFF-call pattern. No pymongo in the bot anymore.

ai_platform_engineering/integrations/slack_bot/utils/slack_agent_routes.py
cisco-erilutz: Rebranded execution identity / obo → "Run as: User vs Service Account" here and across the slack-bot utils (symbols, logs, comments). Wire value `obo_user` kept stable.

ai_platform_engineering/integrations/slack_bot/app.py
```diff
@@ -1128,9 +1249,39 @@ def handle_mention(event, say, client, context=None):
...
+          context=context,
+          event=event,
+          client=client,
+          say=say,
+          is_bot=False,
+          impersonate_fn=impersonate_service_account,
+        )
+        if not should_proceed:
+          return
+      except AttributeError:
```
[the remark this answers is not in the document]
sriaradhyula: done

sriaradhyula (approved):

PR #1793, opened 2026-06-09
add opt-out checkbox to suppress feedback regeneration

jasonc3-sys:
Add an optional "Just record my feedback — don't regenerate a response" checkbox to the feedback modal. When ticked, the feedback is still recorded to the feedback API, but the bot confirms and returns instead of calling the AI. Unticked preserves the existing record-and-regenerate behavior. Scope is limited to the two modal paths (Wrong answer / Other). The non-modal paths (More detail, Briefer, Retry) intentionally keep regenerating, since those actions explicitly request a new answer. The checkbox-parsing logic lives in a small pure helper (suppress_regeneration_selected) in utils/scoring.py so it is unit-testable without importing the Bolt app module.

# Description
When a user gives negative feedback via the feedback modal (Wrong answer /
Other), the bot records the feedback **and** automatically regenerates a
response. There was previously no way to flag a bad answer without triggering
a regeneration.

This adds an optional **"Just record my feedback — don't regenerate a
response"** checkbox to the feedback modal:

- **Checked** → feedback is still recorded to the feedback API, but the bot
  confirms and returns instead of calling the AI.
- **Unchecked** → existing record-and-regenerate behavior is preserved.

Scope is limited to the two modal paths (**Wrong answer** / **Other**). The
non-modal paths (More detail, Briefer, Retry) intentionally keep regenerating,
since those actions explicitly request a new answer.

The checkbox-parsing logic lives in a small pure helper
(`suppress_regeneration_selected`) in `utils/scoring.py` so it is unit-testable
without importing the Bolt app module.

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [ ] Existing issues have been referenced (where applicable)   ← leave unticked (no public upstream issue; SDPL is internal)
- [ ] Functionality is documented   ← see note below

## Test plan

- [x] `pytest tests/test_feedback_modal_suppress.py` — 5 cases pass
- [x] `pytest tests/test_langfuse_feedback.py` — existing feedback tests still pass (7 passed total)
- [x] `ruff check` — new code is clean (pre-existing E402s on unrelated imports unchanged)
- [ ] Manual end-to-end in a live Slack workspace — not yet run


ai_platform_engineering/integrations/slack_bot/tests/test_feedback_modal_suppress.py
```diff
@@ -0,0 +1,51 @@
+# Copyright 2025 CNOE Contributors
+# SPDX-License-Identifier: Apache-2.0
+"""
+Tests for the SDPL-1872 "just record my feedback" checkbox in the feedback modal.
```
kevkantes: This is a public repo, don't leak internal jira ticket IDs

sriaradhyula:
@jasonc3-sys  - Can you please use https://www.conventionalcommits.org/en/v1.0.0/#summary in your commits?

ai_platform_engineering/integrations/slack_bot/tests/test_feedback_modal_suppress.py
```diff
@@ -0,0 +1,51 @@
...
+  )
+  return {
+    "correction_input": {"correction_text": {"type": "plain_text_input", "value": "It was wrong"}},
+    "suppress_regen_input": {
+      "suppress_regen": {"type": "checkboxes", "selected_options": options},
+    },
+  }
+
+
+class TestSuppressRegenerationSelected:
```
kevkantes: Lot of AI slop tests here, do we actually need these? Can we make tests that are just testing the new requirements?

ai_platform_engineering/integrations/slack_bot/app.py
```diff
@@ -2288,6 +2229,22 @@ def _open_feedback_modal(ack, body, client, feedback_type):
...
+          {
+            "type": "input",
+            "block_id": "suppress_regen_input",
+            "optional": True,
+            "element": {
+              "type": "checkboxes",
+              "action_id": "suppress_regen",
+              "options": [
+                {
+                  "text": {"type": "plain_text", "text": "Just record my feedback — don't regenerate a response."},
```
kevkantes: By default it should not regenerate a response. The checkbox should be like "Attempt to regenerate a response based on feedback?"

kevkantes (changes requested):
We should also make the non-modal flows (more detail, etc) also not automatically start a new response

ai_platform_engineering/integrations/slack_bot/tests/test_feedback_modal_regen.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_missing_block_does_not_regenerate():
+  """No regen block present (e.g. malformed payload) defaults to no regen."""
+  assert regenerate_requested({"correction_input": {}}) is False
+
+
+def test_non_dict_input_does_not_regenerate():
+  """Defensive: non-dict input must not raise."""
+  assert regenerate_requested(None) is False
```
kevkantes: Some more tests here that are just testing basic python if checks or type checks. These are not useful tests, we should try to limit these "AI slop" tests

kevkantes: When writing the tests, think what is the requirement here? If you only need 1 test for that, then that's perfectly ok. Tests should be functional chunks of the feature/bug fix.

kevkantes (approved):

PR #1810, opened 2026-06-10
fix(rbac): write org-admin connector tuple for super-admins team on bootstrap

sriaradhyula:
## Summary

- `ensureSuperAdminsTeam()` was writing `user:<sub> admin team:super-admins` for each bootstrap member but never writing the connector tuple `team:super-admins#admin admin organization:<key>`
- Without that tuple, team membership was disconnected from org authorization: `admin_ui#view` checks (which resolve to `can_audit organization:<key>`) failed with `DENY_NO_CAPABILITY` for anyone added to the team via the admin UI
- This fix writes `team:super-admins#admin admin organization:<key>` idempotently on every `ensureSuperAdminsTeam()` call (both create and update paths)

**After this fix:** adding a user to the super-admins team via the admin UI is sufficient — no need to add each person to `BOOTSTRAP_ADMIN_EMAILS`. The bootstrap env var remains useful only as break-glass access before the team is first seeded.

## Test plan

- [ ] Fresh deployment: bootstrap admin logs in, creates a team member via admin UI — member gets `admin_ui` access without restart
- [ ] Existing deployment: restart triggers `ensureSuperAdminsTeam()`, connector tuple is written, existing team members can log in without 403
- [ ] Idempotency: multiple restarts do not error (OpenFGA write of existing tuple is a no-op)


sriaradhyula:
Replaced by clean branch — see new PR

PR #1823, opened 2026-06-11
fix(setup): production-ready setup-caipe.sh — prereqs, domain/TLS, Keycloak, dynamic agents

sriaradhyula:
## Summary

End-to-end hardening of \`setup-caipe.sh\` to deploy CAIPE reliably on bare Kubernetes (kind/k3s) and make it production-reachable over HTTPS with a real domain.

### Prerequisites & installation
- Auto-install missing prereqs (kind, helm, kubectl, jq) with user consent gate for sudo
- Redirect kind stdin to prevent script truncation when run via \`curl | bash\`
- Auto-restore kubeconfig when kind context is missing

### Wizard & config persistence
- Save/restore wizard config via \`~/.config/caipe/config.yaml\`; \`--load-config=FILE\` flag
- Save/restore \`enable_ingress\` and \`selected_agents\` in config (fixes ingress installing when MetalLB disabled; fixes agent selection not persisting across re-runs)
- Prompt to confirm \`deployment_mode\` even when pre-loaded from saved config
- Persist and restore \`enable_metallb\` in saved config; force \`ENABLE_INGRESS=false\` when \`enable_metallb=false\`

### Cluster & Helm
- Add \`--force-conflicts\` to helm upgrade to resolve SSA field manager conflicts
- Delete stale helm hook Jobs before upgrade to prevent "already exists" errors
- Robust postgres StatefulSet discovery via name-prefix match + label selector

### Domain, TLS & ingress
- Export \`KUBECONFIG\` to align k3s kubectl with helm
- Fix \`/etc/hosts\` so \`CAIPE_DOMAIN\` resolves to MetalLB IP for local health checks
- Fix UI health check to use \`https://CAIPE_DOMAIN\` when ingress is enabled

### Keycloak
- Fix UUID extraction: replace broken \`sed '.*\"id\".*' | head -1\` with \`jq -r '.[0].id // empty'\` in three locations
- Add \`update_keycloak_client_urls()\`: idempotent Keycloak Admin REST API sync of redirectUris/webOrigins/rootUrl to current domain on every run
- Escape comma in \`OIDC_GROUP_CLAIM\` helm \`--set\` value

### Web-ingestor
- Pre-load \`RAG_INGESTOR_SECRET_READY\` from existing cluster secret on upgrade runs

### Dynamic agents — LLM model seeding
- Fix seed key path: \`caipe-ui.appConfig.models\` → ConfigMap → \`APP_CONFIG_PATH\` → \`applySeedConfig()\`
- Seed provider-correct model IDs: Bedrock, Azure, OpenAI, Ollama, Anthropic

### Dynamic agents — OIDC auth
- Fix OIDC issuer fallback: derive \`OIDC_ISSUER=https://CAIPE_DOMAIN/realms/caipe\` when \`UI_ENV_FILE\` does not have it
- Auto rollout-restart dynamic-agents after helm upgrade when \`CAIPE_DOMAIN\` is set

### Ollama defaults
- Default LLM model to \`qwen3:0.6b\` — benchmarked on CPU: 18.99 tok/s generation (3.4x faster than smollm2:1.7b), 40,960 token context window, full tool-calling support
- Expand Ollama model menu with MTEB-ranked embeddings options and smollm2 variants (noted as no-tool-support)

## Test plan

- [x] Fresh install via \`bash <(curl -fsSL ...)\` completes without truncation
- [x] Re-run (upgrade path) keeps web-ingestor enabled
- [x] \`CAIPE_DOMAIN\` + TLS cert: Keycloak login redirects work, no \`redirect_uri\` errors
- [x] Dynamic agents UI shows LLM model in selector after deploy
- [x] Chatting with Hello World agent returns a response (no 401 Invalid issuer)
- [x] Validation step passes: pods ready, UI reachable via ingress
- [x] qwen3:0.6b seeded and responding at ~19 tok/s on CPU-only kind cluster


setup-caipe.sh
```diff
@@ -6537,11 +6758,30 @@ DAEOF
...
   fi
 
-  if ! helm upgrade --install caipe "$CAIPE_OCI_REPO" "${helm_args[@]}" 2>&1; then
+  # Delete any stale helm hook Jobs from a previous failed install/upgrade.
+  # If left behind they block pre-upgrade hooks on the next run (helm waits
+  # for them and hits the deadline). Safe to delete: helm recreates them.
+  kubectl delete jobs -n "${CAIPE_NAMESPACE:-caipe}" \
+    -l 'helm.sh/hook' --ignore-not-found 2>/dev/null || true
+
+  if ! helm upgrade --install caipe "$CAIPE_OCI_REPO" --force-conflicts "${helm_args[@]}" 2>&1; then
```
sibuthomasmathew:
```suggestion
  if ! helm upgrade --install caipe "$CAIPE_OCI_REPO" --force "${helm_args[@]}" 2>&1; then
```

sibuthomasmathew (changes requested):

sibuthomasmathew:
Ran the test with the following configuration:
```bash
sibu@computeinstance-e01k64mmy8jp74v1he:~$ cat ~/.config/caipe/config.yaml 
# CAIPE setup configuration — saved by setup-caipe.sh
# Edit or delete this file to change defaults on the next run.
# API keys and passwords are NOT stored here.
cluster_context: "kind-caipe"
chart_version: "0.5.11"
deployment_mode: "all-in-one"
llm_provider: "openai"
enable_ollama: "true"
ollama_model: "qwen2.5:1.5b"
embeddings_provider: "ollama"
embeddings_model: "snowflake-arctic-embed2"
enable_rag: "false"
enable_graph_rag: "false"
enable_dynamic_agents: "true"
enable_tracing: "false"
enable_metallb: "false"
domain: "caipe.local.me"
sibu@computeinstance-e01k64mmy8jp74v1he:~$ 
```
The following are my observation during the interactive setup:
1. `deployment_mode` wasn't prompted
2. Inspite of selectng `No` for `MetalLB`, the `ingress-controller` got installed
3. The selection of agents doesn't seem to be working; it's installing `all` the agents which is the default value

sibuthomasmathew (commented):
It's commit c61aac3a215163901a50f84d57f17d16458e4d97 that stopped the prompting of `deployment mode` during the interactive setup of the first run

sriaradhyula:
Thanks @sibuthomasmathew for the detailed testing! I've pushed a fix commit addressing all three observations:

**1. `deployment_mode` not prompted**
`choose_deployment_mode` was silently skipping the prompt when the value was pre-loaded from saved config. Changed to show a confirm/change prompt (`Keep existing deployment mode 'all-in-one'? [y/n]`) so users can change it on re-run without editing the config file.

**2. Ingress-controller installed despite selecting No for MetalLB**
`ENABLE_INGRESS` was not saved or restored from `~/.config/caipe/config.yaml`. It defaulted to `true`, so on re-run the nginx-ingress install always fired even when MetalLB was disabled. Fixed:
- `enable_ingress` is now saved and restored in config
- Added a safety guard: `ENABLE_INGRESS=false` is forced whenever `enable_metallb=false` (MetalLB is a prerequisite)

**3. Agent selection not preserved across runs**
`SELECTED_AGENTS` was never saved to config, so on re-run it fell back to reading all enabled agents from the cluster (all enabled by default). Fixed by adding `selected_agents` to the saved config — the subset chosen interactively is now restored on the next run.

**Re: inline suggestion (`--force` vs `--force-conflicts`)** — keeping `--force-conflicts` intentionally. `--force` is not a valid `helm upgrade` flag in current Helm versions (it was renamed to `--force-replace`), and `--force-replace` deletes and recreates resources which is more destructive. `--force-conflicts` resolves SSA field manager conflicts without recreation, which is exactly what this PR needed.

sriaradhyula:
### Updated default Ollama model: `qwen3:0.6b`

After testing several compact models for CPU-only kind/k3s clusters, updated the default from `qwen2.5:1.5b` to **`qwen3:0.6b`**:

| Model | Size | Context | Tool support | Generation speed |
|---|---|---|---|---|
| qwen2.5:1.5b | ~1.1 GB | 4,096 tok | ✅ | ~5.5 tok/s |
| smollm2:1.7b | ~1.1 GB | 4,096 tok | ⚠️ marginal | ~5.5 tok/s |
| smollm2:135m / 360m | ~100–230 MB | — | ❌ no tools | — |
| **qwen3:0.6b** | **522 MB** | **40,960 tok** | **✅** | **~19 tok/s** |

qwen3:0.6b is ~3.4× faster on generation, uses less disk space, has a 10× larger context window, and has reliable tool-calling support — a clear improvement for the default on resource-constrained local clusters. Larger Ollama models are still selectable from the wizard menu.

sriaradhyula (commented):
All review comments addressed:
- `deployment_mode` now prompts for confirmation even when pre-loaded from saved config
- MetalLB=false correctly forces `ENABLE_INGRESS=false` (saved/restored in config)
- `SELECTED_AGENTS` persisted and restored across re-runs
- Proprietary content removed

Also added since the review:
- Default model switched to `qwen3:0.6b` (3.4× faster than previous default on CPU, 40K context, tool support)
- Embedding dimension fixes for Ollama models (snowflake-arctic-embed2, nomic-embed-text, etc.)

All CI checks passing. Could you re-review when you get a chance? 🙏

setup-caipe.sh
sriaradhyula: Accepted — you're right, `--force-conflicts` requires `--server-side` (SSA) to have any effect. Without it the flag is silently ignored. Switched to `--force` in e44bec873.

sriaradhyula: Thanks for the suggestion! We accepted `--force` initially, but during testing on Helm v4.2.1 we hit `UPGRADE FAILED: invalid operation: cannot use server-side apply and force replace together` — Helm 4 uses SSA by default, which is mutually exclusive with `--force` (force-replace). Since `--force` is not needed for a clean install (only to resolve ownership conflicts during upgrades), I removed it entirely in b6b3c0b. Happy to add back a targeted `--force-replace` path only on upgrade if you prefer that approach.

Issue #1849, opened 2026-06-15
feat: Add zizmor static analysis for GitHub Actions workflows


sriaradhyula:
## Summary

Integrate [zizmor](https://github.com/zizmorcore/zizmor), a static analysis tool for GitHub Actions, into the CI pipeline to catch security and correctness issues in workflow files before they reach `main`.

## Motivation

GitHub Actions workflows are code and carry the same security risks as application code — script injection, excessive permissions, unpinned third-party actions, and misconfigured triggers. `zizmor` provides automated, rule-based analysis specifically designed for Actions YAML, catching issues that general-purpose linters miss.

Example issues zizmor detects:

- **Script injection** via unquoted `${{ github.event.* }}` expressions in `run:` steps
- **Unpinned actions** (`uses: actions/checkout@v4` instead of a pinned SHA)
- **Overly broad permissions** (`permissions: write-all`)
- **Dangerous `pull_request_target` triggers** with checkout of untrusted code
- **Secrets exposed in environment variables** passed to untrusted steps

### 1. Add zizmor to CI (GitHub Actions workflow)

Create `.github/workflows/zizmor.yml` (or add a job to an existing lint workflow):

```yaml
name: zizmor
on:
  pull_request:
  push:
    branches: [main]

jobs:
  zizmor:
    runs-on: ubuntu-latest
    permissions:
      security-events: write   # for SARIF upload
      contents: read
    steps:
      - uses: actions/checkout@<pinned-sha>
      - name: Run zizmor
        uses: zizmorcore/zizmor-action@<pinned-sha>
        with:
          args: --format sarif .github/workflows/
        continue-on-error: false   # block on findings
      - name: Upload SARIF
        if: always()
        uses: github/codeql-action/upload-sarif@<pinned-sha>
        with:
          sarif_file: zizmor.sarif
```

SARIF upload surfaces findings directly in the PR **Security** tab and inline on diff lines.

### 2. Baseline / suppression file

On first run, zizmor will likely flag pre-existing issues. Add a `.zizmor.toml` ignore file to suppress accepted-risk findings with documented justification, rather than disabling rules globally:

```toml
# .zizmor.toml
[ignore]
# <workflow-file>:<line> = "<reason>"
".github/workflows/release.yml:42" = "controlled internal trigger, no untrusted input"
```

### 3. Fix findings surfaced on adoption

Triage and remediate findings in a follow-up PR:
- Pin all `uses:` references to commit SHAs
[...]

PR #1851, opened 2026-06-15
fix(ci): harden release workflows and add actions scan

sriaradhyula:
## Summary

- Tighten release workflow input handling and move release payload values through environment variables before shell use.
- Pin yq installation to a fixed version with SHA-256 verification.
- Replace the PR version-bump checkout PAT with the existing GitHub App token.
- Add an incremental zizmor workflow for changed GitHub Actions files with SARIF upload.

## Notes

- The new scan is scoped to changed workflow/action files so it can block new high-confidence findings without requiring the existing workflow backlog to be cleared first.
- Related: #1849

## Validation

- `uvx --from zizmor==1.25.2 zizmor --format plain --min-severity high --min-confidence high <staged workflow/action files>`
- `actionlint -ignore 'unexpected key "description"' -ignore 'shellcheck reported issue' .github/workflows/zizmor.yml .github/workflows/release-finalize.yml .github/workflows/pr-version-bump.yml .github/workflows/release-prerelease.yml .github/workflows/release-manual.yml`
- `ruby -e 'require "yaml"; ARGV.each { |f| YAML.load_file(f); puts "ok #{f}" }' .github/workflows/zizmor.yml .github/actions/update-version-files/action.yml .github/actions/prepare-release/action.yml .github/actions/determine-release-tag/action.yml`
- `git diff --check --cached`


.github/actions/determine-release-tag/action.yml
```diff
@@ -4,6 +4,8 @@ description: >
   Supports tag push (primary), workflow_dispatch, and push-to-main (legacy fallback).
   Recognises -dev.N, -rc.N, -hotfix.N, -chart.M suffixes.
 
+# assisted-by Codex Codex-sonnet-4-6
```
suwhang-cisco: Could we please avoid these kind of comments? It seems to be added on every file. Also what model was it actually? because codex-sonnet-4-6 does not exist

suwhang-cisco (commented):
I think the code looks okay. Mainly just putting the input to env and putting permission directly under the job which I believe works. But could we please clean up the AI comments? Github actions are a bit annoying to test so happy to bite the bullet, merge and see if it works. If you do have time and want to test, we'll need to fork it to a private repo and run

suwhang-cisco (approved):

PR #1859, opened 2026-06-15
fix: make AgentGateway provider-token passthrough declarative

### LINKED ISSUE #1857: AgentGateway provider-token passthrough is hardcoded to github/gitlab; make it declarative
## Summary

The AgentGateway provider-token transform (rewriting `X-CAIPE-Provider-Token` → `Authorization: Bearer` so per-user OAuth tokens / per-service-account PATs reach an upstream MCP) is **hardcoded to the agent names `github` and `gitlab`**, and is only reachable on the per-agent routing path. Any other MCP that wants per-caller token passthrough cannot get it declaratively — it requires a code change, and the "which MCP uses a per-caller token" fact is duplicated across 4–6 files that must be hand-synced.

## Where it's hardcoded today

- `charts/ai-platform-engineering/templates/_helpers.tpl` (`agentgatewayMcpTargets`, static routing): `{{- if or (eq $name "github") (eq $name "gitlab") -}}` sets `providerTokenAuth`. The `extraMcpTargets` loop never sets it at all.
- `charts/ai-platform-engineering/templates/agentgateway-mcp.yaml` (gateway-api routing): independent `{{- if or (eq $name "github") (eq $name "gitlab") }}` emitting the `AgentgatewayPolicy` transform.

## The duplication (no single source of truth)

Adding one new per-caller-token MCP today means editing the same fact in all of:

1. `_helpers.tpl` — name match (static mode)
2. `agentgateway-mcp.yaml` — name match (gateway-api mode)
3. `ui/src/lib/credentials/built-in-oauth-connectors.ts` — `BUILT_IN_PROVIDER_KEYS` allowlist (SA-token picker validates against this)
4. `ui/src/lib/rbac/agentgateway-mcp-discovery.ts` — `BUILTIN_MCP_CREDENTIAL_SOURCES`
5. `ai_platform_engineering/dynamic_agents/.../services/config.yaml` — authoritative `credential_sources`
6. `deploy/agentgateway/config.yaml` + `config_bridge.py` — docker-compose static path

`agentgateway-mcp-discovery.ts` even carries a comment that it "MUST stay in sync with config.yaml and the route transforms" — i.e. the codebase already acknowledges the drift risk.

Notably, the **dynamic-agents runtime is already fully generic**: `resolve_mcp_server_credentials` reads `credential_sources` and resolves any provider with no name-matching. Only the chart templates and the UI allowlist are pinned to names.

## Proposed fix

**Phase 1 — declarative flag (removes the chart snowflake).** Replace the `github`/`gitlab` name match in both templates with an explicit `mcp.agentgateway.providerTokenAuth` flag, and pass it through the `extraMcpTargets` loop so externally-hosted MCPs can opt in too. GitLab/GitHub then just set `providerTokenAuth: true` in their default values blocks. This is a **breaking change** (anyone who enabled github/gitlab gateway routing must add the flag) — call it out in the upgrade notes. (Chart defaults ship `agentgateway.enabled: false` for both, so no default behavior changes.)

**Phase 2 — single source of truth.** Have the chart derive `providerTokenAuth` from whether a target declares a `header`-targeted `provider_connection` / `caller_token` credential source — i.e. consume the same `credential_sources` the runtime and UI already use, surfaced into Helm values. This collapses the 4–6 hand-synced declarations into one and lets the gateway, runtime, and UI all read the same source. Also consider whether the UI `BUILT_IN_PROVIDER_KEYS` allowlist should stay curated or open up.

## Acceptance criteria

- [ ] No agent-name string matching for the provider-token transform in either routing-mode template.
- [ ] Any MCP (chart-deployed agent or `extraMcpTargets`) can opt into provider-token passthrough via config alone, no code change.
- [ ] Adding a new per-caller-token MCP requires editing one declaration, not 4–6.
- [ ] Breaking change documented in the chart upgrade notes.

sriaradhyula:
# Description

Fixes #1857.

This removes the hardcoded GitHub/GitLab name checks from the AgentGateway provider-token transform and makes passthrough declarative. MCP routes can now opt in with `providerTokenAuth: true`, or by declaring a header-targeted `credential_sources` entry with `kind: provider_connection` or `kind: caller_token`.

The chart now carries the GitHub/GitLab defaults in values instead of template name matching, passes the provider-token setting through `extraMcpTargets`, and documents the upgrade note in the changelog.

## Type of Change

- [x] Bugfix
- [x] Breaking Change
- [ ] New Feature
- [ ] Refactor
- [x] Documentation
- [ ] Other (please describe)

## Validation

- [x] `uv run pytest tests/test_agentgateway_provider_token_auth.py -q`
- [x] `uv run ruff check tests/test_agentgateway_provider_token_auth.py`
- [x] `git diff --check`

## Checklist

- [x] All new and existing relevant tests pass


kevkantes:
Nice — this fully removes the name matching from both routing modes, and deriving `providerTokenAuth` from a header-targeted `credential_sources` entry (the `agentgatewayProviderTokenAuth` helper) is the consolidation #1857 was really after: the gateway now reads the same declaration the dynamic-agents runtime resolves. Tests covering both static and gateway-api modes + both opt-in paths are 👍.

One residual from the issue's "stays in sync across 4–6 places" list: the **UI provider allowlist** isn't touched here — `BUILT_IN_PROVIDER_KEYS` (derived from `built-in-oauth-connectors.ts`), which `token-providers` and `[id]/credentials` validate against. So the chart/gateway side is now fully declarative, but a genuinely new provider still needs a UI edit before a service account can hold a token for it. GitHub/GitLab are already in that allowlist so they work end-to-end; this is only about future providers. Intended as a follow-up, or out of scope for this PR?

PR #1883, opened 2026-06-17
fix(slack): let team members manage shared integrations

sriaradhyula:
# Description

Fixes Slack integration sharing so channels assigned to a team are visible/editable to team members, while keeping admin-only onboarding/advanced surfaces out of the generic user flow. Also centralizes the Slack/Webex team-assignment authorization rules in a policy manifest and adds an OpenFGA ReBAC Policy Manifest tab to visualize those rules without showing tuple-level details by default.

## Type of Change

- [x] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Validation

- [x] `npm test -- --runTestsByPath src/lib/rbac/__tests__/authorization-policy-catalog.test.ts src/lib/rbac/__tests__/slack/slack-channel-rebac.test.ts src/lib/rbac/__tests__/webex-space-rebac.test.ts src/lib/rbac/migrations/__tests__/messaging-migrations.test.ts src/app/api/admin/rebac/__tests__/policy-catalog-route.test.ts src/components/admin/security/__tests__/OpenFgaRebacTab.test.tsx src/app/api/admin/slack/channels/__tests__/channel-resources-route.test.ts src/app/api/rbac/admin-tab-gates/__tests__/route.test.ts src/lib/rbac/__tests__/openfga.test.ts --runInBand`
- [x] `npx eslint src/lib/rbac/authorization-policy-catalog.ts src/lib/rbac/slack-channel-rebac.ts src/lib/rbac/webex-space-rebac.ts src/app/api/admin/rebac/policies/catalog/route.ts src/components/admin/security/PolicyManifestTab.tsx src/components/admin/security/OpenFgaRebacTab.tsx src/components/admin/rebac/SlackChannelRebacPanel.tsx e2e/rbac/rbac-admin-regression.spec.ts e2e/rbac/slack-run-as.spec.ts`
- [x] `git diff --check`
- [x] `RUN_RBAC_REGRESSION=1 npx playwright test e2e/rbac/slack-run-as.spec.ts e2e/rbac/rbac-admin-regression.spec.ts --config=playwright.rbac.config.ts`
- [x] `RUN_RBAC_REGRESSION=1 CAIPE_UI_BASE_URL=http://localhost:3100 npm run test:e2e:rbac-regression` against a fresh branch dev server; manifest and Slack regressions passed, workflow checks require `WORKFLOWS_ENABLED=true` at server startup.
[...]

PR #1894, opened 2026-06-17
feat(admin): scope teams, users, stats, and feedback for non-admin users

kevkantes:
# Description

Extends the team-scoped admin surfaces pattern (established in #1883 for Slack/Webex) to four remaining admin surfaces:

- **Teams & Users → Users**: non-admins see other members of their own team(s) instead of just themselves. Falls back to self-only when user has no team memberships. Returns a `scoped: "team" | "self"` field to the frontend.
- **Teams & Users → Teams**: non-admins see only teams they belong to. Per-row `can_manage` flag (from OpenFGA `team#admin` check) gates "Manage team" vs "View team" button and dialog editability.
- **Insights → Statistics**: non-admins see stats scoped to Slack channels they `can_read` in OpenFGA plus their own web conversations. Empty scope returns a valid zero-valued payload.
- **Insights → Feedback**: non-admins see feedback for their readable Slack channels plus their own web feedback. Scoped `distinct` queries prevent leaking all channel names / user emails in dropdowns.

New shared helper `ui/src/lib/rbac/user-insights-scope.ts` (`getReadableSlackChannelNames`) is used by both Stats and Feedback to avoid duplication. Follows the per-row capability flag pattern from #1883. No OpenFGA model changes required.

Unit tests added/extended for all four routes.

## Follow-up correctness fixes (review pass)

A self-review surfaced two scoping bugs in the original implementation, now fixed:

- **Users — team scope never matched (broken feature).** Team members were resolved by passing the OpenFGA `team:<slug>` slug to a `team_id`-keyed membership lookup, so it matched nobody and every non-admin silently fell back to `scoped: "self"`. Now resolves members via `listActiveTeamMembershipSourcesBySlug` (slug-keyed). The original test mocks encoded the same slug-as-id mismatch, masking the bug — rewritten to key by `team_slug`.
- **Users — full-realm scan (perf).** The team-scoped path scanned the entire Keycloak realm in 100-user batches on every non-admin Users-tab load. Now resolves each member by exact email (indexed lookup), O(team size) instead of O(realm size).
- **Stats — partial scope leak.** The non-admin scope was applied only to conversation counts; `feedback_summary`, the Slack block, `available_channels`, Slack message counts, and `total_users` all still ran platform-wide — leaking global aggregates and every channel name. The scope is now threaded through every query: `total_users` derives from scoped conversations, the Slack block is bounded to readable channels (and skipped entirely when none), and `available_channels` returns only the readable set. Also hardened a `$or` clobber in the shared-conversations count.

Added stats leak-boundary tests (feedback, Slack messages, `total_users`, `available_channels`, Slack-block probe) and a users de-dup/perf test; each was verified to fail against the pre-fix code.

## Type of Change

- [x] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/ui/src/app/api/admin/feedback/route.ts
+++ b/ui/src/app/api/admin/feedback/route.ts
@@ -128,7 +128,7 @@ export const GET = withErrorHandler(async (request: NextRequest) => {
           entries: [],
           channels: [],
           users: [],
-          pagination: { page: 1, limit: 50, total: 0, total_pages: 0 },
+          pagination: { page, limit, total: 0, total_pages: 0 },
         });
       }
       if (filter.$or) {
--- a/ui/src/app/api/admin/feedback/route.ts
+++ b/ui/src/app/api/admin/feedback/route.ts
@@ -11,6 +11,7 @@ withErrorHandler,
 } from '@/lib/api-middleware';
 import { getConfig } from '@/lib/config';
 import { getCollection,isMongoDBConfigured } from '@/lib/mongodb';
+import { getReadableSlackChannelNames } from '@/lib/rbac/user-insights-scope';
 import { NextRequest,NextResponse } from 'next/server';
 
 export const GET = withErrorHandler(async (request: NextRequest) => {
@@ -33,7 +34,25 @@ export const GET = withErrorHandler(async (request: NextRequest) => {
   }
 
   const { session } = await getAuthFromBearerOrSession(request);
-  await requireRbacPermission(session, 'admin_ui', 'view');
+  const isFullAdmin = await requireRbacPermission(session, 'admin_ui', 'view').then(
+    () => true,
+    () => false
+  );
+
+  let scopedChannelNames: string[] | null = null;
+  let scopedOwnerEmail: string | null = null;
+  if (!isFullAdmin) {
+    const sub = typeof session.sub === 'string' ? session.sub.trim() : '';
+    const email = typeof session.user?.email === 'string' ? session.user.email.trim() : '';
+    if (!sub) {
+      return NextResponse.json(
+        { success: false, error: 'Unauthorized', code: 'UNAUTHORIZED' },
+        { status: 401 }
+      );
+    }
+    scopedChannelNames = await getReadableSlackChannelNames(`user:${sub}`);
+    scopedOwnerEmail = email || null;
+  }
 
     const { searchParams } = new URL(request.url);
     const rating = searchParams.get('rating'); // 'positive' | 'negative' | null (all)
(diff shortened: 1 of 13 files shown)


ui/src/lib/rbac/user-insights-scope.ts
```diff
@@ -0,0 +1,36 @@
...
+      .find({ active: { $ne: false } } as never)
+      .limit(500)
+      .toArray();
+
+    const names: string[] = [];
+    for (const row of rows) {
+      if (!row.slack_channel_id || !row.channel_name) continue;
+      const object = `slack_channel:${slackChannelSubjectId(row.slack_workspace_id ?? '', row.slack_channel_id)}`;
+      const result = await checkOpenFgaTuple({ user: openfgaUser, relation: 'can_read', object }).catch(() => ({ allowed: false }));
+      if (result.allowed) names.push(row.channel_name);
```
sriaradhyula, line 30: [P1] Returning only channel names collapses distinct Slack channel resources. The OpenFGA check is against `slack_channel:<workspace>--<channel>`, but the downstream stats/feedback filters use only `channel_name`. In a multi-workspace deployment, or with duplicate names like `general`, `can_read` on one channel will also include feedback/stats for another channel with the same name. Please carry the workspace/channel IDs through this scope and filter by `channel_id`/workspace where available, using `channel_name` only as display or legacy fallback.

ui/src/app/api/admin/feedback/route.ts
```diff
@@ -90,6 +109,44 @@ export const GET = withErrorHandler(async (request: NextRequest) => {
...
+      if (scopedChannelNames && scopedChannelNames.length > 0) {
+        scopeClauses.push({
+          source: 'slack',
+          channel_name: scopedChannelNames.length === 1
+            ? scopedChannelNames[0]
+            : { $in: scopedChannelNames },
+        });
+      }
+      if (scopedOwnerEmail) {
+        scopeClauses.push({ user_email: scopedOwnerEmail });
```
sriaradhyula, line 124: [P1] This "own feedback" branch is not limited to web feedback. Slack feedback writes `user_email` too, so a non-admin can read feedback from unreadable Slack channels whenever that email matches, despite the scoped contract being readable Slack channels plus own web feedback. This clause should be constrained to web feedback, for example `{ source: "web", user_email: scopedOwnerEmail }`, leaving Slack visibility to the readable-channel branch.

ui/src/app/api/admin/stats/route.ts
```diff
@@ -96,6 +131,79 @@ export const GET = withErrorHandler(async (request: NextRequest) => {
...
+            completion_rate: 0,
+            avg_messages_per_workflow: 0,
+          },
+          available_channels: [],
+        });
+      }
+
+      nonAdminScopeFilter = scopeClauses.length === 1 ? scopeClauses[0] : { $or: scopeClauses };
+      andInto(convSourceFilter, nonAdminScopeFilter);
+      andInto(msgOwnerFilter, nonAdminScopeFilter);
```
sriaradhyula, line 204: [P2] This applies the conversation-shaped scope to `messages` queries, so readable-channel Slack messages get dropped from message-derived stats. The scope includes fields like `source`, `client_type`, `slack_meta.channel_name`, and `metadata.channel_name`, but the message documents queried below do not carry those conversation fields. As a result, `total_messages`, daily/hourly Slack messages, top agents, latency, workflow counts, and web-hours automation only see the `owner_id === email` branch instead of the readable Slack channels. Please derive scoped conversation IDs first and filter messages by `conversation_id`, or keep a separate message scope that matches fields actually present on `messages`.

ui/src/app/api/admin/stats/route.ts
```diff
@@ -180,6 +304,22 @@ export const GET = withErrorHandler(async (request: NextRequest) => {
...
+      const fbScope: Record<string, unknown>[] = [];
+      if (nonAdminChannelNames.length > 0) {
+        fbScope.push({
+          source: 'slack',
+          channel_name: nonAdminChannelNames.length === 1
+            ? nonAdminChannelNames[0]
+            : { $in: nonAdminChannelNames },
+        });
+      }
+      if (nonAdminScope.ownerEmail) fbScope.push({ user_email: nonAdminScope.ownerEmail });
```
sriaradhyula, line 319: [P1] Same feedback-scope issue here: this owner-email branch is not web-only, but Slack feedback rows also store `user_email`. A scoped user can therefore influence `feedback_summary`, `by_source`, categories, and daily feedback with Slack feedback from channels they cannot read. Please constrain the owner branch to web feedback, and keep Slack feedback bounded only by the readable Slack channel scope.

ui/src/app/api/admin/stats/route.ts
```diff
@@ -499,7 +650,7 @@ export const GET = withErrorHandler(async (request: NextRequest) => {
         ];
         delete slackFilter.created_at;
       }
-      const slackHasData = await conversations.countDocuments(SLACK_CONV_MATCH, { limit: 1 });
+      const slackHasData = skipSlackBlock ? 0 : await conversations.countDocuments(SLACK_CONV_MATCH, { limit: 1 });
```
sriaradhyula, line 653: [P1] For scoped users with at least one readable channel, this still probes global Slack data and then renders the Slack block, whose `slack.channels` summary is populated from the platform-wide `platform_config` totals below. That leaks global Slack channel/config counts even though the rest of the block is intended to be channel-scoped. Please make this probe use the scoped `slackFilter` and either derive the channel summary from `slackChannelScope` or omit that summary for non-admin responses.

sriaradhyula (approved):

PR #1907, opened 2026-06-18
fix(docker-compose): Update to wire openfga into first install ui

sriaradhyula:
## Summary
- use the published CAIPE UI release image in `docker-compose.yaml`
- pass OpenFGA connection defaults into the UI service for the local RBAC profile
- switch the UI healthcheck to `127.0.0.1` so it does not fail on container-local IPv6 localhost resolution

## Why
Fresh OSS all-in-one installs can reach the migrations page but ReBAC migrations fail with `OPENFGA_HTTP is not set`. The production compose file also referenced a `-prod` UI image tag that is not published for the release image, which forces local overrides before first launch works.

## Validation
- `COMPOSE_PROFILES="mcp-servers,caipe-ui-prod,rbac,caipe-supervisor,dynamic-agents,rag,caipe-mongodb" NEXTAUTH_SECRET=test docker compose --env-file /Users/sraradhy/outshift/ai-platform-engineering/.env -f docker-compose.yaml config --quiet`
- confirmed rendered UI config resolves `ghcr.io/cnoe-io/caipe-ui:0.5.16`, `OPENFGA_HTTP=http://openfga:8080`, `OPENFGA_STORE_NAME=caipe-openfga`, and the `127.0.0.1` healthcheck

### DIFF
--- a/.claude/skills/README.md
+++ b/.claude/skills/README.md
@@ -60,6 +60,21 @@ docker restart caipe-supervisor  # after agents are warm
 
 See [integration-testing/SKILL.md](./integration-testing/SKILL.md) for full testing procedure.
 
+### 🐳 [docker-compose-first-install](./docker-compose-first-install/)
+Validate and repair the plain OSS first-install path for `docker-compose.yaml`
+and `.env.example`.
+
+**Quick Start:**
+```bash
+COMPOSE_PROFILES="mcp-servers,caipe-ui-prod,rbac,caipe-supervisor,dynamic-agents,rag,caipe-mongodb" \
+NEXTAUTH_SECRET=test \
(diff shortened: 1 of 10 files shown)


sibuthomasmathew:
Deployed the `docker-compose` from this branch and ran into the following issue:

1. `keycloak-init` was crashing due to the missing environment variables -` IDP_ISSUER`, `IDP_CLIENT_ID`, and `IDP_CLIENT_SECRET`. Added a fix, however, not sure whether it breaks `kind` or production deployment

---

Added a tiny option to download the latest tag using `curl`, if `gh` CLI is missing

However, I haven't tested the RAG option; I do it after setting up an embedding model

sibuthomasmathew (approved):
lgtm. `rag-server` came up without errors for an embedding model by Ollama

PR #1917, opened 2026-06-18
feat(audit): configurable storage backend — local disk and S3 (closes #1903)

### LINKED ISSUE #1903: Store audit logs outside MongoDB using configurable S3 or local disk backend
## Feature request

Move audit log storage out of MongoDB/DocumentDB and into a configurable storage backend, initially S3 or local disk.

## Motivation

The dev environment is already showing UI slowdown and occasional hangs while only lightly used. CPU and memory do not appear saturated, so one suspected contributor is heavy audit log read/write activity against MongoDB/DocumentDB.

Writing high-volume audit logs to the primary app database can become expensive and may create avoidable performance pressure. We should address this before production so we can avoid a future migration and potentially delete existing audit logs from MongoDB.

## Proposed behavior

Add configuration that controls where audit logs are stored:

- S3/object storage for shared or production-like deployments
- Local disk for lightweight/local deployments
- MongoDB should no longer be the default long-term audit log storage target

## Notes

Potential storage pattern for S3/object storage:

- Write compressed audit log files, such as parquet or another compact format
- Use bucket lifecycle/tiering policies for cold storage and cost control
- Keep the app database focused on operational state instead of append-heavy audit history

## Source context

Slack discussion noted that DocumentDB should be scaling, but the UI slowdown in dev suggests audit log writes/reads may still be taxing the database. S3 was suggested as a near-term feature before production rollout.

sriaradhyula:
## Description

Moves audit log persistence out of MongoDB and into a pluggable storage backend controlled by a single environment variable. Addresses the UI slowdown in the dev environment caused by heavy audit log read/write activity against MongoDB/DocumentDB.

Closes #1903

### New abstraction layer
- `ai_platform_engineering/utils/audit_backend.py` — `AuditBackend` Protocol + `get_audit_backend()` factory (Python)
- `ui/src/lib/audit/backend.ts` — `AuditBackend` interface + `getAuditBackend()` factory (TypeScript)

### Three backends

| Backend | Config | Writes to |
|---------|--------|-----------|
| `mongodb` | default / unset | MongoDB `audit_events` + `authorization_decision_records` (unchanged) |
| `local` | `AUDIT_LOG_BACKEND=local` | `<AUDIT_LOG_LOCAL_PATH>/YYYY/MM/DD/<type>-<date>-<uuid>.ndjson` |
| `s3` | `AUDIT_LOG_BACKEND=s3` | `s3://<bucket>/<prefix>/YYYY/MM/DD/<type>-<ts>-<uuid>.ndjson.gz` |

### Wire-up
- `audit_logger.py` — `_persist_to_mongo()` replaced with `get_audit_backend().write()`
- `ui/src/lib/rbac/audit.ts` — dual `persistAuthzDecisionToMongo` + `persistToUnifiedAuditEvents` calls replaced with a single `getAuditBackend().write()`
- `ui/src/lib/credentials/audit.ts` — direct MongoDB insert replaced with backend write; `auditCollection` dependency removed from `SecretService`

### IRSA support
S3 backend uses the standard AWS credential chain — IRSA (`AWS_ROLE_ARN` + `AWS_WEB_IDENTITY_TOKEN_FILE`) is picked up automatically by boto3 and `@aws-sdk/client-s3`. Helm `serviceAccount.annotations` guidance added to `supervisor-agent/values.yaml`.

### All writes remain fire-and-forget
Both Python and TypeScript backends are non-blocking (`void asyncOp()` in TS, exception-swallowing in Python). Backend errors are logged, never propagated to the caller.

## Type of Change

- [x] New Feature

## Test plan

- [x] 12 Python unit tests pass (`test_audit_backend.py`, `test_audit_logger.py`)
- [x] TypeScript tests added for `LocalBackend` and `S3Backend` (`ui/src/lib/audit/__tests__/`)
- [x] Existing credential tests updated for removed `auditCollection` dependency
- [x] `uv run ruff check` — clean
- [x] `AUDIT_LOG_BACKEND` unset keeps MongoDB behaviour unchanged (no regression)

## Checklist

- [x] Existing issues have been referenced (closes #1903)
- [x] Functionality is documented (spec + contracts in `docs/docs/specs/2026-06-18-audit-logs-local-storage/`)


ai_platform_engineering/utils/audit_backends/__init__.py
```diff
@@ -0,0 +1,13 @@
+# assisted-by claude code claude-sonnet-4-6
+from ai_platform_engineering.utils.audit_backend import AuditBackend, get_audit_backend
+from ai_platform_engineering.utils.audit_backends.local_backend import LocalBackend
+from ai_platform_engineering.utils.audit_backends.mongo_backend import MongoBackend
```
kevkantes: Imo we shouldn't support MongoBackend, as it will cause perf issues

kevkantes:
Can we also remove AI noise like: 
`# assisted-by claude code claude-sonnet-4-6` everywhere

ai_platform_engineering/utils/audit_backends/s3_backend.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def _build_client(self) -> Any:
+        import boto3  # type: ignore[import-untyped]
+
+        kwargs: Dict[str, Any] = {"region_name": self._region}
+        if self._endpoint_url:
+            kwargs["endpoint_url"] = self._endpoint_url
+        return boto3.client("s3", **kwargs)
+
+    def write(self, event: Dict[str, Any]) -> None:
```
kevkantes, line 101: This issues one `put_object` call per audit event. At meaningful event volume this will hit S3 PUT rate limits and inflate costs significantly (S3 pricing is per-request). Consider buffering events in a deque and flushing on a time interval or count threshold (e.g. every 100 events or 5 seconds), writing them as a single multi-event NDJSON object.

ui/src/lib/audit/backends/s3-backend.ts
```diff
@@ -0,0 +1,91 @@
...
+
+      const eventType = String(event["type"] ?? "audit");
+      const keyUuid = randomUUID().replace(/-/g, "").slice(0, 12);
+      const key = `${this.prefix}/${yyyy}/${mm}/${dd}/${eventType}-${tsCompact}-${keyUuid}.ndjson.gz`;
+
+      const serialized = JSON.stringify(event, (_k, v) =>
+        v instanceof Date ? v.toISOString() : v
+      );
+      const body = await gzipAsync(Buffer.from(serialized + "\n", "utf-8"));
+
```
kevkantes, line 92: Same one-PUT-per-event concern as the Python backend: every `write()` call triggers a separate `PutObjectCommand`. Batch events into an in-memory buffer and flush on an interval or size threshold to avoid per-request cost blow-up and rate limiting under load.

ai_platform_engineering/utils/audit_backends/s3_backend.py
```diff
@@ -0,0 +1,93 @@
+# Copyright 2025 CNOE Contributors
+# SPDX-License-Identifier: Apache-2.0
+# assisted-by claude code claude-sonnet-4-6
+"""S3-compatible audit log backend.
+
+Writes each event as a gzip-compressed single-line NDJSON object to::
+
```
kevkantes, line 7: Nit / future consideration: gzip-compressed NDJSON works fine for archiving, but if you ever need to query these logs with Athena / Spark / DuckDB, Parquet (e.g. via `pyarrow`) gives you columnar pushdown and significantly smaller files (5–10× vs gzip-NDJSON on structured audit payloads). Worth noting in the spec as a follow-up if query-side tooling is on the roadmap.

ui/src/lib/audit/backends/s3-backend.ts
```diff
@@ -0,0 +1,91 @@
+// assisted-by claude code claude-sonnet-4-6
+/**
+ * S3-compatible audit log backend.
+ *
+ * Writes each event as a gzip-compressed single-line NDJSON object to:
+ *   s3://<bucket>/<prefix>/YYYY/MM/DD/<type>-<YYYYMMDDTHHMMSSZ>-<uuid>.ndjson.gz
+ *
```
kevkantes, line 12: Same format note as the Python backend: gzip-NDJSON is fine for archiving, but if Athena / DuckDB queries are ever needed, Parquet would be worth considering. Flag this in the spec as a potential follow-up.

kevkantes: AI said potential follow up, but I think we should do this now. Better compression now will save us later

ai_platform_engineering/utils/audit_backends/s3_backend.py
```diff
@@ -0,0 +1,93 @@
...
+
+def _serialize(obj: Any) -> Any:
+    if isinstance(obj, datetime):
+        return obj.astimezone(timezone.utc).isoformat()
+    raise TypeError(f"Object of type {type(obj).__name__} is not JSON serialisable")
+
+
+class S3Backend:
+    """Writes audit events as gzip-compressed NDJSON objects to S3."""
```
kevkantes: The `endpoint_url` parameter makes this work with MinIO, but it's still S3-SDK-only. If native Azure Blob Storage or GCS (without S3 compat mode) is a realistic deployment target, worth considering a `BlobBackend` abstraction or at minimum a separate `AzureBlobBackend` class rather than cramming everything through the boto3 path. The `AuditBackend` Protocol already supports this cleanly — just noting it's not covered today.

ui/src/lib/audit/backends/s3-backend.ts
```diff
@@ -0,0 +1,91 @@
...
+
+function toIsoDate(ts: unknown): string {
+  if (ts instanceof Date) return ts.toISOString();
+  if (typeof ts === "string") return ts;
+  return new Date().toISOString();
+}
+
+export class S3Backend implements AuditBackend {
+  private readonly client: S3Client;
+  private readonly bucket: string;
```
kevkantes, line 36: Same as the Python side: `endpoint` + `forcePathStyle` covers MinIO but not native Azure Blob or GCS. The `AuditBackend` interface leaves room for additional backends cleanly — just flag that this isn't truly cloud-agnostic blob storage yet.

ai_platform_engineering/utils/audit_backends/s3_backend.py
```diff
@@ -0,0 +1,93 @@
...
+                f"{event_type}-{ts_compact}-{key_uuid}.ndjson.gz"
+            )
+
+            body = gzip.compress(
+                (json.dumps(event, default=_serialize) + "\n").encode("utf-8")
+            )
+
+            self._client.put_object(
+                Bucket=self._bucket,
+                Key=key,
```
kevkantes, line 142: Setting `ContentType: application/x-ndjson` + `ContentEncoding: gzip` together is valid HTTP, but S3 consumers (Athena, boto3 `get_object`, browser downloads) treat `ContentEncoding: gzip` as a _transfer encoding_ and will auto-decompress transparently, meaning callers see raw NDJSON bytes and the `.gz` extension in the key becomes misleading. The alternative is `ContentType: application/gzip` (no `ContentEncoding`) which tells consumers the object _is_ a gzip file and forces explicit decompression. Pick whichever matches your downstream tooling, but be intentional — the current combo may cause double-decompression surprises.

ui/src/lib/audit/backends/s3-backend.ts
```diff
@@ -0,0 +1,91 @@
...
+      const body = await gzipAsync(Buffer.from(serialized + "\n", "utf-8"));
+
+      await this.client.send(
+        new PutObjectCommand({
+          Bucket: this.bucket,
+          Key: key,
+          Body: body,
+          ContentType: "application/x-ndjson",
+          ContentEncoding: "gzip",
+        }),
```
kevkantes, line 103: Same `ContentType: application/x-ndjson` + `ContentEncoding: gzip` issue as the Python backend. The AWS SDK and Athena interpret `ContentEncoding: gzip` as a transfer encoding and auto-decompress on read, which makes the `.ndjson.gz` key extension misleading and can cause double-decompression. Consider `ContentType: application/gzip` (drop `ContentEncoding`) if you want the object treated as an opaque gzip file.

ai_platform_engineering/utils/audit_backends/local_backend.py
```diff
@@ -0,0 +1,66 @@
...
+            os.makedirs(date_dir, exist_ok=True)
+
+            filename = f"{event_type}-{utc.strftime('%Y%m%d')}-{self._file_uuid}.ndjson"
+            filepath = os.path.join(date_dir, filename)
+
+            line = json.dumps(event, default=_serialize) + "\n"
+            with open(filepath, "a", encoding="utf-8") as fh:
+                fh.write(line)
+        except Exception as exc:  # noqa: BLE001
+            logger.warning(f"[audit/local] Failed to write audit event: {exc}")
```
kevkantes, line 71: Appending without a file lock is safe on CPython for writes that fit in a single `fh.write()` call due to the GIL, but if this backend is ever used from multiple threads or async tasks writing the same file concurrently, lines can interleave. A `threading.Lock` keyed by filepath (or a per-instance lock since the filename is stable per process) would make this explicit and safe without much overhead.

tests/test_audit_backend.py
```diff
@@ -0,0 +1,152 @@
+# Copyright 2025 CNOE Contributors
```
kevkantes: This file is byte-for-byte identical to `ai_platform_engineering/utils/test_audit_backend.py`. One of them should be removed — having two copies means fixes/additions will inevitably get made in only one place.

kevkantes (changes requested):

sriaradhyula:
Superseded by the smaller stacked PR series below. This original branch collected the full audit storage/backend work in one large PR, but we split it into reviewable slices and will continue with those instead.

Replacement stack:

1. #1945 chore(repo): remove beads issue tracking → main
2. #1946 feat(audit): add lightweight audit service runtime → #1945
3. #1947 refactor(audit): route producers through audit service → #1946
4. #1948 feat(audit): add audit-service read UI → #1947
5. #1949 perf(ui): cache health and RBAC gate checks → #1948
6. #1950 fix(rbac): address admin and chat regressions → #1949
7. #1951 docs(audit): document audit performance benchmarks → #1950

Closing this PR so review and CI focus stay on the split stack.

PR #1939, opened 2026-06-19
fix(rag): add OIDC SSL bypass and hostAliases

alexanghh:
# Description

## Problem

When deploying CAIPE on local Kind clusters with self-signed certificates (or any environment without trusted CA chains), the RAG server and ingestors fail OIDC token validation because `httpx` and `aiohttp` reject the self-signed TLS certificates used by Keycloak. Additionally, the `rag-server` pod cannot resolve the public `CAIPE_DOMAIN` hostname to reach the in-cluster Keycloak/ingress, unlike `caipe-ui` which already has a `hostAliases` patch.

## Solution

- **`OIDC_VERIFY_SSL`** env var for the RAG server's `OIDCProvider` (`auth.py`) — controls `httpx.AsyncClient(verify=...)` on JWKS and discovery fetches
- **`INGESTOR_OIDC_VERIFY_SSL`** env var for the ingestor's OAuth2 client (`ingestor.py`) — controls `aiohttp.TCPConnector(ssl=...)` on discovery and token fetches
- **`hostAliases` patch** for `rag-server` deployment in `setup-caipe.sh` — maps `CAIPE_DOMAIN` → nginx ClusterIP so the pod can reach Keycloak via the ingress hostname (matching the existing `caipe-ui` pattern)
- **`setup-caipe.sh` wiring** — reads `OIDC_VERIFY_SSL` / `INGESTOR_OIDC_VERIFY_SSL` from `.env`, stores them in the appropriate K8s secrets, and maps to `NODE_TLS_REJECT_UNAUTHORIZED` for the UI

### Files changed

| File | Change |
|---|---|
| `rag/common/src/common/ingestor.py` | Add `INGESTOR_OIDC_VERIFY_SSL` env, disable SSL on `TCPConnector` when `false` |
| `rag/server/src/server/auth.py` | Add `OIDC_VERIFY_SSL` env, pass `verify=` to `httpx.AsyncClient` |
| `setup-caipe.sh` | Wire both env vars through secrets, `hostAliases` for `rag-server`, `NODE_TLS_REJECT_UNAUTHORIZED` for UI |

Assisted-by: Antigravity:gemini-3.5-flash

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/ai_platform_engineering/knowledge_bases/rag/common/src/common/ingestor.py
+++ b/ai_platform_engineering/knowledge_bases/rag/common/src/common/ingestor.py
@@ -48,6 +48,8 @@ class Client:
     self.oidc_discovery_url = os.getenv("INGESTOR_OIDC_DISCOVERY_URL")
     # Scope is optional - if not set, don't send any scope (many providers don't need it for client credentials)
     self.oidc_scope = os.getenv("INGESTOR_OIDC_SCOPE", "")
+    # Verify SSL option
+    self.oidc_verify_ssl = os.getenv("INGESTOR_OIDC_VERIFY_SSL", "true").lower() == "true"
 
     # Token cache
     self._access_token: Optional[str] = None
@@ -73,6 +75,7 @@ class Client:
       logger.info(f"  - INGESTOR_OIDC_CLIENT_ID: {self.oidc_client_id}")
       logger.info(f"  - INGESTOR_OIDC_CLIENT_SECRET: {'***' if self.oidc_client_secret else 'NOT SET'}")
       logger.info(f"  - INGESTOR_OIDC_SCOPE: {self.oidc_scope if self.oidc_scope else '(none - will omit scope parameter)'}")
+      logger.info(f"  - INGESTOR_OIDC_VERIFY_SSL: {self.oidc_verify_ssl}")
       logger.info("  - Authentication mode: OAUTH2 (will send authenticated requests)")
     else:
       logger.info(f"Ingestor '{self.ingestor_name}': OAuth2 client credentials NOT fully configured")
@@ -189,7 +192,8 @@ class Client:
     Raises:
         Exception if fetch fails or token_endpoint not found
     """
-    async with aiohttp.ClientSession() as session:
+    connector = aiohttp.TCPConnector(ssl=None if self.oidc_verify_ssl else False)
+    async with aiohttp.ClientSession(connector=connector) as session:
       async with session.get(discovery_url, allow_redirects=True) as resp:
         if resp.status == 404:
           raise aiohttp.ClientError(f"Discovery endpoint not found (404): {discovery_url}")
@@ -238,7 +242,8 @@ class Client:
     logger.info(f"Ingestor '{self.ingestor_name}': Fetching new OAuth2 access token from {token_endpoint}")
 
     try:
-      async with aiohttp.ClientSession() as session:
+      connector = aiohttp.TCPConnector(ssl=None if self.oidc_verify_ssl else False)
+      async with aiohttp.ClientSession(connector=connector) as session:
         # Build token request data
         token_data = {
           "grant_type": "client_credentials",
--- a/ai_platform_engineering/knowledge_bases/rag/common/src/common/ingestor.py
+++ b/ai_platform_engineering/knowledge_bases/rag/common/src/common/ingestor.py
@@ -49,7 +49,13 @@ class Client:
     # Scope is optional - if not set, don't send any scope (many providers don't need it for client credentials)
     self.oidc_scope = os.getenv("INGESTOR_OIDC_SCOPE", "")
     # Verify SSL option
-    self.oidc_verify_ssl = os.getenv("INGESTOR_OIDC_VERIFY_SSL", "true").lower() == "true"
+    raw_verify_ssl = os.getenv("INGESTOR_OIDC_VERIFY_SSL", "true").strip().lower()
+    if raw_verify_ssl in {"true", "1", "yes"}:
+      self.oidc_verify_ssl = True
+    elif raw_verify_ssl in {"false", "0", "no"}:
+      self.oidc_verify_ssl = False
+    else:
+      raise ValueError("INGESTOR_OIDC_VERIFY_SSL must be one of: true, false, 1, 0, yes, no")
 
     # Token cache
     self._access_token: Optional[str] = None
--- a/ai_platform_engineering/knowledge_bases/rag/server/src/server/auth.py
+++ b/ai_platform_engineering/knowledge_bases/rag/server/src/server/auth.py
@@ -48,6 +48,7 @@ class OIDCProvider:
     self.jwks_cache: Dict[str, Any] = {}
     self.jwks_cache_time: float = 0
     self.jwks_cache_ttl: int = 3600  # Cache JWKS for 1 hour
+    self.verify_ssl = os.getenv("OIDC_VERIFY_SSL", "true").lower() == "true"
 
     if discovery_url:
       logger.info(f"Initialized OIDC provider '{name}': issuer={issuer}, audience={audience}, discovery_url={discovery_url}")
@@ -122,7 +123,7 @@ class OIDCProvider:
 
     # Fetch JWKS
     logger.debug(f"Fetching JWKS from {self.jwks_uri}")
-    async with httpx.AsyncClient(follow_redirects=True) as client:
+    async with httpx.AsyncClient(follow_redirects=True, verify=self.verify_ssl) as client:
       response = await client.get(self.jwks_uri, timeout=10.0)
       response.raise_for_status()
       return response.json()
@@ -140,7 +141,7 @@ class OIDCProvider:
     Raises:
         Exception if fetch fails
     """
-    async with httpx.AsyncClient(follow_redirects=True) as client:
+    async with httpx.AsyncClient(follow_redirects=True, verify=self.verify_ssl) as client:
       response = await client.get(well_known_url, timeout=10.0)
       response.raise_for_status()
       return response.json()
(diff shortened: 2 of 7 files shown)


ai_platform_engineering/knowledge_bases/rag/server/src/server/auth.py
```diff
@@ -48,6 +48,7 @@ def __init__(
     self.jwks_cache: Dict[str, Any] = {}
     self.jwks_cache_time: float = 0
     self.jwks_cache_ttl: int = 3600  # Cache JWKS for 1 hour
+    self.verify_ssl = os.getenv("OIDC_VERIFY_SSL", "true").lower() == "true"
```
sriaradhyula: This parsing makes every value except exactly `"true"` disable TLS verification. That includes common operator values like `1`, `yes`, or `True `, and also typos. For OIDC discovery/JWKS fetching, invalid config should fail closed instead of silently disabling verification.

```suggestion
    raw_verify_ssl = os.getenv("OIDC_VERIFY_SSL", "true").strip().lower()
    if raw_verify_ssl in {"true", "1", "yes"}:
      self.verify_ssl = True
    elif raw_verify_ssl in {"false", "0", "no"}:
      self.verify_ssl = False
    else:
      raise ValueError("OIDC_VERIFY_SSL must be one of: true, false, 1, 0, yes, no")
```

ai_platform_engineering/knowledge_bases/rag/common/src/common/ingestor.py
```diff
@@ -48,6 +48,8 @@ def __init__(self, ingestor_name: str, ingestor_type: str, ingestor_description:
     self.oidc_discovery_url = os.getenv("INGESTOR_OIDC_DISCOVERY_URL")
     # Scope is optional - if not set, don't send any scope (many providers don't need it for client credentials)
     self.oidc_scope = os.getenv("INGESTOR_OIDC_SCOPE", "")
+    # Verify SSL option
+    self.oidc_verify_ssl = os.getenv("INGESTOR_OIDC_VERIFY_SSL", "true").lower() == "true"
```
sriaradhyula: Same fail-open parsing issue here: a typo or trailing space disables TLS verification for token discovery/acquisition. Please normalize known boolean values and reject invalid values.

```suggestion
    raw_verify_ssl = os.getenv("INGESTOR_OIDC_VERIFY_SSL", "true").strip().lower()
    if raw_verify_ssl in {"true", "1", "yes"}:
      self.oidc_verify_ssl = True
    elif raw_verify_ssl in {"false", "0", "no"}:
      self.oidc_verify_ssl = False
    else:
      raise ValueError("INGESTOR_OIDC_VERIFY_SSL must be one of: true, false, 1, 0, yes, no")
```

sriaradhyula (commented):
Thanks for the focused local Kind/RAG auth fix. The hostAliases direction looks reasonable; the main concern is making TLS bypass behavior explicit, scoped, and fail-closed so a dev workaround does not accidentally weaken other deployments.

setup-caipe.sh
```diff
@@ -6072,12 +6091,31 @@ deploy_caipe() {
...
+    if [[ "$OIDC_VERIFY_SSL" == "false" || "$OIDC_VERIFY_SSL" == "0" ]]; then
+      reject_unauthorized="0"
+    else
+      reject_unauthorized="1"
+    fi
+  else
+    if [[ -z "$TLS_CERT_FILE" || -z "$TLS_KEY_FILE" ]]; then
+      reject_unauthorized="0"
+    fi
+  fi
```
sriaradhyula: This block has two problems: the bool parsing is case-sensitive/fail-open, and the self-signed detection does not work for the generated cert path because `setup_tls()` sets `TLS_CERT_FILE` and `TLS_KEY_FILE` to `/tmp/caipe-tls-*`. Please normalize the env value and track generated self-signed certs explicitly, e.g. set `TLS_SELF_SIGNED=true` in `setup_tls()` when the script generates the cert.

```suggestion
  local reject_unauthorized=""
  local oidc_verify_ssl_normalized=""
  if [[ -n "$OIDC_VERIFY_SSL" ]]; then
    case "$(printf '%s' "$OIDC_VERIFY_SSL" | tr '[:upper:]' '[:lower:]')" in
      true|1|yes) oidc_verify_ssl_normalized="true" ;;
      false|0|no) oidc_verify_ssl_normalized="false" ;;
      *) err "OIDC_VERIFY_SSL must be true/false, 1/0, or yes/no"; exit 1 ;;
    esac

    if [[ "$oidc_verify_ssl_normalized" == "false" ]]; then
      reject_unauthorized="0"
    else
      reject_unauthorized="1"
    fi
  elif [[ "${TLS_SELF_SIGNED:-false}" == "true" ]]; then
    reject_unauthorized="0"
  fi
```

setup-caipe.sh
```diff
@@ -6072,12 +6091,31 @@ deploy_caipe() {
...
+
   local helm_args=(
     --namespace caipe
     --version "$CAIPE_CHART_VERSION"
     --set tags.caipe-ui=true
     --set tags.agent-weather=false
     --set tags.agent-netutils=true
+    ${reject_unauthorized:+--set "caipe-ui.config.NODE_TLS_REJECT_UNAUTHORIZED=${reject_unauthorized}"}
```
sriaradhyula: Mapping `OIDC_VERIFY_SSL=false` to `NODE_TLS_REJECT_UNAUTHORIZED=0` disables TLS verification for the entire Next.js server process, not just OIDC/Keycloak calls. That is a much broader blast radius than the local self-signed workaround needs. Please prefer mounting/trusting the generated CA/cert, or otherwise scoping the bypass to the specific OIDC client path rather than setting this global Node.js env var.

sriaradhyula (changes requested):
Please check the comments

setup-caipe.sh
alexanghh: I have removed NODE_TLS_REJECT_UNAUTHORIZED=0 and mounted the generated self-signed TLS certificates to the UI container.

setup-caipe.sh
alexanghh: Fixed in the latest commit.

ai_platform_engineering/knowledge_bases/rag/common/src/common/ingestor.py
alexanghh: Fixed in the latest commit.

ai_platform_engineering/knowledge_bases/rag/server/src/server/auth.py
alexanghh: Fixed in the latest commit.

ai_platform_engineering/knowledge_bases/rag/common/src/common/ingestor.py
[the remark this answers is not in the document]
alexanghh: Added test in the latest commit.

ai_platform_engineering/knowledge_bases/rag/server/src/server/auth.py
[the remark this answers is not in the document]
alexanghh: Added test in the latest commit.

setup-caipe.sh
```diff
@@ -190,6 +190,8 @@ CAIPE_DOMAIN_DEFAULT="${CAIPE_DOMAIN_DEFAULT:-caipe.local.me}"
 CAIPE_DOMAIN=""
 TLS_CERT_FILE=""
 TLS_KEY_FILE=""
+OIDC_VERIFY_SSL="${OIDC_VERIFY_SSL:-}"
+INGESTOR_OIDC_VERIFY_SSL="${INGESTOR_OIDC_VERIFY_SSL:-}"
```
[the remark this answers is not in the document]
alexanghh: Added docs to .env.example and README.md (in knowledge_bases/rag/server)

setup-caipe.sh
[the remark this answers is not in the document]
alexanghh: Fixed in the latest commit.

sriaradhyula:
Thanks @alexanghh for this, and apologies it sat for so long. Closing as **superseded** — most of it has since landed or is in flight through other PRs. Recording the state here so nothing gets lost.

### Where each part stands on `main` today

| This PR | Status |
|---|---|
| `TLS_SELF_SIGNED` flag set in `setup_tls()` | Already on `main` (`setup-caipe.sh:202`, `:2191`) via #2677 |
| caipe-ui trusting the setup-generated self-signed cert | Already on `main` (`setup-caipe.sh:4258-4261`) via #2677 |
| `hostAliases` for `rag-server` | Covered by #2675, which also extends it to `caipe-rag-ingestors-webloader` |
| `OIDC_VERIFY_SSL` / `INGESTOR_OIDC_VERIFY_SSL` (Python) | **Not on `main`** — see below |

The `setup-caipe.sh` half is now a large conflict against a file that has moved a lot since June, so rebasing it isn't worth it.

### The part that is still an open problem

The Python change here is the only proposal anywhere for the failure #2675 calls out as a known follow-up: once the hostAlias lands, the web-ingestor still fails OIDC with `SSL: CERTIFICATE_VERIFY_FAILED` against the self-signed cert. Nothing on `main` addresses that yet.

Your second commit did address the fail-closed parsing feedback properly, and added tests for both parsers — that work is not wasted, it just needs a different shape. My hesitation on merging it as-is: `OIDC_VERIFY_SSL=false` / `INGESTOR_OIDC_VERIFY_SSL=false` are documented, discoverable "turn off TLS verification" knobs in a public repo, and they will eventually get set somewhere that isn't a Kind cluster. I'd rather fix it by trusting the generated CA inside the rag pods (mount `caipe-tls`, set `SSL_CERT_FILE` / `REQUESTS_CA_BUNDLE`) so verification stays on everywhere.

### One thing from here I do want to pick up

Your latest revision replaced `NODE_TLS_REJECT_UNAUTHORIZED=0` with a `caipe-tls` mount plus `NODE_EXTRA_CA_CERTS` for caipe-ui. That is the tighter approach, and it's better than what #2677 actually merged. I'll pull that in as a follow-up and credit you.

If you'd like to take the CA-trust fix for the rag pods, please do — happy to review it quickly this time. Otherwise I'll pick it up. Thanks again for finding and diagnosing this.

PR #1970, opened 2026-06-22
ci(rag): build only changed RAG components instead of all three

dabcoder:
## Summary

- `ci-a2a-rag.yml` rebuilt **all** RAG images (`agent-ontology`, `server`, `ingestors` + `slim`/`huggingface` variants) on *any* change under `rag/**`, because of a single coarse `paths-filter` feeding a static all-component matrix. Now it uses **per-component path filters** and a **dynamically computed build/retag matrix** — touching one component builds only that one.
- A change to the shared `common/` library still fans out to all three (all Dockerfiles `COPY common`).
- First pre-release, final release, `workflow_dispatch`, and `force_build` still build everything (release safety unchanged).
- `prebuild-a2a-rag.yml`: added the missing `ingestors` filter + `CHANGED_INGESTORS` env (ingestors-only changes previously never triggered a prebuild), and removed the dead `CHANGED_AGENT_RAG` leftover.

## Why

Editing one RAG component (e.g. `rag-server`) wasted CI rebuilding the unrelated `agent-ontology` and `ingestors` images on every PR and dev build.

## Behavior (verified by local simulation of the matrix logic)

| Trigger | Builds | Retags |
|---|---|---|
| PR, only `server/` changed | `server` (+huggingface) | — |
| PR, `common/` changed | all 3 | — |
| `-dev.3`, only `ingestors/` | `ingestors` (+slim) | `agent-ontology`, `server` |
| first `-dev.1` / final release | all 3 | — |
| `workflow_dispatch` / `force_build` | all 3 | — |
| chart-only tag | — | — |

_LLM used: Claude Opus 4.8_

### DIFF
--- a/.github/workflows/prebuild-a2a-rag.yml
+++ b/.github/workflows/prebuild-a2a-rag.yml
@@ -75,6 +75,9 @@ jobs:
             server:
               - 'ai_platform_engineering/knowledge_bases/rag/server/**'
               - 'ai_platform_engineering/knowledge_bases/rag/build/Dockerfile.server'
+            ingestors:
+              - 'ai_platform_engineering/knowledge_bases/rag/ingestors/**'
+              - 'ai_platform_engineering/knowledge_bases/rag/build/Dockerfile.ingestors'
 
 
       - name: Set matrix based on changes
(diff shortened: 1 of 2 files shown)


.github/workflows/prebuild-a2a-rag.yml
```diff
@@ -75,16 +74,19 @@ jobs:
             server:
               - 'ai_platform_engineering/knowledge_bases/rag/server/**'
               - 'ai_platform_engineering/knowledge_bases/rag/build/Dockerfile.server'
+            ingestors:
```
suwhang-cisco, line 77: Oh interesting, did we never build ingrester prebuild image then??

suwhang-cisco (approved):

.github/workflows/prebuild-a2a-rag.yml
dabcoder: Not in the path detection, no. But any PR targetting `'.github/**'` allowed to build all RAG components including the ingestors, hence why I was able to build and rebuild that component's container images previously.

PR #2000, opened 2026-06-23
fix(rbac): atomically acquire migration lock to prevent multi-replica race

sriaradhyula:
## Problem

When `replicaCount > 1`, all pods start simultaneously. Each calls `recordRunning` — a plain `updateOne` with `upsert: true` that has no guard. All pods see status ≠ `running/completed`, all proceed through the same Keycloak API calls, and the one that hits a conflict surfaces `warnings=1`, which shows as **keycloak-bootstrap: fetch failed** in the Platform Health widget.

## Fix

Replace `recordRunning` with `acquireMigrationLock`: an atomic `findOneAndUpdate` that sets `status=running` only when the document is absent or not already `running`/`completed`.

```
filter:
  status NOT IN [running, completed]
  OR (status = running AND locked_at < now - 5min)
```

- **Only one pod proceeds** — the others get `null` or an E11000 duplicate-key error (document exists but filter rejected it) and return `{ status: "skipped" }` immediately.
- **5-minute stale-lock TTL** — if a pod crashes mid-migration, the `locked_at` field lets the next deploy reclaim the lock after 5 minutes rather than being permanently blocked.

## Workaround (until merged)

Set `caipe-ui.replicaCount: 1` before upgrading, scale back to 3 after the first pod is healthy.

## Test plan

- [ ] Deploy with `replicaCount: 3` and confirm only one pod logs `completed`, others log `skipped`
- [ ] Confirm Platform Health widget shows keycloak-bootstrap as healthy after upgrade
- [ ] Confirm a single-replica deploy is unaffected

### DIFF
--- a/ui/e2e/rbac/platform-health-probes.spec.ts
+++ b/ui/e2e/rbac/platform-health-probes.spec.ts
@@ -95,8 +95,8 @@ const healthyProbes: Probe[] = [
     label: "Keycloak Bootstrap",
     group: "bootstrap",
     status: "healthy",
-    detail: "Realm and clients ready",
-    target: "caipe realm",
+    detail: "completed by caipe-ui-0 · 07:31 UTC",
+    target: "caipe",
     latency_ms: 13,
   },
   {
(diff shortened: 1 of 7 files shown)


sriaradhyula:
The core fix () landed on main as e438fa9cd. Closing as superseded.

PR #2003, opened 2026-06-23
fix(mcp): GitLab credential probe fallback and SA tool picker fixes

cisco-erilutz:
## Description

Three fixes to get the GitLab MCP integration working end-to-end with service accounts, plus a pre-existing test fix:

**1. MCP credential probe fallback (`ui/src/lib/mcp-credential-headers.ts`)**

`McpCredentialUnavailableError` was being rethrown before the `fallback_env` code path could execute. When no user credential is registered (expected for service accounts and fresh installs), the probe was failing instead of falling back to the `GITLAB_PERSONAL_ACCESS_TOKEN` environment variable. Fixed to only hard-fail for pinned connections or truly unexpected errors.

**2. SA tool picker always shows wildcard (`ui/src/app/api/admin/service-accounts/grantable/route.ts`)**

The `server/*` wildcard was only included in the grantable tools list when the tool catalog was empty. Once the probe ran and populated the catalog, the wildcard disappeared — removing the user's ability to grant all tools at once. Fixed to always prepend the wildcard regardless of catalog state.

**3. INVOKE_PERSIST_HISTORY compose env var (`docker-compose.dev.yaml`)**

Wires the `INVOKE_PERSIST_HISTORY` flag through to the dynamic-agents container so multi-turn history can be enabled for local dev without code changes.

**4. Fix pre-existing RBAC test failure (`ui/src/lib/rbac/__tests__/keycloak-rbac-reconciliation.test.ts`)**

The `fix(rbac): atomically acquire migration lock` commit added `findOneAndUpdate` to the production code but didn't update the test mock's `createCollection` helper. All three reconciliation tests were failing with `TypeError: migrations.findOneAndUpdate is not a function`.

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


ui/src/lib/mcp-credential-headers.ts
```diff
@@ -131,7 +131,12 @@ async function resolveSourceCredential(
         };
       }
     } catch (error) {
-      if (pinned || isMcpCredentialUnavailableError(error)) {
+      if (pinned) {
+        throw error;
+      }
+      // Non-pinned: a missing/unconnected provider connection is expected when
+      // the caller hasn't registered a credential yet — fall through to fallback_env.
+      if (!isMcpCredentialUnavailableError(error)) {
```
cisco-erilutz, line 139: do we need an `else throw error` now below in the `if (fallbackEnv)` block in case there isn't a fallback env?

cisco-erilutz: No — if there's no `fallback_env` (or it's empty), the function falls through to `return { credential: "", origin: "none" }` a few lines below. That's the intended signal for "no credential available" — the caller decides whether to 401 or continue. Throwing here would crash the probe rather than gracefully reporting no credential.

sriaradhyula (approved):

PR #2007, opened 2026-06-24
fix(chart): wire AGENTGATEWAY_TARGETS_TOKEN into caipe-ui and add dev-mode default

sriaradhyula:
## Problem

The `/api/internal/agentgateway/mcp-targets` endpoint in caipe-ui returns HTTP 503 immediately when `AGENTGATEWAY_TARGETS_TOKEN` is not set — before checking the caller's Authorization header. This means the agentgateway config-bridge sidecar can never authenticate against caipe-ui, even in an otherwise working cluster.

Two related gaps:
- `AGENTGATEWAY_TARGETS_TOKEN` was completely absent from the caipe-ui deployment template
- The agentgateway config-bridge sidecar had no fallback token when `global.agentgateway.static.configBridge.bff.existingSecret.name` was not configured, so in dev/demo deployments both sides had no token at all

## Fix

**caipe-ui `deployment.yaml`** — add `AGENTGATEWAY_TARGETS_TOKEN` to the `env:` block alongside `AGENTGATEWAY_TARGETS_URL`:
- When `global.agentgateway.static.configBridge.bff.existingSecret.name` is set: mount from that Secret (mirrors the agentgateway sidecar's existing logic
- Otherwise: fall back to the well-known dev-mode default , overridable via  in values

**agentgateway ** — add the same dev-mode default for the config-bridge sidecar when no  is configured

Both sides now use a matching token in a default/dev installation with no Secret required. Production operators who set  continue to get their secret-mounted token on both sides, unchanged.

## Test plan

- [ ]  with no  configured: both caipe-ui and config-bridge sidecar env blocks show 
- [ ]  with : both containers use 
- [ ]  with  and no : caipe-ui uses  (config override respected)
- [ ] Fresh cluster install:  agentgateway-config-bridge check returns 200 instead of 503

EOF
)


kevkantes (dismissed):

charts/ai-platform-engineering/charts/caipe-ui/templates/deployment.yaml
```diff
@@ -138,6 +138,20 @@ spec:
...
+            {{- /*
+            Wire AGENTGATEWAY_TARGETS_TOKEN so /api/internal/agentgateway/mcp-targets can
+            authenticate config-bridge poll requests. Use the operator-provided Secret when
+            supplied; otherwise reference the auto-generated Secret created by the parent chart
+            (charts/ai-platform-engineering/templates/agentgateway-config-bridge-secret.yaml).
+            */ -}}
+            {{- $cbBffSecret := ((((.Values.global.agentgateway).static).configBridge).bff).existingSecret | default dict }}
+            {{- $cbTokenSecretName := $cbBffSecret.name | default (printf "%s-agentgateway-config-bridge-token" .Release.Name) }}
+            {{- $cbTokenSecretKey := $cbBffSecret.key | default "AGENTGATEWAY_TARGETS_TOKEN" }}
+            - name: AGENTGATEWAY_TARGETS_TOKEN
```
kevkantes, line 153: **Unconditional `secretKeyRef` can break caipe-ui when config-bridge is disabled.**

This env block has no guard, but the Secret it references (`<release>-agentgateway-config-bridge-token`) is only created when `global.agentgateway.enabled && configBridge.enabled` (auto-gen template) or when `existingSecret.name` is set. Chart defaults have both enabled, so the default install is fine — but anyone running caipe-ui with agentgateway or config-bridge **disabled** will get a caipe-ui pod stuck in `CreateContainerConfigError`, since the referenced Secret won't exist.

Suggest either wrapping this block in the same `and $agwGlobal.enabled $cb.enabled` condition used by the auto-gen Secret template, or adding `optional: true` to the `secretKeyRef`.

charts/ai-platform-engineering/templates/agentgateway-config-bridge-secret.yaml
```diff
@@ -0,0 +1,27 @@
...
+{{- $existing := lookup "v1" "Secret" .Release.Namespace $secretName }}
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {{ $secretName }}
+  labels:
+    {{- include "ai-platform-engineering.labels" . | nindent 4 }}
+type: Opaque
+data:
+  AGENTGATEWAY_TARGETS_TOKEN: {{ if $existing }}{{ index $existing.data "AGENTGATEWAY_TARGETS_TOKEN" }}{{ else }}{{ randAlphaNum 32 | b64enc }}{{ end }}
```
kevkantes: **`lookup` + `randAlphaNum` is incompatible with rendered / GitOps flows.**

The `lookup`-to-avoid-rotation pattern returns empty under `helm template`, `helm diff`, `--dry-run`, and ArgoCD/Flux, so a fresh `randAlphaNum 32` renders on every such invocation. That produces noisy diffs and — on a partial apply where only one side re-renders — a token mismatch between the config-bridge sidecar and caipe-ui.

This is the well-known Bitnami pattern so it's "accepted" for live `helm install/upgrade`, and it doesn't affect operators who set `existingSecret`. But it's worth documenting as a limitation: GitOps users should set `existingSecret.name` or the ESO path rather than rely on auto-generation.

charts/ai-platform-engineering/charts/caipe-ui/templates/deployment.yaml
sriaradhyula: Fixed in 0046fb0 — wrapped the  block in `{{- if and $agwGlobal.enabled $cb.enabled }}` so the env var is only rendered when the Secret exists. Verified with `helm template` that token is absent when agentgateway/configBridge disabled.

charts/ai-platform-engineering/templates/agentgateway-config-bridge-secret.yaml
sriaradhyula: Added a NOTE block to the template comment in 0046fb0 documenting the `lookup`/GitOps limitation and pointing GitOps users to `bff.existingSecret.name` or `bff.externalSecret.enabled`.

charts/ai-platform-engineering/charts/caipe-ui/templates/deployment.yaml
sriaradhyula: Fixed in 0046fb0 — wrapped the secretKeyRef block in the same guard used by the auto-gen Secret template (agentgateway.enabled && configBridge.enabled) so the env var is only rendered when the Secret exists. Verified with helm template that the token is absent when agentgateway or configBridge is disabled.

sriaradhyula:
Reviewed the diff. Two blocking issues and a few non-blocking ones.

## Blocking

### 1. `authzHost` / `authzHostSecret` values are orphaned

`values.yaml` adds:

```yaml
authzHost: ""
authzHostSecret:
  name: ""
  key: AGENTGATEWAY_AUTHZ_HOST
```

No template in this PR (or in the current chart) renders these into a container env. The agentgateway deployment diff only touches the token wiring, not the authz host. Either add the corresponding `deployment.yaml` change to wire `AGENTGATEWAY_AUTHZ_HOST` into the config-bridge sidecar env, or drop these values from this PR and ship them separately with their template.

### 2. Asymmetric `configBridge.enabled` guard between caipe-ui and the sidecar

`caipe-ui/templates/deployment.yaml` only injects `AGENTGATEWAY_TARGETS_TOKEN` when **both** `global.agentgateway.enabled` and `configBridge.enabled` are true. That guard is correct and prevents `CreateContainerConfigError` when config-bridge is off.

`agentgateway/templates/deployment.yaml` removes the old `{{- if existingSecret.name }}` conditional and now **always** injects the token (unconditionally, as long as the config-bridge container block is rendered). If a deployer enables agentgateway but has not explicitly set `configBridge.enabled: true`, the sidecar gets a token secretKeyRef but caipe-ui does not inject the matching env — 503, same failure this PR fixes. The sidecar injection should be gated on `configBridge.enabled` the same way caipe-ui is.

## Non-blocking

**`lookup`-based auto-gen token breaks under GitOps (ArgoCD / Flux):** The PR comment already notes this. Both `dev` and `edge` overlays use `existingSecret` pointing at the caipe-ui ExternalSecret-created secret, so the auto-gen path is not exercised in practice. Consider flipping the default so auto-generation is opt-in and the `existingSecret` / `externalSecret` path is the documented first-class recommendation — it avoids the token-rotation footgun entirely.

**`externalSecret.data: []` default with no guard:** If an operator sets `externalSecret.enabled: true` and forgets to populate `data`, the template renders a valid ExternalSecret with no mappings. ESO creates an empty Secret, both sides get an empty token, and the failure is silent (no error, just 503). Worth a `{{- if not $bffES.data }}{{- fail "..." }}{{- end }}` guard in the template.

**`# Use your key path` comment in `values-external-secrets.yaml`:** Reads as unfinished. Use a clearly fake placeholder like `path/to/your/agentgateway-bridge-secret` or a `NOTE:` block.

sriaradhyula:
Hey @kevkantes — addressed the GitOps token issue by removing auto-generation entirely (consistent with how every other secret in this chart is handled). Latest changes:

- Deleted the `lookup`/`randAlphaNum` auto-gen Secret — no more noisy diffs or token mismatches
- Renamed `configBridge.bff` → `configBridge.ui` (clearer name)
- Added `helm fail` in both agentgateway and caipe-ui with an actionable error when the secret isn't set
- Both pods read from `global.agentgateway.static.configBridge.ui.existingSecret` — set once, shared by both

Operators create the secret once:
```
TOKEN=$(openssl rand -base64 32)
kubectl create secret generic <name> --from-literal=AGENTGATEWAY_TARGETS_TOKEN=$TOKEN
```

Could you re-approve when you get a chance?

kevkantes (dismissed):

PR #2100, opened 2026-07-01
feat(conversations): open Slack conversations to all authenticated users

cisco-erilutz:
_[GAI]_

## Context

Slack threads are inherently multi-participant — anyone in the channel should be able to interact with the bot in the thread, not just the user who originally triggered it.

Previously, only the conversation owner (the user whose email was stamped at creation time) could write to a conversation. Any other user in the same Slack thread who tried to invoke the agent would receive a 403 `conversation#write` error.

Two changes in this PR:

**1. `user:* writer` grant on Slack conversation creation** (`conversations/route.ts`)

When a new conversation is created with `client_type === 'slack'`, a `user:* writer conversation:<id>` OpenFGA tuple is written (best-effort, non-blocking). This gives any authenticated user write access to the conversation so they can invoke the agent within the thread without needing an explicit per-user share grant.

The idempotency-hit path (existing conversation returned for a repeated `thread_ts`) also heals pre-existing Slack conversations by writing the same tuple if missing.

This is safe because agent, MCP tool, and knowledge base access are all independently gated by their own authorization checks. Getting write access to a conversation container does not grant access to any tools or data the agent controls.

**2. Tighten `PATCH /metadata` to `can_manage`** (`[id]/metadata/route.ts`)

Changed the metadata endpoint's permission check from `can_write` to `can_manage` (owner/manager only). This prevents the `user:*` writer grant from enabling arbitrary metadata key injection — only the conversation owner or an explicit manager can mutate metadata.

## Validation

- TypeScript: clean (`tsc --noEmit`)
- Tests: `routes.test` suite (66 tests) and `chat-conversation-metadata-auth` — all pass

### DIFF
--- a/ui/src/app/api/chat/conversations/[id]/metadata/route.ts
+++ b/ui/src/app/api/chat/conversations/[id]/metadata/route.ts
@@ -15,7 +15,7 @@ validateUUID,
 withErrorHandler,
 } from '@/lib/api-middleware';
 import { getCollection,isMongoDBConfigured } from '@/lib/mongodb';
-import { requireConversationResourcePermission } from '@/lib/rbac/conversation-implicit-authz';
+import { isImplicitConversationOwner } from '@/lib/rbac/conversation-implicit-authz';
 import type { Conversation,PatchConversationMetadataRequest } from '@/types/mongodb';
 import { NextRequest,NextResponse } from 'next/server';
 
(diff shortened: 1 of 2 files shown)


cisco-erilutz:
Superseded by #2101 — a simpler and safer approach: bypass the `conversation#write` check for `client_type === 'slack'` instead of granting `user:*` / `service_account:*` writer tuples.

The wildcard-tuple approach here leaked read access: because the OpenFGA model defines `can_read: … or can_write`, a `user:*` writer tuple transitively grants read on the conversation, exposing Slack thread history to any authenticated user. #2101 touches only the write path, so reads stay ReBAC-protected, and needs no model change or idempotency-path heal loop. #2101 is on a `prebuild/` branch for direct dev image testing.

PR #2136, opened 2026-07-07
feat(agentgateway): fall back to caller Authorization for provider-token routes

cisco-erilutz:
# Description

The AgentGateway `providerTokenAuth` transform **unconditionally** rewrites the upstream `Authorization` header from `X-CAIPE-Provider-Token`:

```
authorization: '"Bearer " + default(request.headers["x-caipe-provider-token"], "")'
```

This means any caller that authenticates with a plain `Authorization: Bearer <token>` — but does not also set `X-CAIPE-Provider-Token` — has its token replaced with a literal empty `Bearer `, and the upstream MCP (e.g. the RAG/knowledge-base server, which enforces its own OIDC) rejects it with a 401.

This is fine for the primary path (the runtime injects `X-CAIPE-Provider-Token`), but it makes direct callers — CLI clients, scripts, local development — awkward: they must duplicate their token into a second, non-obvious header.

## Change

Make the transform **prefer** `X-CAIPE-Provider-Token` when present, and otherwise **fall back** to the caller's existing (already gateway-validated) `Authorization` header, unchanged:

```
authorization: 'default(request.headers["x-caipe-provider-token"], "") != "" ? "Bearer " + request.headers["x-caipe-provider-token"] : default(request.headers["authorization"], "")'
```

- **Backward compatible** — whenever `X-CAIPE-Provider-Token` is set (the injected path), it still wins, byte-for-byte unchanged.
- **No new privilege** — the fallback token is the same one the gateway's `jwtAuth` already validated; the upstream independently re-validates it. A caller can only forward the identity they authenticated with.
- The raw provider token is `Bearer `-prefixed as before; the fallback `Authorization` value already carries its own `Bearer ` prefix, so it is forwarded as-is (no double prefix).

Applied to all four transform sites (3 Gateway-API policies in `agentgateway-mcp.yaml`, 1 in the standalone static config) plus the `values.yaml` comments. Verified with `helm template` in both `static` and `gateway-api` routing modes.

## Type of Change

- [x] Bugfix

## Checklist

### DIFF
--- a/charts/ai-platform-engineering/templates/agentgateway-mcp.yaml
+++ b/charts/ai-platform-engineering/templates/agentgateway-mcp.yaml
@@ -130,6 +130,9 @@ spec:
 ---
 # Per-request provider/identity token: rewrite X-CAIPE-Provider-Token into the
 # upstream Authorization header for MCPs that opt into caller-token passthrough.
+# When X-CAIPE-Provider-Token is absent/empty, fall back to the caller's existing
+# Authorization header (already validated by the gateway's jwtAuth) so direct
+# callers can use a plain Bearer token without setting the provider-token header.
 apiVersion: agentgateway.dev/v1alpha1
 kind: AgentgatewayPolicy
 metadata:
@@ -144,7 +147,7 @@ spec:
       request:
         set:
           - name: authorization
-            value: '"Bearer " + default(request.headers["x-caipe-provider-token"], "")'
+            value: 'default(request.headers["x-caipe-provider-token"], "") != "" ? "Bearer " + request.headers["x-caipe-provider-token"] : default(request.headers["authorization"], "")'
 {{- end }}
 {{- end }}
 {{- end }}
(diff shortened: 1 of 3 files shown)


cisco-erilutz:
Closing this — the approach is not feasible with agentgateway's JWT policy.

`jwtAuth` runs at the listener level and **removes the validated `Authorization` header** before any route-level transformation runs (`crates/agentgateway/src/http/jwt.rs`: `self.location.remove(req)`, unconditional, no opt-out). By the time the `providerTokenAuth` transform evaluates, `request.headers["authorization"]` is already gone, so the proposed fallback always resolves to an empty string and the upstream MCP rejects the request.

Verified live: a request sent with only `Authorization: Bearer <token>` passes the gateway's JWT gate (jwt.sub populated) but reaches the upstream with an empty Authorization → 401.

This is exactly why `X-CAIPE-Provider-Token` exists — it's a distinct header that survives the JWT gate and is the intended channel for per-caller token passthrough. Callers that need their identity to reach the upstream MCP must send it on `X-CAIPE-Provider-Token`; there's no way to derive it from the stripped `Authorization` header at the route layer.

No functional change was warranted, so closing.

PR #2185, opened 2026-07-10
fix(rbac): check slack channel FGA grant for service-account agent authz

sriaradhyula:
## Summary

- `requireAgentUsePermission` never consulted the `slack_channel:<workspace>--<channel> user agent:<id>` FGA tuple when the caller is a service account (SA), causing 403 `pdp_denied` even for channels with a correctly configured agent route
- Added `slackChannelId` / `slackWorkspaceId` params to `AgentUsePermissionInput`; when provided and the SA direct check fails, the channel-level FGA grant is checked as a fallback
- The conversation POST handler extracts `channel_id` from `body.metadata.channel_id` (sent by the Slack bot) and passes it to the authz check

## Root Cause

The Slack admin UI writes two things when a channel is assigned an agent:
1. MongoDB `slack_channel_agent_routes` row (used by the bot to pick the agent)
2. OpenFGA tuple `slack_channel:<workspace>--<channel> user agent:<id>` (intended to authorize SA callers)

But `requireAgentUsePermission` only checked `service_account:<sub> can_use agent:<id>` for SA callers — the channel-level tuple was written but never read on the authz path. SA callers get a 403 unless they have a *direct* per-SA grant or the agent has `user:*` (wildcard/org visibility).

## Test Plan
- [ ] Slack bot SA creates conversation for a channel with a route configured → 200 (was 403)
- [ ] Slack bot SA for a channel with NO route → still 403
- [ ] Human user (non-SA) auth path unchanged — team-union fallback still runs

### DIFF
--- a/docs/src/pages/index.tsx
+++ b/docs/src/pages/index.tsx
@@ -6,7 +6,7 @@ import Heading from '@theme/Heading';
 import styles from './index.module.css';
 
 const CURL_CMD = 'bash <(curl -fsSL https://raw.githubusercontent.com/cnoe-io/ai-platform-engineering/main/setup-caipe.sh)';
-const HELM_CMD = 'helm upgrade --install ai-platform-engineering \\\n    oci://ghcr.io/cnoe-io/charts/ai-platform-engineering \\\n    --version 0.5.47 -f your-values.yaml';
+const HELM_CMD = 'helm upgrade --install ai-platform-engineering \\\n    oci://ghcr.io/cnoe-io/charts/ai-platform-engineering \\\n    --version 0.5.48 -f your-values.yaml';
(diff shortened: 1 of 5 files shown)


sriaradhyula:
Closing in favor of #2189 which enables `SLACK_RBAC_ENABLED=true` by default in the chart.

With OBO on by default, the Slack bot authenticates as the human Slack user (`user:<sub>`) rather than the bot service account, so per-user FGA grants and team membership checks apply directly — the channel-level SA fallback this PR adds is no longer needed for correctly-configured deployments.

The real fix is in the identity layer (#2189 + cisco-eti/platform-apps-deployment#747), not in expanding SA authorization.

PR #2186, opened 2026-07-10
prebuild: slack-channel-agent-authz-fix

sriaradhyula:
Prebuild PR to build caipe-ui image with fix for slack channel SA agent authz (see PR #2185).


sriaradhyula:
Closing — the prebuild image `ghcr.io/cnoe-io/prebuild/caipe-ui:slack-channel-agent-authz-fix-3` was built from this branch and pinned in cisco-eti/platform-apps-deployment#747 as a temporary measure alongside enabling OBO (#2189). Once #2189 merges and the chart is released, the prebuild pin can be removed.

PR #2215, opened 2026-07-13
fix(slack-bot): deduplicate and truncate oversized messages in thread context

sriaradhyula:
## Summary

Two issues in `slack_context.py` that cause oversized context being sent to the model:

1. **Message duplication**: `build_thread_context` included the triggering message in the history loop AND appended it as "Current question" — doubling large pastes. Fixed by adding `current_ts` param and filtering that ts from history.

2. **No per-message size cap**: A single large paste could inflate context to the point of hitting Bedrock token limits. Fixed by `_truncate_message()` which caps any individual message to `MESSAGE_CHAR_LIMIT` chars (default 2000, configurable via `SLACK_INTEGRATION_MESSAGE_CHAR_LIMIT` env var).

3. **Bug found during E2E**: When the thread had exactly one message and it was the current (triggering) message, filtering by `current_ts` left the list empty *after* the initial empty-list guard, causing a bare `"Previous conversation:\n---\n---\nCurrent question:"` scaffold to be returned instead of just the message. Fixed by adding a second early-return guard after the `current_ts` filter.

## Unit tests

12 tests in `ai_platform_engineering/integrations/slack_bot/tests/test_slack_context.py` (11 original + 1 regression test for the single-message bug):

```
collected 12 items
tests/test_slack_context.py ............ [100%]
12 passed in 0.09s
```

## E2E tests on live docker-compose stack (32/32 passed)

Run directly inside the `slack-bot` container against the installed code:

**`_truncate_message` — truncation fires (positive):**
| Input | Result |
|---|---|
| 2001-char message | kept part = 2000 chars, suffix `[... truncated, 1 characters omitted]` ✅ |
| 5000-char message | kept = 2000, suffix reports 3000 omitted ✅ |
| Custom `limit=10` on 20-char string | truncated at 10 ✅ |
| Unicode string (3000 `あ` chars) | truncated without crash ✅ |
| Newline before truncation point preserved | ✅ |

**`_truncate_message` — no truncation (negative):**
| Input | Result |
|---|---|
| Short message `"hello"` | unchanged ✅ |
| Empty string | unchanged ✅ |
| Exactly 2000 chars | unchanged ✅ |
| 1999 chars | unchanged ✅ |
| `custom limit=10`, exactly 10-char string | unchanged ✅ |

**`build_thread_context` — deduplication:**
| Case | Result |
|---|---|
| `current_ts` set: triggering msg appears exactly once | ✅ |
| Appears in `"Current question:"` section only | ✅ |
| Other (prior) messages still included in history | ✅ |
| `current_ts` not set: legacy duplication preserved | ✅ |

**`build_thread_context` — truncation:**
| Case | Result |
|---|---|
| 5000-char history message truncated | ✅ |
| Truncation marker `[... truncated,` present | ✅ |
| Current question appended untruncated | ✅ |

**`build_thread_context` — edge cases:**
| Case | Result |
|---|---|
| Empty thread → returns `current_message` | ✅ |
| Single-message thread, message is current (filtered) → returns `current_message` directly (regression fix) | ✅ |
| Non-matching `current_ts` → message stays in history | ✅ |
| After filtering, other messages still included | ✅ |

**`build_delta_context`:**
| Case | Result |
|---|---|
| 5000-char delta message truncated | ✅ |
| Truncation marker in delta context | ✅ |
| Current question appended | ✅ |
| Empty delta → returns `current_message` | ✅ |
| Short message not truncated; no marker | ✅ |

**`MESSAGE_CHAR_LIMIT`:** default is 2000 ✅

## Docker image

Built and verified locally: `ghcr.io/cnoe-io/caipe-slack-bot:pr-2215`

## Test plan

- [x] Unit tests: 12 pass (including regression test for single-message thread bug)
- [x] E2E: 32/32 cases pass inside live container
- [x] Docker image builds cleanly
- [ ] Staging: large paste in Slack thread confirms truncation note, no duplication
- [ ] Confirm `SLACK_INTEGRATION_MESSAGE_CHAR_LIMIT` env var respected at runtime

### DIFF
--- a/ai_platform_engineering/integrations/slack_bot/app.py
+++ b/ai_platform_engineering/integrations/slack_bot/app.py
@@ -1392,7 +1392,7 @@ def handle_mention(event, say, client, context=None):
     context_message = message_text
     if event.get("thread_ts"):
       if conv_created:
-        context_message = slack_context.build_thread_context(app, channel_id, thread_ts, message_text, bot_user_id)
+        context_message = slack_context.build_thread_context(app, channel_id, thread_ts, message_text, bot_user_id, current_ts=event.get("ts"))
       else:
         since_ts = conv_metadata.get("last_processed_ts", thread_ts)
         context_message = slack_context.build_delta_context(app, channel_id, thread_ts, message_text, bot_user_id, since_ts=since_ts)
@@ -1902,7 +1902,7 @@ def handle_dm_message(event, say, client, context=None):
     context_message = message_text
     if event.get("thread_ts"):
       if conv_created:
-        context_message = slack_context.build_thread_context(app, channel_id, thread_ts, message_text, bot_user_id)
+        context_message = slack_context.build_thread_context(app, channel_id, thread_ts, message_text, bot_user_id, current_ts=event.get("ts"))
       else:
         since_ts = conv_metadata.get("last_processed_ts", thread_ts)
         context_message = slack_context.build_delta_context(app, channel_id, thread_ts, message_text, bot_user_id, since_ts=since_ts)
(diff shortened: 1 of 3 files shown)


sriaradhyula:
@kevkantes @ErikMLutz - Do we still need this?

PR #2216, opened 2026-07-14
feat(dynamic-agents): support multimodal file (text files, images, documents) chat input

jasonc3-sys:
# Description

Adds multimodal file input (image + PDF + text, alongside the message) to the dynamic_agents chat path, so vision/document-capable models can receive uploaded files. **The instructions below let you build, run, and validate it locally yourself — no external services needed.**

Code changes
- ChatRequest gains an optional files: list[InputFile] field (each file: required mime_type + base64 data or a uri).
- AgentRuntime.stream() builds a LangChain-standard multimodal content list (text block + one block per file) for the user turn; the plain-string path is unchanged when no files are attached.
- Images map to LangChain image blocks; PDFs/docs to document blocks — the langchain_aws Bedrock adapter converts both natively. Unsupported types are skipped with a warning; if nothing survives, it degrades to the plain string.
- Slack bot forwards native Slack attachments into the same files[] seam.
- Logic is factored into a pure _build_user_content helper with unit tests.

Note: The web UI paperclip wiring works for upload, but local dev runs SSO-off ("No Auth") so the browser carries no bearer and the request can't reach the agent yet. That browser auth path is a follow-up PR. This PR validates the full backend transport via the same gateway route, using a minted service-account bearer.

# Local Validation
**One-time setup**
Enter repo root directory
`make e2e-test-minimal`
**Run**
`./tests/fixtures/multimodal/validate_multimodal.sh`

**e2e playwright test commands**
Enter repo root directory

```
RUN_RBAC_E2E=1 npx playwright test \
  --config=playwright.rbac.config.ts chat-multimodal-input.spec.ts
```
```
RUN_RBAC_E2E=1 npx playwright test \
  --config=playwright.rbac.config.ts chat-multimodal-input.spec.ts --headed
```
`npx playwright show-report`
```
open "$(find test-results -name 'multimodal-jpeg-composer-chip.png' | head -1)"
```
```
open "$(find test-results -name 'multimodal-jpeg-sent.png' | head -1)"
```

The script (see header comments) mints a service-account bearer, creates a conversation, and uploads each committed fixture (tests/fixtures/multimodal/sample.{txt,jpeg,pdf}) through the same UI-gateway → dynamic-agents route the browser and Slack bot use, printing the model's answer for each. It applies two idempotent local-dev fixups so a fresh stack reproduces cleanly:

(a) aliases caipe-ui-prod → caipe-ui on the docker network — e2e-test-minimal boots the caipe-ui hot profile, but dynamic-agents' AUTHZ_SERVICE_URL points at caipe-ui-prod, so without the alias authz fails closed (503). (Alternative: boot make caipe-ui-prod.)
(b) grants the OpenFGA service_account → agent:hello-world capability (a fresh store has no such tuple → 403).

**Expected output**
sample.txt → model reads the text verbatim (the "Jack Sparrow" profile).
sample.jpeg → model describes the image (a red apple).
sample.pdf → model summarizes the PDF.

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist


sriaradhyula:
@jasonc3-sys - Thank you for the PR. 

Couple of things:
- Rebase PR and resolve conflicts
- Can you please add E2E playwright tests?
- Can you please add a screen recording or playwright screenshot of the changes?

cisco-erilutz:
I was able to paste an image in the UI and it was able to process the small text in it which was AMAZING, but the image thumbnail
Screenshot 2026-07-16 at 4 34 53 PM
was borked, and the image never showed up in the chat so the convo didn't make sense. this may require some more work since we'd need to store images now in the DB, but those images need to show up in the chat like they would a text message thread or slack thread

jasonc3-sys:
> @jasonc3-sys - Thank you for the PR.
> 
> Couple of things:
> 
> * Rebase PR and resolve conflicts
> * Can you please add E2E playwright tests?
> * Can you please add a screen recording or playwright screenshot of the changes?

Hi Sri, done!

cisco-erilutz:
While debugging a local repro of this feature (Slack image attachment → `invalid_image_format` from Azure OpenAI), I found the root cause was actually a missing `files:read` OAuth scope on the bot token — Slack silently returns an HTML login page (HTTP 200) instead of the file bytes when the token lacks that scope, and `file_ingest.py` forwards those bytes to the model as if they were the declared `mime_type`.

`file_ingest.py`'s own docstring/comment on `_download_one` already calls this out as a known risk ("Slack serves an HTML login page (HTTP 200) for missing/invalid auth, so callers should also sanity-check the returned content type / size upstream") but that sanity-check was never implemented. Worth adding defense-in-depth here: check the response `Content-Type` header (and/or sniff magic bytes) against the declared `mimetype` and skip+warn on mismatch, rather than silently forwarding bad data to the LLM and surfacing a confusing provider-side error.

Separately: while chasing this I found and fixed an unrelated local-dev issue in `docker-compose.dev.yaml` — the `dynamic-agents` service was leaking `OIDC_AUDIENCE` from the shared `.env` via `env_file`, which overrides `jwks_validate.py`'s default accepted-audience list and rejects Slack-bot OBO tokens (`aud=caipe-platform`) as `bearer_invalid`/401. I have an uncommitted fix for this locally (adds an explicit blank `OIDC_AUDIENCE=` override to the `dynamic-agents` environment block) — flagging in case it's worth folding into this PR since it blocks testing this exact feature locally via Slack.

kevkantes:
Multimodal chat attachments (files/images, capped at 10 files / 5MB each / 20MB per turn) are currently stored inline in DocumentDB alongside the rest of our chat data (messages, turns, auth cache). DocumentDB charges per read/write IO and storage is priced above S3, so as attachment volume grows with usage, we're scaling our most expensive, hardest-to-resize resource with data that doesn't need to live there. It also means attachment bytes only get backed up as part of a full DocumentDB snapshot rather than as independently managed objects.

I suggest we should move attachment bytes into S3, keeping only small metadata docs (s3Key, chatId, messageId, size, mimeType) in DocumentDB. Reuse the same bucket and SA/role that audit-service already uses for S3 (just a new `chat-attachments/` prefix), and extract audit-service's boto3 client setup into a small shared module so both features build their S3 client the same way.

cisco-erilutz (approved):

PR #2227, opened 2026-07-15
fix(slack-bot): register conversation before VictorOps on-call lookup

cisco-erilutz:
## Summary

- `_ping_victorops_oncall` generates a fresh `conversation_id` via `uuid.uuid4()` before querying the AI for who's on-call, so the lookup doesn't inherit the channel's (potentially very large) thread history.
- However, it never registers that id with the server via `create_conversation()` before streaming against it with `stream_chat()`. Every lookup's `stream_chat()` call fails with a 404 (`"Conversation not found"`), so every VictorOps escalation silently falls back to `"Could not determine on-call for team X. Please check VictorOps manually."` — regardless of whether the on-call lookup itself would have succeeded.
- Fix: call `sse_client.create_conversation()` first (no `idempotency_key`, so it's always a fresh conversation) and stream against the `conversation_id` it returns, instead of an ad hoc unregistered UUID.

Added regression tests covering both the successful lookup path (asserts `create_conversation` is called before `stream_chat`, and `stream_chat` uses the returned `conversation_id`) and the failure path.

## Test plan

- [x] Added `test_escalation_victorops_conversation.py` — verified it fails against the pre-fix code (mocked `stream_chat` 404 matches the observed production error) and passes against the fix
- [x] Full `slack_bot` test suite passes (526 tests)
- [ ] Configure a channel with VictorOps escalation enabled, trigger a "Get help" escalation, and confirm the on-call ping succeeds instead of falling back to the manual-check message

### DIFF
--- a/ai_platform_engineering/integrations/slack_bot/tests/test_escalation_victorops_conversation.py
+++ b/ai_platform_engineering/integrations/slack_bot/tests/test_escalation_victorops_conversation.py
@@ -1,0 +1,86 @@
+# Copyright 2025 CNOE Contributors
+# SPDX-License-Identifier: Apache-2.0
+"""Regression test: VictorOps on-call lookup must register its conversation.
+
+`_ping_victorops_oncall` streams a one-off "who is on-call" prompt through
+`sse_client.stream_chat()` using a conversation_id that is never registered
+with the server via `create_conversation()`. The server's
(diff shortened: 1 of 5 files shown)


cisco-erilutz:
Superseded by #2228 (same fix, rebranched to follow the prebuild/fix/ naming convention).

PR #2273, opened 2026-07-22
fix(workflow): do global check correctly

suwhang-cisco:
# Description

Team-visible workflows incorrectly reported global step agents as inaccessible because validation only checked explicit team grants and ignored the global user:* grant.
The fix checks global agent access first and treats it as accessible to every team, falling back to the explicit team grant check when needed. Regression tests cover both paths.

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

For chart changes, you can test pre-release versions before merging:
- **Base repo contributors:** Create a branch starting with `pre/` for automatic pre-release builds
- Pre-release charts are published to `ghcr.io/cnoe-io/pre-release-helm-charts`

## Checklist

### DIFF
--- a/ui/src/app/api/workflow-configs/check-agent-access/__tests__/route.test.ts
+++ b/ui/src/app/api/workflow-configs/check-agent-access/__tests__/route.test.ts
@@ -1,0 +1,98 @@
+/**
+ * @jest-environment node
+ */
+
+import { NextRequest } from "next/server";
+
+const mockCheckOpenFgaTuple = jest.fn();
+const mockGetCollection = jest.fn();
+
+jest.mock("@/lib/api-middleware", () => ({
+  withAuth: async (
+    _request: NextRequest,
+    handler: () => Promise<Response>,
+  ): Promise<Response> => handler(),
+}));
+
+jest.mock("@/lib/rbac/openfga", () => ({
+  checkOpenFgaTuple: (...args: unknown[]) => mockCheckOpenFgaTuple(...args),
+}));
+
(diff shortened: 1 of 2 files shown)


subbaksh (approved):

sriaradhyula (approved):
Correct fix — approving with two minor notes for follow-up.

**What looks good**

- Right semantics: `user:*` covers every authenticated user by definition, so checking it first and short-circuiting is the correct approach.
- Safe error handling: a transient OpenFGA failure in the global check falls through to the team-specific check rather than silently granting access.
- Tests assert call order (not just outcome), which prevents the logic from being accidentally reordered later.
- Minimal diff, nothing rearranged outside the scope of the fix.

**Two things worth a follow-up (neither blocks merging)**

1. **Redundant global lookups per team** — `teamHasAgentAccess` is called inside a `Promise.all` over `teams x agents`, so a globally-accessible agent triggers one `user:*` OpenFGA call per team instead of once. For a workflow shared with 5 teams that is 5 identical round-trips. Consider hoisting the global check out of `teamHasAgentAccess` and computing it once per `agentId` before the per-team fan-out.

2. **Missing negative test** — the two new tests cover both happy paths (global grant short-circuits; no global grant falls through to team check). A third case asserting that `gaps` is non-empty when both checks return `allowed: false` would complete the coverage triangle.

Neither blocks this — good fix, thanks!

PR #2288, opened 2026-07-23
fix(ui): add proactive session keepalive to prevent SSO idle timeout

sriaradhyula:
## Description

Users were being logged out mid-session when their Keycloak SSO session idle timer expired while the tab was open but idle. Root cause analysis:

- Keycloak `ssoSessionIdleTimeout` defaults to **8h**
- Access token lifespan is **4h** — the UI only calls `updateSession()` within 5 minutes of access token expiry
- A user idle >8h has their refresh token invalidated server-side
- The next token refresh returns `invalid_token` (not `invalid_grant`), which `auth-config.ts` correctly maps to `RefreshTokenExpired` — triggering immediate logout

**Fix:** add a 6h keepalive interval in `TokenExpiryGuard` that calls `updateSession({ forceRefresh: true })`. This triggers the existing `forceRefresh` path in `auth-config.ts` (line 612), performs a real OIDC token exchange with Keycloak, and resets the SSO idle timer before it expires. If a keepalive fails, the next cycle fires 6h later — well within the 8h window.

Evidence from production Keycloak logs: 139 `REFRESH_TOKEN_ERROR` events over 6 days, trending upward as user adoption grows. All with `reason="Token is not active"` and `userId=null` (session already gone server-side).

## Type of Change

- [x] Bugfix

## Checklist

### DIFF
--- a/ui/src/components/token-expiry-guard.tsx
+++ b/ui/src/components/token-expiry-guard.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-// assisted-by Codex Codex-sonnet-4-6
+// assisted-by claude code claude-sonnet-4-6
 
 import { Button } from "@/components/ui/button";
 import { formatTimeUntilExpiry,getTimeUntilExpiry,getWarningTimestamp,isTokenExpired } from "@/lib/auth-utils";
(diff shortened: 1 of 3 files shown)


ui/src/components/token-expiry-guard.tsx
```diff
@@ -1,6 +1,6 @@
 "use client";
 
-// assisted-by Codex Codex-sonnet-4-6
+// assisted-by claude code claude-sonnet-4-6
```
kevkantes: can we remove these slop comments?

kevkantes: Going forward as well, and anywhere else in this area

sriaradhyula: Removed — good call. Confirmed it's the only occurrence in this PR's diff.

PR #2293, opened 2026-07-24
fix(rag): recurse into sitemap index children instead of scraping them as pages

sriaradhyula:
## Summary
- `WorkerSpider.parse_sitemap()` extracted every `<loc>` URL with a flat regex regardless of whether the document was a `<sitemapindex>` (lists other sitemaps) or a `<urlset>` (lists real pages).
- Any site whose `sitemap.xml` is an index -- e.g. `outshift.cisco.com`, which splits into `sitemap-pages.xml`, `sitemap-blogs.xml`, `sitemap-events.xml` -- had those child sitemap URLs dispatched straight to `parse_page`. Those XML documents fetch fine (HTTP 200) but yield zero extractable content, so ingestion silently reported "N URLs found in sitemap, 0 scraped".
- `parse_sitemap` now detects a `<sitemapindex>` root and recurses into each child sitemap via itself instead of `parse_page`. `total_pages_to_crawl` / `urls_found_in_sitemap` now accumulate instead of overwrite, since an index fans out to multiple children each reaching that code path once.

## Test plan
- [x] Added `TestParseSitemapIndex` in `tests/webloader/test_spiders.py` using the real `outshift.cisco.com` sitemap structure as fixture data: one test asserts the index recurses into its 3 children via `parse_sitemap` (not `parse_page`); the other asserts a child `<urlset>` dispatches to `parse_page` and that totals accumulate correctly across multiple children.
- [x] `uv run pytest tests/webloader/test_spiders.py` -- 40 passed, 3 skipped (optional Playwright deps).
- [x] `uv run pytest` (full ingestors suite) -- same 10 pre-existing failures reproduce identically on a clean `main` checkout (unrelated dependency-version issues in `backstage`/`s3`/`slack`/`readthedocs`/`tls` tests, untouched by this change); no new failures introduced.

### DIFF
--- a/ai_platform_engineering/knowledge_bases/rag/ingestors/src/ingestors/webloader/loader/scrapy_worker.py
+++ b/ai_platform_engineering/knowledge_bases/rag/ingestors/src/ingestors/webloader/loader/scrapy_worker.py
@@ -299,7 +299,26 @@ class WorkerSpider(Spider):
 
     # Extract URLs from sitemap
     urls = re.findall(r"<loc>(.*?)</loc>", response.text)
-    self.urls_found_in_sitemap = len(urls)
+
+    # A sitemap *index* (root element <sitemapindex>) lists other sitemaps,
+    # not pages — its <loc> entries point to child sitemap files. Dispatching
+    # those straight to parse_page fetches XML and "successfully" gets a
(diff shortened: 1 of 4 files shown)


juliarvalenti (approved):
lgtm!

PR #2329, opened 2026-07-30
fix(slack-bot): apply execution_identity on retry/regenerate actions

cisco-erilutz:
## Summary

- `handle_mention` and `_route_to_agent` both resolve the channel route's `execution_identity` and call `apply_execution_identity()` (minting a service-account token when the route is configured to run as a service account) before binding the OBO token used for MCP calls.
- The "Retry" button handler (`handle_caipe_retry`) and the feedback modal's "regenerate with feedback" submission handler (`handle_feedback_modal_submission`) skipped this step entirely — they only bound whatever OBO token happened to already be on the request, i.e. the clicking human's own token.
- For channels routed to run as a service account, this meant retry/regenerate dispatched under the human's identity instead, silently dropping identity-gated MCP tools (e.g. one that authorizes per-request via OAuth) from the agent's tool list, while non-gated tools kept working — a confusing, hard-to-diagnose symptom.

## Fix

- Added `_resolve_agent_binding()` in `app.py`, mirroring the existing `_resolve_escalation()` pattern: check static YAML config first, then fall back to the DB-backed route resolver.
- Added a corresponding `agent_binding_for()` method on `SlackAgentRouteResolver` (refactored `escalation_for()` to share it).
- `handle_caipe_retry` and `handle_feedback_modal_submission` now resolve the route's `AgentBinding`, apply its `execution_identity` the same way the original dispatch does, and only then bind the OBO token — matching the working `handle_mention` / `_route_to_agent` paths.

## Test plan

- [x] Added `test_regen_execution_identity.py` covering both handlers under a `service_account` route (asserts the SA token is minted and overwrites the OBO token) and under the default `obo_user` route (asserts no minting occurs and the human's token is left alone).
- [x] Added `agent_binding_for()` coverage in `test_slack_agent_routes.py`.
- [x] Full existing suite passes: `uv run --project ai_platform_engineering/integrations/slack_bot pytest ai_platform_engineering/integrations/slack_bot/tests/ -m "not integration"` (564 passed).
- [x] `ruff check` passes on all changed files.

### DIFF
--- a/ai_platform_engineering/integrations/slack_bot/utils/slack_agent_routes.py
+++ b/ai_platform_engineering/integrations/slack_bot/utils/slack_agent_routes.py
@@ -291,6 +291,30 @@ class SlackAgentRouteResolver:
                 matches.append(binding)
         return matches
 
+    def agent_binding_for(
+        self,
+        *,
+        workspace_id: str,
+        channel_id: str,
+        agent_id: str,
+    ) -> AgentBinding | None:
+        """Return the full ``AgentBinding`` for a DB-backed channel/agent route.
+
+        Lets callers that only have a ``channel_id``/``agent_id`` pair (button
+        and modal-submission handlers, which have no live message event to
+        match against) resolve the same ``execution_identity`` used by the
+        original dispatch, so retry/regenerate actions run under the route's
+        configured identity instead of defaulting to the clicking user.
+        """
+
+        if not agent_id:
+            return None
+        for route in self._cached_routes(workspace_id, channel_id):
+            if route.get("agent_id") != agent_id:
+                continue
+            return _route_to_agent_binding(route)
+        return None
+
     def escalation_for(
         self,
         *,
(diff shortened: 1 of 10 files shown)


ai_platform_engineering/integrations/slack_bot/utils/slack_agent_routes.py
```diff
@@ -291,6 +291,30 @@ def match_routes(
...
+        and modal-submission handlers, which have no live message event to
+        match against) resolve the same ``execution_identity`` used by the
+        original dispatch, so retry/regenerate actions run under the route's
+        configured identity instead of defaulting to the clicking user.
+        """
+
+        if not agent_id:
+            return None
+        for route in self._cached_routes(workspace_id, channel_id):
+            if route.get("agent_id") != agent_id:
```
cisco-erilutz, line 313: i believe it's possible to have two routes with the same agent, e.g. one that's run as SA and one that's run as user. This could break on that right?

cisco-erilutz: Good catch — fixed in 8d02a3623.

Instead of trying to re-derive the binding from `agent_id` alone (which is genuinely ambiguous when a channel has both an obo_user and a service_account binding for the same agent), the original dispatch now persists the actual `execution_identity`/`escalation` it applied into the conversation's metadata (`dispatch_execution_identity`/`dispatch_escalation`). Retry/regenerate/escalation handlers read that back to recover the authoritative route config for the thread, falling back to the old agent_id-only lookup only for legacy threads that predate this fix.

One wrinkle: resolving the conversation itself needs *some* identity already bound, since `create_conversation`'s permission check runs even for an already-existing conversation (idempotency lookup happens after the check server-side) — but the authoritative identity is only recoverable from that same call's response. So retry/regenerate now resolve in two passes: apply a provisional (possibly ambiguous) guess to authorize the lookup, then correct to the authoritative identity from the metadata just returned, before dispatching.

Added `test_dispatch_metadata_resolution.py` plus new regression cases in `test_regen_execution_identity.py` covering both correction directions (obo_user→service_account and vice versa).

PR #2330, opened 2026-07-30
feat: shared global.storage.s3 block 

jasonc3-sys:
## Description

Two related SDPL-2012 follow-ups for the multimodal file-upload work, bundled here.

### 1. Resolve object storage from a shared `global.storage.s3` block (helm)

Today operators must set the same S3 bucket in two places — `audit-service.storage.s3.bucket` and `dynamic-agents.config.ATTACHMENT_S3_BUCKET`. This introduces a top-level `global.storage.s3` block (bucket / region / endpointUrl) so the operator sets the bucket **once**, and each subchart derives from it using distinct prefixes:

- `audit-service` writes audit logs under an `audit` prefix
- `dynamic-agents` writes attachment blobs under `attachments`

A subchart-local value always takes priority when set, so this is a backward-compatible default, not a breaking change. A new `dynamic-agents.attachmentConfig` helper resolves `ATTACHMENT_BACKEND: "auto"` to `s3` when a bucket is available and `local` otherwise, omitting orphan `ATTACHMENT_S3_*` keys when no bucket is configured.

Validated with `helm template` across four cases (global fallback, no-bucket→local, explicit override wins, mixed); `helm lint` clean.

### 2. Deterministic fallback for unreadable Slack attachments (slack-bot)

When a Slack user attaches a file the bot can't read — most commonly because the bot lacks the `files:read` scope, so Slack serves an HTML login page instead of the file bytes — the file is dropped at ingest. Previously the per-file notice was folded into the agent turn and the model paraphrased it, producing a differently-worded reply every time.

This splits the two audiences:
- **User** gets a fixed, verbatim message posted to the thread: *"I can't read files attached in Slack — please upload the file directly in the Forge UI, where file reading works."* — deterministic, never routed through the model.
- **Model** gets only a terse system note so it still answers any accompanying question without re-explaining the missing file.

Wired into all three routing paths: `@`-mention, ambient channel, and DM.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## Testing

- `helm template` (4 storage scenarios) + `helm lint` clean.
- `test_slack_file_share_routing.py`: fixed phrase posted verbatim, caption still reaches CAIPE, no fallback for readable files.
- Full `slack_bot` suite: 563 passed.


ai_platform_engineering/integrations/slack_bot/utils/file_ingest.py
```diff
@@ -50,6 +50,18 @@ class IngestResult(NamedTuple):
...
+
+# Fixed, deterministic message posted to the Slack thread when one or more
+# attached files could not be read. Posted verbatim (not routed through the
+# model, which would paraphrase it differently every time), so users always
+# see the same fallback. The per-file ``notices`` above still ride into the
+# turn as a terse note so the model doesn't answer as if no file were attached,
+# but the *user-facing* copy is this constant.
+SLACK_FILE_FALLBACK = (
+    "I can't read files attached in Slack — please upload the file directly in "
+    "the Forge UI, where file reading works."
```
cisco-erilutz: no forge references in the upstream repo, there should be an env variable like APP_NAME iirc

kevkantes (changes requested):
>  Deterministic fallback for unreadable Slack attachments (slack-bot)

> When a Slack user attaches a file the bot can't read — most commonly because the bot lacks the files:read scope, so Slack serves an HTML login page instead of the file bytes — the file is dropped at ingest. Previously the per-file notice was folded into the agent turn and the model paraphrased it, producing a differently-worded reply every time.

> This splits the two audiences:

> User gets a fixed, verbatim message posted to the thread: "I can't read files attached in Slack — please upload the file directly in the Forge UI, where file reading works." — deterministic, never routed through the model.
    
I don't think we need this. I would just have the model be aware it can't read it, and it can respond accordingly. What does this look like to the end user?

jasonc3-sys:
Removed the deterministic fallback entirely — the model now just sees that it couldn't read the attachment and responds naturally, per your suggestion. This PR is now S3 storage only.

kevkantes (approved):

PR #2366, opened 2026-08-11
fix(dynamic-agents): constrain curl request arguments

sriaradhyula:
# Description

Make the built-in curl tool use a defined request interface instead of forwarding unrestricted command-line arguments.

- Validate every request URL against the configured domain and address rules.
- Allow only the HTTP method, header, inline body, timeout, and response flags used by the tool.
- Reject local-file input/output options and unsupported schemes.
- Disable URL glob expansion and automatic redirects.
- Preserve inline JSON requests and write-method result warnings.

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

No chart changes.

## Checklist

## Validation

- 24 focused tests passed, including a real local curl subprocess and HTTP server
- Rejection cases verify no subprocess starts
- Ruff passed

### DIFF
--- a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/builtin_tools.py
+++ b/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/builtin_tools.py
@@ -27,6 +27,55 @@ logger = logging.getLogger(__name__)
 
 _REDIRECT_STATUS_CODES = {301, 302, 303, 307, 308}
 _MAX_FETCH_REDIRECTS = 10
+_CURL_BOOLEAN_OPTIONS = {
+    "-f",
+    "--fail",
+    "--fail-with-body",
+    "-G",
+    "--get",
+    "-i",
+    "--include",
+    "-I",
+    "--head",
+    "-s",
+    "--silent",
+    "-S",
+    "--show-error",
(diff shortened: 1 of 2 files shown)


subbaksh (approved):

ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/builtin_tools.py
```diff
@@ -92,6 +141,101 @@ def _validate_fetch_url(url: str, allowed_domains: str, allow_non_public_urls: b
...
 
+def _validate_curl_option_value(option: str, value: str) -> str | None:
+    """Reject curl option values that can make curl read local files."""
+    if option not in _CURL_FILE_CAPABLE_OPTIONS:
+        return None
+
+    stripped_value = value.lstrip()
+    if stripped_value.startswith("@") or (option == "--data-urlencode" and "@" in value):
+        return f"Option '{option}' cannot read request data from a local file"
+    return None
```
subbaksh, line 152: We should return a agent friendly error instead of None
like `This curl argument is forbidden for security reasons.`

PR #2371, opened 2026-08-11
fix(ui): stop failed conversation deletes from silently reappearing

### LINKED ISSUE #2167: Bug: Deleted conversations reappear in chat history
## Description

When attempting to delete certain conversations from the chat history, the deletions do not persist — the deleted conversations reappear after some time.

## Steps to Reproduce

1. Open the chat history in the UI
2. Select one or more conversations and delete them
3. Observe that the deleted conversations reappear later

## Expected Behavior

Deleted conversations should be permanently removed from the chat history.

## Actual Behavior

Some conversations cannot be deleted — they reappear in the history after deletion. The issue appears to affect specific conversations (not all).

## Additional Context

- Reported via internal service desk (OPENSD-2420)
- Priority: P3
- A screenshot was attached to the original report showing the affected conversations

rayl15:
# Description

Fixes #2167 — "Deleted conversations reappear in chat history".

## Root cause

Sidebar delete is optimistic: `deleteConversation` in `ui/src/store/chat-store.ts`
removes the row from the store first, then calls
`DELETE /api/chat/conversations/[id]`. Every server failure except 404 was
swallowed with a `console.error`, and `ui/src/components/layout/Sidebar.tsx`
unconditionally followed up with a `"<title>" moved to Archive` success toast.

The route gates on `conversation#can_delete`
(`requireConversationResourcePermission`, `ui/src/lib/rbac/conversation-implicit-authz.ts:99`),
so a conversation **shared with — but not owned by — the viewer** returns 403.
The sidebar renders the archive button for every row in the list, shared ones
included.

Result: the row disappeared locally, the server kept it, and it reappeared on
the next hydrate — which is exactly the reported symptom that only *specific*
conversations come back rather than all of them.

## Changes

**`ui/src/store/chat-store.ts`**
- Snapshot the removed conversation before the optimistic removal; on a
  non-404 server failure, restore it at its original index and rethrow so
  callers can report the failure.
- Only the auto-advance that this call performed is undone, so an active
  conversation the user switched to in the meantime is left alone.
- Decide "already gone" from the `APIClientError` HTTP status rather than
  substring-matching the error message.

**`ui/src/components/layout/Sidebar.tsx`**
- On failure, show an error toast (`Couldn't archive "<title>": <reason>`)
  instead of the success toast, and skip the post-archive navigation.

## Testing

New tests, all of which fail on the pre-fix code:

- `ui/src/store/__tests__/chat-store.test.ts`
  - restores the conversation in place and rethrows when the server returns 403
  - still treats a 404 *status* as an accepted delete even when the message
    doesn't say "not found"
- `ui/src/components/layout/__tests__/Sidebar.test.tsx`
  - reports the failure instead of claiming the conversation was archived
  - still confirms the archive when the server accepts it

```
npx jest src/store src/components/layout   →  9 suites, 312 tests passed
npx tsc --noEmit                            →  clean
npm run lint                                →  clean
```

## Note on scope

This makes the failure honest and non-destructive; it does not change who is
allowed to delete a shared conversation. If the intent is that a viewer should
be able to remove a shared conversation from *their own* list, that needs a
separate per-viewer dismissal concept on the backend — happy to open a
follow-up issue if maintainers want that.

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

### DIFF
--- a/ui/src/components/layout/Sidebar.tsx
+++ b/ui/src/components/layout/Sidebar.tsx
@@ -11,6 +11,7 @@ import { ScrollArea } from "@/components/ui/scroll-area";
 import { useToast } from "@/components/ui/toast";
 import { Tooltip,TooltipContent,TooltipProvider,TooltipTrigger } from "@/components/ui/tooltip";
 import { resolveUsableChatAgentId } from "@/lib/chat-agent-selection";
+import { getErrorMessage } from "@/lib/error-utils";
 import { getStorageMode } from "@/lib/storage-config";
 import { cn,formatDate,truncateText } from "@/lib/utils";
 import { useChatStore } from "@/store/chat-store";
@@ -715,8 +716,25 @@ export function Sidebar({ activeTab, collapsed, onCollapse, onUseCaseSaved }: Si
                                 console.log('[Sidebar] Created replacement conversation:', navigateToId);
                               }
 
-                              // Archive the conversation (updates store + server)
-                              await deleteConversation(conv.id);
+                              // Archive the conversation (updates store + server).
+                              // The store rolls the conversation back into the list
+                              // when the server rejects the delete (e.g. a shared
+                              // conversation the viewer does not own), so report the
+                              // failure instead of claiming success — otherwise the
+                              // row silently returns on the next reload.
+                              try {
+                                await deleteConversation(conv.id);
+                              } catch (error) {
+                                toast(
+                                  `Couldn't archive "${archivedTitle}": ${getErrorMessage(error, 'the server rejected the request')}`,
+                                  "error",
+                                  6000,
+                                );
+                                if (navigateToId) {
+                                  router.replace(`/chat/${navigateToId}`);
+                                }
+                                return;
+                              }
 
                               // Show toast
                               if (storageMode === 'mongodb') {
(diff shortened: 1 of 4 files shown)


rayl15:
Reviewers, a heads up: the red `test` check isn’t a test failure; it’s due to the fork-permission limitation.

Here’s a breakdown of the per-step outcomes for the job ([run](https://github.com/cnoe-io/ai-platform-engineering/actions/runs/31463071235/job/93690340351)):

- `Run Jest tests`: **success**. Test suites passed 547 out of 547 and tests passed 6860 out of 6860, achieving a coverage of 63.88% exceeding the 40% gate.
- `Fail if tests failed`: **skipped** since no tests failed.
- The only failing step is `Comment test results on PR`, which throws an error stating `Resource not accessible by integration`. This is because the workflow token can’t post a comment on a PR from a fork. Unfortunately, this issue can’t be resolved from the fork side.

All other steps, including `lint`, `playwright-rbac-regression`, both `build-and-push` arch jobs, `gitleaks`, `Filesystem scan`, and `DCO`, are green.

sriaradhyula (approved):

PR #2373, opened 2026-08-11
feat(dynamic-agents): integrate remote agents as tool calls via A2A protocol

### LINKED ISSUE #2013: feat(dynamic-agents): integrate remote agents as tool calls via A2A protocol
## Background

PR #1728 removed the legacy standalone A2A agent architecture (supervisor model, `BaseLangGraphAgent`, `BaseLangGraphAgentExecutor`, per-agent A2A servers). The netutils agent was one of those standalone agents.

This issue tracks the follow-on work: restore netutils as a **thin A2A server** and add a `create_remote_agent_tool` factory to dynamic agents so remote A2A agents can be invoked as LangChain tool calls — without reinstating the monolithic base classes.

## Architecture

```
User → BFF (Next.js) → dynamic-agents (LangGraph + AG-UI/SSE)
                              │
                              └─ tool call: RemoteAgentTool
                                        │
                                        └─ POST /  (A2A tasks/send, JSON-RPC)
                                                 │
                                          netutils-agent
                                          (thin FastAPI + A2A SDK)
                                                 │
                                          mcp/netutils (stdio)
```

Key design decisions:
- No A2A SDK needed on the **client** side — `tasks/send` is a plain JSON-RPC POST over `httpx`
- Agent card at `/.well-known/agent.json` auto-populates tool name + description for the LLM
- Bearer token is forwarded from the caller's JWT context (same pattern as MCP client)
- Netutils A2A server is a new, lightweight implementation — not a restore of `BaseLangGraphAgent`

## Part 1 — `create_remote_agent_tool` factory in dynamic agents

New file: `ai_platform_engineering/dynamic_agents/src/dynamic_agents/tools/remote_agent_tool.py`

```python
import uuid
import httpx
from langchain_core.tools import BaseTool
from pydantic import BaseModel, Field


class _RemoteAgentInput(BaseModel):
    message: str = Field(description="The message to send to the remote agent")


class RemoteAgentTool(BaseTool):
    """LangChain tool that delegates to an A2A-compatible remote agent."""

    name: str
    description: str
    a2a_url: str
    bearer_token: str | None = None
    timeout: int = 120

    args_schema: type[BaseModel] = _RemoteAgentInput

    def _run(self, message: str) -> str:
        raise NotImplementedError("use async")

    async def _arun(self, message: str) -> str:
        headers = {"Content-Type": "application/json"}
        if self.bearer_token:
            headers["Authorization"] = f"Bearer {self.bearer_token}"

        payload = {
            "jsonrpc": "2.0",
            "method": "tasks/send",
            "id": str(uuid.uuid4()),
            "params": {
                "id": str(uuid.uuid4()),
                "message": {
                    "role": "user",
                    "parts": [{"type": "text", "text": message}],
                },
            },
        }

        async with httpx.AsyncClient(timeout=self.timeout) as client:
            resp = await client.post(self.a2a_url, json=payload, headers=headers)
            resp.raise_for_status()
            data = resp.json()

        artifacts = data.get("result", {}).get("artifacts", [])
        texts = [
            part["text"]
            for artifact in artifacts
            for part in artifact.get("parts", [])
            if part.get("type") == "text"
        ]
        return "\n".join(texts) or str(data.get("result", {}))


def create_remote_agent_tool(
    *,
    a2a_url: str,
    name: str | None = None,
    description: str | None = None,
    bearer_token: str | None = None,
) -> RemoteAgentTool:
    """Fetch the A2A agent card and return a LangChain tool wrapping it."""
    resolved_name = name
    resolved_desc = description

    if not (resolved_name and resolved_desc):
        try:
            card_url = a2a_url.rstrip("/").rsplit("/", 1)[0] + "/.well-known/agent.json"
            with httpx.Client(timeout=10) as c:
                card = c.get(card_url).json()
            resolved_name = resolved_name or card["name"].replace(" ", "_").lower()
            resolved_desc = resolved_desc or card.get("description", f"Remote agent at {a2a_url}")
        except Exception:
            resolved_name = resolved_name or "remote_agent"
            resolved_desc = resolved_desc or f"Remote agent at {a2a_url}"

    return RemoteAgentTool(
        name=resolved_name,
        description=resolved_desc,
        a2a_url=a2a_url,
        bearer_token=bearer_token,
    )
```

Wire into dynamic agents tool assembly (wherever MCP tools are loaded):

```python
from dynamic_agents.tools.remote_agent_tool import create_remote_agent_tool
from dynamic_agents.auth.token_context import get_user_token

remote_agent_urls = os.getenv("REMOTE_AGENT_URLS", "").split(",")
for url in filter(None, remote_agent_urls):
    tools.append(create_remote_agent_tool(
        a2a_url=url.strip(),
        bearer_token=get_user_token(),
    ))
```

## Part 2 — netutils A2A server (thin rewrite)

The old netutils agent was ~100 lines of `BaseLangGraphAgent` subclass plus `BaseLangGraphAgentExecutor`. The new version is a standalone FastAPI app using the A2A SDK only for the server-side protocol, backed by `mcp/netutils` via stdio.

New structure inside `ai_platform_engineering/agents/netutils/`:

```
agent_netutils/
  __main__.py   # FastAPI app + A2A server startup
  agent.py      # LangGraph react agent wired to mcp/netutils
  executor.py   # A2A TaskExecutor: receives tasks/send, calls agent, returns artifacts
```

Minimal `executor.py` sketch:

```python
from a2a.server.agent_execution import AgentExecutor, RequestContext
from a2a.server.events import EventQueue
from a2a.utils import new_agent_text_artifact


class NetUtilsExecutor(AgentExecutor):
    def __init__(self, agent):
        self.agent = agent  # LangGraph react agent with netutils MCP tools

    async def execute(self, context: RequestContext, event_queue: EventQueue) -> None:
        query = context.get_user_input()
        result = await self.agent.ainvoke({"messages": [("user", query)]})
        final_text = result["messages"][-1].content
        await event_queue.enqueue_event(new_agent_text_artifact(final_text))

    async def cancel(self, context: RequestContext, event_queue: EventQueue) -> None:
        raise NotImplementedError
```

`agent.py` creates a standard LangGraph `create_react_agent` backed by `mcp/netutils` (already exists at `ai_platform_engineering/mcp/netutils`).

## Acceptance criteria

- [ ] `RemoteAgentTool` / `create_remote_agent_tool` implemented and unit tested
- [ ] `REMOTE_AGENT_URLS` env var wired into dynamic agents tool assembly
- [ ] netutils A2A server (`agent_netutils/`) reimplemented without `BaseLangGraphAgent`
- [ ] netutils A2A server added to `docker-compose.dev.yaml` with health check
- [ ] End-to-end: dynamic agents can invoke netutils via `tasks/send` and return the result to the LLM
- [ ] Agent card (`/.well-known/agent.json`) served by netutils with correct name + description

## Notes

- No A2A SDK needed on the client (dynamic agents) side — just `httpx` + LangChain `BaseTool`
- The A2A SDK is used only on the server (netutils) side for protocol compliance
- `a2a_client.py` in `webex_bot/` is an AG-UI SSE client, unrelated to this issue — keep as-is
- The old `base_langgraph_agent.py` (2500 lines) should NOT be restored; the new executor is ~30 lines

abhinavaditya811:
Fixes #2013.

## What changed
- `ai_platform_engineering/dynamic_agents/src/dynamic_agents/config.py`
- `ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py`
- `ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/remote_agent_tool.py`
- `ai_platform_engineering/dynamic_agents/tests/test_remote_agent_tool.py`

## Verification
The project's own test suite was run before and after this change; it introduces no new test failures or lint violations.

### DIFF
--- a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/config.py
+++ b/ai_platform_engineering/dynamic_agents/src/dynamic_agents/config.py
@@ -88,6 +88,11 @@ class Settings(BaseSettings):
     # When set, MCP HTTP/SSE clients use this base URL (e.g. http://agentgateway:4000/mcp/{server_id})
     agent_gateway_url: str | None = None
 
+    # Comma-separated JSON-RPC endpoints of remote A2A agents (env:
+    # REMOTE_AGENT_URLS). Each becomes a tool every agent can call to delegate
+    # to that remote agent — e.g. "http://netutils-agent:8000/".
+    remote_agent_urls: str = ""
+
     # CAIPE credential service API used when USE_IMPERSONATION_TOKENS=true.
     credential_api_url: str | None = None
     credential_service_audience: str = "caipe-credential-service"
--- a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py
+++ b/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py
@@ -1215,10 +1215,29 @@ class AgentRuntime:
         if not urls:
             return []
 
-        tools = [
-            await create_remote_agent_tool(a2a_url=url, bearer_token=self._auth_bearer)
-            for url in urls
-        ]
+        # Resolve every card at once. Awaiting them one at a time meant an
+        # unreachable endpoint burned the full card timeout before the next
+        # request even started, so startup delay grew linearly with the number of
+        # configured agents. Concurrently, the slowest agent sets the cost
+        # instead of the sum of all of them. Cards are cached by URL, so only the
+        # first build after a restart pays anything at all.
+        results = await asyncio.gather(
+            *(
+                create_remote_agent_tool(a2a_url=url, bearer_token=self._auth_bearer)
+                for url in urls
+            ),
+            return_exceptions=True,
+        )
+
+        # One bad endpoint must not cost the agent every other remote tool, which
+        # is what a bare gather would do by propagating the first exception.
+        tools = []
+        for url, result in zip(urls, results, strict=True):
+            if isinstance(result, BaseException):
+                logger.warning(f"Skipping remote agent {url}: {result}")
+                continue
+            tools.append(result)
+
         logger.info(
             f"Agent '{self.config.name}': added {len(tools)} remote agent tools: "
             f"{[t.name for t in tools]}"
--- a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py
+++ b/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py
@@ -87,6 +87,7 @@ from dynamic_agents.services.model_capabilities import (
     ModelCapabilities,
     get_model_capabilities,
 )
+from dynamic_agents.services.remote_agent_tool import create_remote_agent_tool
 from dynamic_agents.services.skills import build_skills_files, detect_missing_skills, load_skills
 
 if TYPE_CHECKING:
@@ -924,6 +925,11 @@ class AgentRuntime:
         if builtin_tools:
             tools = tools + builtin_tools
 
+        # 2b. Add remote A2A agents as delegation tools
+        remote_agent_tools = await self._build_remote_agent_tools()
+        if remote_agent_tools:
+            tools = tools + remote_agent_tools
+
         # 3. Wrap all tools with error handling
         #    Exceptions become LLM-visible "ERROR: ..." strings instead of crashing the agent loop.
         if tools:
@@ -1198,6 +1204,26 @@ class AgentRuntime:
             f"[agent] Agent '{self.config.name}' initialized in {init_duration:.2f}s: "
             f"tools={len(tools)}, subagents={len(subagents) if subagents else 0}"
         )
+
+    async def _build_remote_agent_tools(self) -> list:
+        """Build one delegation tool per remote A2A agent in ``REMOTE_AGENT_URLS``.
+
+        Returns:
+            List of LangChain tools, one per configured remote agent URL.
+        """
+        urls = [u.strip() for u in (self.settings.remote_agent_urls or "").split(",") if u.strip()]
+        if not urls:
+            return []
+
+        tools = [
+            await create_remote_agent_tool(a2a_url=url, bearer_token=self._auth_bearer)
+            for url in urls
+        ]
+        logger.info(
+            f"Agent '{self.config.name}': added {len(tools)} remote agent tools: "
+            f"{[t.name for t in tools]}"
+        )
+        return tools
 
     def _build_builtin_tools(
         self,
--- a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/remote_agent_tool.py
+++ b/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/remote_agent_tool.py
@@ -1,0 +1,193 @@
+"""Remote A2A agent tools for Dynamic Agents.
+
+Wraps an A2A-compatible remote agent as a plain LangChain tool so the
+LangGraph agent can delegate to it with a normal tool call:
+
+    dynamic-agents ──tool call──> POST {a2a_url}  (JSON-RPC ``tasks/send``)
+
+No A2A SDK is needed on this side — ``tasks/send`` is a JSON-RPC POST over
+``httpx``. The agent card at ``/.well-known/agent.json`` supplies the tool
+name and description the LLM sees, so a remote agent describes itself.
+"""
+
+from __future__ import annotations
+
+import logging
+import re
+import uuid
+from typing import Any
+from urllib.parse import urljoin, urlparse
+
+import httpx
+from langchain_core.tools import BaseTool
+from pydantic import BaseModel, Field
+
+from dynamic_agents.auth.token_context import current_user_token
+
+logger = logging.getLogger(__name__)
+
+AGENT_CARD_PATH = "/.well-known/agent.json"
+AGENT_CARD_TIMEOUT_SECONDS = 10
+
+_UNSAFE_NAME_CHARS = re.compile(r"[^a-zA-Z0-9_-]+")
+
+
+def _sanitize_tool_name(raw: str) -> str:
+    """Reduce an agent card name to a tool name the LLM APIs accept."""
+    return _UNSAFE_NAME_CHARS.sub("_", raw).strip("_").lower()
+
+
+def _part_text(part: Any) -> str | None:
+    """Return the text of an A2A message/artifact part, if it is a text part.
+
+    A2A spec revisions tag parts with either ``type`` or ``kind``; accept both.
+    """
+    if not isinstance(part, dict):
+        return None
+    if part.get("type", part.get("kind")) != "text":
+        return None
+    text = part.get("text")
+    return text if isinstance(text, str) else None
+
+
+def _extract_result_text(data: dict[str, Any]) -> str:
+    """Flatten an A2A ``tasks/send`` JSON-RPC response into text.
+
+    Raises:
+        RuntimeError: if the remote agent returned a JSON-RPC error.
+    """
+    error = data.get("error")
+    if error:
+        message = error.get("message", error) if isinstance(error, dict) else error
+        raise RuntimeError(f"Remote agent returned an error: {message}")
+
+    result = data.get("result") or {}
+
+    texts = [
+        text
+        for artifact in result.get("artifacts") or []
+        if isinstance(artifact, dict)
+        for part in artifact.get("parts") or []
+        if (text := _part_text(part))
+    ]
+
+    # Agents that answer without producing an artifact put the reply on the
+    # task status message instead.
+    if not texts:
+        status_message = (result.get("status") or {}).get("message") or {}
+        texts = [
+            text for part in status_message.get("parts") or [] if (text := _part_text(part))
+        ]
+
+    return "\n".join(texts) or str(result)
+
+
+class _RemoteAgentInput(BaseModel):
+    message: str = Field(description="The message to send to the remote agent")
+
+
+class RemoteAgentTool(BaseTool):
+    """LangChain tool that delegates to an A2A-compatible remote agent."""
+
+    name: str
+    description: str
+    a2a_url: str
+    bearer_token: str | None = None
+    timeout: int = 120
+
+    args_schema: type[BaseModel] = _RemoteAgentInput
+
+    def _run(self, message: str) -> str:
+        raise NotImplementedError("RemoteAgentTool is async-only; use ainvoke()")
+
+    async def _arun(self, message: str) -> str:
+        # Prefer the token bound to the current request over the one captured
+        # when the tool was built: agent runtimes are cached and replayed
+        # across requests, so a captured token goes stale. Same per-request
+        # forwarding contract as the MCP httpx client factory.
+        token = current_user_token.get() or self.bearer_token
+
+        headers = {"Content-Type": "application/json"}
+        if token:
+            headers["Authorization"] = f"Bearer {token}"
+
+        payload = {
+            "jsonrpc": "2.0",
+            "method": "tasks/send",
+            "id": str(uuid.uuid4()),
+            "params": {
+                "id": str(uuid.uuid4()),
+                "message": {
+                    "role": "user",
+                    "parts": [{"type": "text", "text": message}],
+                },
+            },
+        }
+
+        async with httpx.AsyncClient(timeout=self.timeout) as client:
+            resp = await client.post(self.a2a_url, json=payload, headers=headers)
+            resp.raise_for_status()
+            data = resp.json()
+
+        return _extract_result_text(data)
+
+
+async def _fetch_agent_card(a2a_url: str) -> dict[str, Any] | None:
+    """Fetch ``/.well-known/agent.json`` for a remote agent, or None if unavailable.
+
+    A remote agent that is still starting up must not fail agent
+    initialization, so every failure degrades to ``None``.
+    """
+    card_url = urljoin(a2a_url, AGENT_CARD_PATH)
+    try:
+        async with httpx.AsyncClient(timeout=AGENT_CARD_TIMEOUT_SECONDS) as client:
+            resp = await client.get(card_url)
+            resp.raise_for_status()
+            card = resp.json()
+    except (httpx.HTTPError, ValueError) as e:
+        logger.warning(f"Could not fetch A2A agent card from {card_url}: {e}")
+        return None
+    return card if isinstance(card, dict) else None
+
+
+def _fallback_name(a2a_url: str) -> str:
+    """Name a remote agent after its host when no agent card is available.
+
+    Deriving from the URL keeps two unreachable agents from colliding on a
+    single generic tool name, which would break the LLM tool binding.
+    """
+    return _sanitize_tool_name(urlparse(a2a_url).hostname or "") or "remote_agent"
+
+
+async def create_remote_agent_tool(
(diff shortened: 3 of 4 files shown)


ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/remote_agent_tool.py
```diff
@@ -0,0 +1,193 @@
...
+        # forwarding contract as the MCP httpx client factory.
+        token = current_user_token.get() or self.bearer_token
+
+        headers = {"Content-Type": "application/json"}
+        if token:
+            headers["Authorization"] = f"Bearer {token}"
+
+        payload = {
+            "jsonrpc": "2.0",
+            "method": "tasks/send",
```
sriaradhyula, line 147: **[P1] Implement a supported A2A wire contract.** `tasks/send` and this legacy payload are not accepted by the intended A2A SDK. A2A 0.3 uses `message/send`, while current A2A uses `SendMessage`; both require a message identifier and differ in response shape. Please update request construction and response extraction end to end, and add a contract test against an actual A2A SDK server. See the [official specification](https://github.com/a2aproject/A2A/blob/main/docs/specification.md).

ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/remote_agent_tool.py
```diff
@@ -0,0 +1,193 @@
...
+
+import httpx
+from langchain_core.tools import BaseTool
+from pydantic import BaseModel, Field
+
+from dynamic_agents.auth.token_context import current_user_token
+
+logger = logging.getLogger(__name__)
+
+AGENT_CARD_PATH = "/.well-known/agent.json"
```
sriaradhyula: **[P1] Use the standard Agent Card path.** A2A publishes the public card at `/.well-known/agent-card.json`. Using `agent.json` makes discovery fail against conforming servers and silently replaces the advertised name and description with fallback values.

ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py
```diff
@@ -1199,6 +1205,26 @@ async def initialize(self) -> None:
...
+        """Build one delegation tool per remote A2A agent in ``REMOTE_AGENT_URLS``.
+
+        Returns:
+            List of LangChain tools, one per configured remote agent URL.
+        """
+        urls = [u.strip() for u in (self.settings.remote_agent_urls or "").split(",") if u.strip()]
+        if not urls:
+            return []
+
+        tools = [
```
sriaradhyula: **[P2] Avoid serial discovery timeouts.** Each card request is awaited before starting the next one. An unavailable endpoint can consume the full 10-second timeout, producing linear startup delay across configured agents. Please resolve cards concurrently with per-agent fallback, or cache/lazily fetch them.

sriaradhyula (commented):
Thanks for taking this on. I found three blocking concerns in the current patch:

- The client implements an obsolete A2A wire contract; see the inline comments.
- This PR closes #2013 but only implements the client-side tool. The issue also requires the thin netutils A2A server, Docker Compose service/health check, and end-to-end verification. Please either complete those acceptance criteria or stop closing #2013 and track the remaining work explicitly.
- The sole commit is missing its DCO `Signed-off-by` trailer, and the DCO check is currently `ACTION_REQUIRED`.

Validation performed: the 11 new tests pass and Ruff passes. The full dynamic-agents suite produced 379 passed, 3 skipped, and one unrelated failure that also reproduces on the parent commit.

ai_platform_engineering/dynamic_agents/src/dynamic_agents/config.py
```diff
@@ -88,6 +88,11 @@ class Settings(BaseSettings):
     # When set, MCP HTTP/SSE clients use this base URL (e.g. http://agentgateway:4000/mcp/{server_id})
     agent_gateway_url: str | None = None
 
+    # Comma-separated JSON-RPC endpoints of remote A2A agents (env:
+    # REMOTE_AGENT_URLS). Each becomes a tool every agent can call to delegate
+    # to that remote agent — e.g. "http://netutils-agent:8000/".
+    remote_agent_urls: str = ""
```
sriaradhyula, line 94: **[P1] Make remote-agent access configurable per dynamic agent.** `REMOTE_AGENT_URLS` is process-wide, and `_build_remote_agent_tools()` consequently gives every dynamic agent every configured A2A tool. The Agent Builder only persists MCP `allowed_tools`, so users cannot discover, select, remove, clone, or apply approval/RBAC rules to A2A agents. Please introduce an admin-managed A2A-agent registry plus per-agent references (for example `allowed_remote_agents`), expose list/probe endpoints, add an A2A section to the Tools step, persist the selection through the TypeScript/Python models and create/update routes, and build only the selected tools at runtime. Avoid storing arbitrary endpoint URLs directly in user-editable agent documents.

sriaradhyula (commented):
One additional product/authorization gap surfaced while tracing the Agent Builder flow: A2A agents need to be first-class, per-agent tool selections rather than a process-wide environment list.

ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/remote_agent_tool.py
```diff
@@ -0,0 +1,193 @@
+"""Remote A2A agent tools for Dynamic Agents.
+
+Wraps an A2A-compatible remote agent as a plain LangChain tool so the
+LangGraph agent can delegate to it with a normal tool call:
+
+    dynamic-agents ──tool call──> POST {a2a_url}  (JSON-RPC ``tasks/send``)
+
+No A2A SDK is needed on this side — ``tasks/send`` is a JSON-RPC POST over
```
sriaradhyula: **Suggestion: use the official `a2a-sdk` client here.** The hand-written client has already drifted from the protocol in its card path, method name, request identifiers, and response envelope. The official Python SDK supports both A2A 1.0 and 0.3 compatibility and provides `create_client()` / `ClientFactory.create_from_url()` for card discovery and transport negotiation. Pin the same SDK version for the dynamic-agents client and netutils server, use an interceptor (or supported HTTP configuration) for the request-scoped bearer token, and keep `RemoteAgentTool` as the thin LangChain adapter around SDK `send_message()`. The tradeoff is one dependency, but it removes protocol/version logic from this module and gives us supported handling for task, direct-message, streaming, and error responses. SDK docs: https://a2a-protocol.org/latest/sdk/python/api/a2a.client.client_factory.html

sriaradhyula (commented):
A concrete implementation recommendation for the protocol concerns: use the official A2A Python client SDK rather than maintaining a custom JSON-RPC client.

ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py
abhinavaditya811: Fixed. _build_remote_agent_tools now resolves cards with asyncio.gather, so the slowest agent sets the cost rather than the sum: five unreachable endpoints cost 10s instead of 50s. Uses return_exceptions=True with per-URL logging, since a bare gather would cost the agent every other remote tool over one bad endpoint.

ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/remote_agent_tool.py
abhinavaditya811: Fixed. The path is now /.well-known/agent-card.json, with /.well-known/agent.json kept as a fallback; it is tried only on a 404, so pre-0.3.0 servers still resolve. Any other status stops immediately rather than retrying a second path.

ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/remote_agent_tool.py
abhinavaditya811: This and your a2a-sdk suggestion are the same change, so I would rather not fix tasks/send by hand and then throw it away. Can you confirm the SDK as a client dependency?

If yes: pin one version across the dynamic-agents client and the netutils server, use ClientFactory.create_from_url() for discovery and negotiation, move bearer forwarding to a ClientCallInterceptor, and keep RemoteAgentTool as a thin adapter over send_message().

If no: I will implement SendMessage only and let 0.3 servers fail loudly rather than maintain two protocols by hand. Worth flagging why that matters: message/send does not appear in the current spec at all, only in the SDK's compat/v0_3 layer, so hand-rolling 0.3 support means implementing a method name the spec no longer documents.

Either way, your contract test makes a2a-sdk a test dependency of dynamic-agents, so the question is really whether we also take it at runtime.

Also confirming the target: pin 1.x on both sides and let the compat layer handle any third-party 0.3 server, or do you need netutils itself to advertise 0.3?

ai_platform_engineering/dynamic_agents/src/dynamic_agents/config.py
abhinavaditya811: Agreed on the substance, but I would like this as a follow-up issue rather than in this PR.

It is not in #2013, whose acceptance criterion is "REMOTE_AGENT_URLS env var wired into dynamic agents tool assembly". What you describe needs a registry schema, list and probe endpoints, a Tools step section, TS and Python models, and create/update routes. For scale, allowed_tools alone spans 81 files here including the RBAC e2e specs. That is mostly registry, RBAC and frontend work rather than A2A protocol work.

On the security point, which should not wait: nothing stores endpoint URLs in user-editable agent documents today. REMOTE_AGENT_URLS is deployment config, so it is admin controlled and not user reachable. Happy to make "no user-supplied endpoint URLs" an explicit constraint on the follow-up.

Cards are already cached by URL rather than by agent, so per-agent selection will only decide which tools get built.

Shall I open that issue and link it from #2013?

ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/remote_agent_tool.py
sriaradhyula: @abhinavaditya811 

> If yes: pin one version across the dynamic-agents client and the netutils server, use ClientFactory.create_from_url() for discovery and negotiation, move bearer forwarding to a ClientCallInterceptor, and keep RemoteAgentTool as a thin adapter over send_message().

Let's use the dependency and pin a specific A2A SDK version.

ai_platform_engineering/dynamic_agents/src/dynamic_agents/config.py
sriaradhyula: @abhinavaditya811 - Yes, please track it as new issue

sriaradhyula: @abhinavaditya811 Thank you for the thoughtful response and for the work you have already put into this PR. I reconsidered the product flow after tracing how remote agents need to be configured and selected in Agent Builder. My earlier reply suggested moving the per-agent selection work to a follow-up issue; I am changing that direction. I would like us to keep the work in this PR and make it the complete end-to-end implementation of #2013.

I also want you to remain the author of the PR and receive credit for the contribution. I am happy for the implementation to be organized into several logical commits within this same PR so the individual pieces remain reviewable.

### A2A protocol and client

- Replace the hand-written JSON-RPC client with the official Python `a2a-sdk`.
- Pin one exact SDK version and use the same version for both the dynamic-agents client and the netutils A2A server.
- Use the SDK for Agent Card discovery, transport negotiation, request construction, and response handling.
- Keep `RemoteAgentTool` as a thin LangChain adapter around the SDK's message-sending API.
- Forward the caller's bearer token at invocation time through the SDK's supported interceptor or per-call context. The runtime is cached, so a token captured while the tool is constructed must not be reused for later callers.
- Continue isolating failures so one unavailable remote agent does not prevent the other selected agents from loading.
- Add a contract test against an actual SDK-backed A2A server rather than validating the protocol only with a custom HTTP test handler.

### Remote-agent registry and administration UI

Please add an admin-managed `remote_agents` registry. A registry entry should hold a stable ID, endpoint URL, enabled state, and Agent Card metadata such as the name and description. Endpoint URLs should be managed in this registry, not copied into user-editable dynamic-agent documents.

Add the BFF/API and UI needed for an administrator to:

- Add a remote A2A endpoint.
- Probe the endpoint and retrieve its Agent Card.
- Review the discovered name, description, and connection status.
- Update, enable/disable, or remove the registry entry.
- Receive a clear error when discovery or validation fails.

Please validate URLs server-side and do not allow ordinary Agent Builder users to introduce arbitrary endpoints.

### Per-agent selection in Agent Builder

Add an `allowed_remote_agents: string[]` field to the TypeScript and Python dynamic-agent models. It should contain stable registry IDs.

In the Agent Builder Tools step, add a “Remote A2A Agents” section that lists the available registered agents and allows the user to select which ones this dynamic agent may invoke. Please carry that selection through all relevant create, edit, update, clone, import, and seed paths.

At runtime, resolve only the selected registry IDs and build only those tools. A remote agent configured by an administrator should not automatically become available to every dynamic agent.

`REMOTE_AGENT_URLS` may remain as an optional deployment/bootstrap mechanism, but I would treat it as a way to seed registry entries rather than as the runtime authorization source.

### Complete the remaining #2013 scope

Because this PR currently says it closes #2013, please also include the remaining acceptance criteria:

- The thin netutils A2A server, without restoring the removed legacy base-agent architecture.
- The Docker Compose service and health check.
- End-to-end verification that a dynamic agent can select and invoke netutils and return its result to the LLM.
- Documentation for configuring and selecting remote agents.

The branch is now substantially behind `main`, so please rebase before implementing the UI/API portion. That should reduce the chance of building against Agent Builder and authorization paths that have since changed.

Finally, the DCO check is still failing because the author identity on the two commits does not match their `Signed-off-by` trailers. Please amend those commits yourself using the human identity you intend to certify, then force-push the rebased branch. I do not want to rewrite your authorship or replace your contribution while addressing the DCO issue.

Please let me know if any part of this expanded scope is unclear or if you want to propose a different internal breakdown. The important outcome for me is that this remains your PR while delivering a supported A2A implementation that administrators can register and users can select safely per dynamic agent.

abhinavaditya811: @sriaradhyula A review came through on my notifications with updated direction for this PR, but I don't see it on the PR now, so I want to confirm before I start.

What I took from it:

- Keep the expanded scope in this PR using logically separated commits, rather than deferring the registry work to a new issue
- Admin-managed remote_agents registry storing endpoint URLs and discovered Agent Card metadata, with add/probe/update/remove API and UI
- allowed_remote_agents: string[] on each dynamic agent holding registry IDs, not URLs
- A "Remote A2A Agents" section in the Agent Builder Tools step, persisted through create, update and clone
- Runtime loading of only selected agents
- Pinned official A2A SDK with a contract test against an SDK server, plus request-scoped bearer forwarding and failure isolation
- REMOTE_AGENT_URLS demoted to an optional deployment seed
- The remaining #2013 criteria: thin netutils A2A server, Compose service and health check, end-to-end verification
- Rebase onto main first, and amend the existing commits so the DCO trailers match my commit-author identity

Two questions before I begin:

1. Does that still stand? I ask because #2013 is still scoped to wiring the env var into tool assembly and is still labeled good first issue, while the registry, admin API and Agent Builder work is a substantially larger piece of platform work.
2. If it does, should the registry and admin UI half live here, or under one of the new v2 epics such as #2553 or #2564? Happy either way, I just want the scope settled before I build it.

In the meantime I'm proceeding with the parts that are needed under any version of this: the a2a-sdk client rewrite you confirmed on Aug 13, and the DCO fix.

PR #2397, opened 2026-08-18
feat(feedback): add provider-neutral issue routing

sriaradhyula:
# Description

Add direct, provider-neutral issue routing for the existing Report a Problem flow.

- routes reports to the explicitly configured GitHub or Jira provider
- adds authenticated BFF endpoints for GitHub and Jira issue creation
- applies the existing `feedback:submit` RBAC policy to both ticket endpoints
- records created tickets in the unified feedback collection
- supports optional GitHub screenshot hosting and Jira attachments
- removes the redundant thumbs-down submit-and-report action while preserving pending feedback
- documents neutral configuration examples and excludes downstream product-specific routing

This consolidates and sanitizes the reusable portions of cisco-eti/ai-platform-engineering-mirror#232, #237, #250, and #272.

Validation:

- 379 focused Jest tests passed
- ESLint passed for all changed TypeScript files

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

Not applicable.

## Checklist


artdroz (changes requested):
## Request changes

Independently rechecked the three Jira findings against current head `c23567810304dac988ea841e60784514d1a7017a` (unchanged from the spot-check). These require fixes before approval.

### 1. P1 — The Report a Problem dialog never supplies the area required by the Jira endpoint

**Code:** [ui/src/app/api/tickets/jira/route.ts:78–80](https://github.com/caipe-io/ai-platform-engineering/blob/c23567810304dac988ea841e60784514d1a7017a/ui/src/app/api/tickets/jira/route.ts#L78-L80); caller [ReportProblemDialog.tsx:207–225](https://github.com/caipe-io/ai-platform-engineering/blob/c23567810304dac988ea841e60784514d1a7017a/ui/src/components/ticket/ReportProblemDialog.tsx#L207-L225); serialization [ticket-client.ts:186–198](https://github.com/caipe-io/ai-platform-engineering/blob/c23567810304dac988ea841e60784514d1a7017a/ui/src/lib/ticket-client.ts#L186-L198).

**Scenario and impact:** Configure valid Jira credentials and a project, open Report a Problem from the header or chat feedback, and submit. The dialog supplies no top-level `area`, the client serializes its undefined value away, and the endpoint returns HTTP 400 (`area is required`) before calling Jira. This affects every submission through this dialog when Jira is selected and the configuration/authentication checks pass; it is independent of the project issue-type scheme. Retrying cannot resolve it.

**Required correction:** Align the dialog/client/server contract: collect and pass a valid area, derive an appropriate area, or make it optional with a meaningful server-side default. Cover both header and chat-feedback entry points.

**Validation expectation:** Exercise the real dialog-to-client request shape against the route validation with Jira selected, for both entry points. A valid report must reach issue creation and show success rather than `missing_area`. The current isolated route tests provide an area manually, and the dialog tests mock ticket creation, so neither catches the contract mismatch.

### 2. P2 — Explicit `JIRA_TICKET_ENABLED=false` is overridden by credential presence

**Code:** [ui/src/lib/config.ts:432–444](https://github.com/caipe-io/ai-platform-engineering/blob/c23567810304dac988ea841e60784514d1a7017a/ui/src/lib/config.ts#L432-L444); dispatch [ticket-client.ts:227–234](https://github.com/caipe-io/ai-platform-engineering/blob/c23567810304dac988ea841e60784514d1a7017a/ui/src/lib/ticket-client.ts#L227-L234); endpoint configuration checks [jira/route.ts:35–57](https://github.com/caipe-io/ai-platform-engineering/blob/c23567810304dac988ea841e60784514d1a7017a/ui/src/app/api/tickets/jira/route.ts#L35-L57).

**Scenario and impact:** Keep `JIRA_BASE_URL`, `JIRA_EMAIL`, and a supported Jira token in the deployment, explicitly set `JIRA_TICKET_ENABLED=false`, and enable/configure GitHub. The OR expression still sets `jiraTicketEnabled=true`; provider selection gives Jira precedence and directs reports to Jira instead of the explicitly enabled GitHub destination. Currently those dialog requests encounter finding 1; fixing that mismatch alone would still leave routing to the wrong provider. The Jira endpoint also checks credentials/project but does not enforce the enabled flag.

**Required correction:** Honor explicit false before any credential-based auto-detection. If auto-detection is retained, apply it only when the enablement setting is absent. Enforce the resolved enablement policy at the Jira endpoint as well, so disabling the feature is respected by direct requests.

**Validation expectation:** Add a configuration case with complete Jira credentials, explicit false, and GitHub enabled: Jira must be disabled and GitHub selected. Also cover false with no alternative provider, true, and absent-setting behavior. A direct Jira POST while disabled must not invoke Jira issue creation.

### 3. P2 — Bug mapping assumes a service-management issue type exists in every target project

**Code:** [ui/src/lib/jira-ticket.ts:72–80](https://github.com/caipe-io/ai-platform-engineering/blob/c23567810304dac988ea841e60784514d1a7017a/ui/src/lib/jira-ticket.ts#L72-L80), used in [the creation payload at lines 93–100](https://github.com/caipe-io/ai-platform-engineering/blob/c23567810304dac988ea841e60784514d1a7017a/ui/src/lib/jira-ticket.ts#L93-L100).

**Scenario and impact:** Send a valid Jira report containing an area and `issueType: "Bug"` to a Jira Software project whose issue-type scheme includes `Bug` but not `[System] Problem`. The helper always sends `[System] Problem`; Jira rejects the issue type and no ticket is created. This is **deployment- and input-dependent**, not a universal Jira failure: projects exposing `[System] Problem` can accept this mapping, and the current dialog does not supply `issueType`. The `Task` branch is a default for other inputs, not a retry/fallback when Jira rejects the Bug mapping.

**Required correction:** Make the mapping compatible with the configured project, for example through configurable issue-type IDs/names or project metadata resolution. Preserve service-management support without hard-coding its type for all projects; document the configuration and do not assume `Task` exists in every scheme.

**Validation expectation:** Cover both a software-project mapping (`Bug`) and a service-management mapping (`[System] Problem`), including the exact outgoing payload. Verify that an unavailable configured type produces an actionable configuration error or a deliberately supported fallback. The current mocked test explicitly expects the hard-coded mapping and does not establish project compatibility.

### Review scope

All three prior findings remain, with the third explicitly qualified as deployment-dependent. No findings removed. This recheck used source inspection; no live Jira requests or local test runs were performed. The earlier CI inspection found no failed checks at this SHA, but the passing tests do not cover the integration/configuration cases above.

Issue #2392, opened 2026-08-17
Cmd+K search


kevkantes:
## Summary

Add a global `Cmd+K` / `Ctrl+K` command palette so people can quickly navigate the application and find authorized resources.

## Problem

As navigation grows, users must remember where destinations live and may need several clicks to reach them. The application needs a keyboard-first way to discover both fixed destinations and resources.

## Proposed outcome

Provide a single command palette that:

- Opens with `Cmd+K` on macOS and `Ctrl+K` on other platforms.
- Shows static navigation destinations and quick actions in the first result group.
- Searches user-accessible resources by name in a separate result group.
- Respects existing RBAC rules: resources the user cannot access must not appear.
- Supports keyboard selection and activation, including arrow keys and Enter.

## Acceptance criteria

- [ ] `Cmd+K` / `Ctrl+K` opens the command palette from the application UI.
- [ ] A user can find a static destination by name and open it without using a pointer.
- [ ] A user can use arrow keys and Enter to select and activate a result.
- [ ] Results distinguish static navigation/quick actions from matching resources.
- [ ] Search returns matching, authorized resources by name across supported resource types.
- [ ] Users never see resources they are not authorized to access.
- [ ] The implementation identifies the initial supported resource types and documents any intentionally deferred types.

## Constraints and considerations

- Keep the first iteration focused on command-palette navigation and resource discovery.
- Evaluate LLM-backed question answering as a follow-up capability, not a prerequisite for this feature.
- Avoid introducing a dedicated search service unless measured requirements justify it.

## Source

- Public example: [Perplexica documentation](https://itzcrazykns-perplexica.mintlify.app/introduction)

PR #2407, opened 2026-08-19
feat(ui): improve global navigation and breadcrumbs

sriaradhyula:
## Summary

- Move page breadcrumbs into the shared application header so pages start from one consistent top edge.
- Restore restrained brand and active-navigation transitions.
- Keep Admin navigation mounted across route changes to prevent collapse/reopen flashing.
- Fix collapsed navigation flyouts so the full submenu stays usable inside the viewport.
- Preserve feature gates and authorization-aware navigation behavior.

## Scope

The global command palette requested in #2392 has been moved to the independent PR #2463. This PR now contains only the breadcrumb, sidebar stability, transition, and collapsed-navigation UX changes.

## Test plan

- [x] `npm test -- --runInBand --runTestsByPath src/components/layout/__tests__/AppHeader.test.tsx` — 56 tests passed
- [x] Targeted ESLint for the affected application-navigation files
- [x] `npx tsc --noEmit --pretty false`
- [x] `npm run build`
- [x] Existing live navigation regressions verified on caipe-oss


cisco-erilutz (dismissed):

ui/src/components/layout/ApplicationNavigation.tsx
```diff
@@ -220,6 +238,8 @@ function ApplicationNavigationContents({
       label: "Admin",
       icon: Shield,
       disabled: storageMode !== "mongodb",
+      disabledReason:
```
kevkantes, line 248: can you apply this to the Knowledge Bases no-access case from Valentina's feedback as well? right now this only disables Admin when storage isn't Mongo. KnowledgeSidebar still renders the full subnav and then the banner, and there is no Contact admin CTA there. the ask was to disable that destination and show only the disclaimer + CTA.

kevkantes (changes requested):
the animation changes look good, but the unavailable Knowledge Base flow from the feedback is still not implemented.

kevkantes (dismissed):
lgtm now, this addresses the Knowledge Base flow I called out. approved.

cisco-erilutz (dismissed):

cisco-erilutz (approved):

PR #2620, opened 2026-09-01
docs: prune 0.4.18 from published versions

sriaradhyula:
## Summary
- Removes `0.4.18` from `docs/published-versions.json`, leaving `["0.6.0", "0.5.69"]`.
- `0.4.x` is outside the retention policy's `MAX_MINORS=3` window now that `0.6.x`/`0.5.x` are the two newest active minor series. `update-published-versions.js` would drop it automatically once a final (non-rc) `1.0.0` tag runs through the docs-snapshot workflow, but pre-release tags are excluded by its version-format check, so it hasn't fired yet.
- No file trees to remove — versioned docs are generated at build time from this JSON list via `generate-versioned-docs.js`, never committed.

## Test plan
- [ ] Confirm docs site build picks up the updated `published-versions.json` and no longer generates/links `0.4.18`
- [ ] Confirm `0.6.0` and `0.5.69` versioned docs still build correctly


sibuthomasmathew (approved):
A trivial change. CI is green. lgtm

PR #2654, opened 2026-09-04
fix(ui): fail closed on application routes

sriaradhyula:
## Description

The hosted Agentic Apps migration removed `/apps/embed/<app-id>` in favor of the canonical `/apps/<app-id>` route. Existing links then resolve through the catch-all route as `appId=embed` and return a 404. This patch restores a compatibility redirect while adding a server-side authentication gate for every route below `/(app)`.

The gate requires a valid NextAuth/Keycloak session and SSO admission in production before rendering any application page. Agentic App catalog, authorization, and runtime APIs already require `allowAnonymous: false`; this change adds regression coverage for the runtime boundary and the shared page boundary. The explicit anonymous provider remains restricted to non-production development mode.

Preview evidence from the mirror deployment: `ssoEnabled=true`, `unsafeRbacBypassEnabled=false`, unauthenticated Agentic App API requests return 401, and the authenticated canonical dashboard shell reaches the configured runtime, which currently returns a separate upstream nginx 404.

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other (please describe)

## Checklist

- [x] New code is covered by automated tests
- [x] Focused tests and the production UI build pass


artdroz (changes requested):
This PR couples a useful legacy-URL compatibility fix to a broader authentication-policy change, and both need adjustment before merge.

The shared layout currently makes authentication mandatory whenever NODE_ENV is production, independent of SSO_ENABLED. The upstream production Compose and chart support production-mode installations with SSO disabled, so those deployments would redirect every application page to login without a usable session. This policy change needs its own explicit hosted/auth-required contract rather than being inferred from NODE_ENV. The login redirect should also preserve the requested callback/deep-link URL.

The compatibility route handles only the legacy root /apps/embed/<id>. Nested legacy links still fall through with embed interpreted as the app ID and return 404. Please make the compatibility route an optional catch-all that preserves and independently encodes all nested segments, repeated/empty query parameters, and then redirects to the canonical path.

Please split the redirect from the application-wide authentication change. Also replace the deployment-specific app name in the generic upstream test fixture.

artdroz (approved):
Re-reviewed with the server authentication boundary as the primary release requirement. Application pages currently depend on individual client-side AuthGuard placement; the shared server layout is the correct fail-closed backstop for missing and future page wrappers. Production fail-closed behavior is intentional, while anonymous access remains a local-development concern. The legacy nested-path and callback improvements are useful follow-ups but do not weaken this security fix and should not delay it. Withdrawing my prior block and approving for inclusion before the next tag.

PR #2655, opened 2026-09-04
feat(agentic-apps): authorize configured app access

sriaradhyula:
# Description

Adds an opt-in CAS/OpenFGA boundary for deployment-configured External Apps. Configured apps can declare an agentic_app authorization resource and route-level CAS actions; the host then enforces discovery, launch, and proxy decisions before minting app-bound tokens.

This also adds the agentic_app resource model, startup reconciliation for globally visible deployment-owned apps, documentation, and an updated Weather example. Private/team sharing remains outside this deployment-owned catalog contract.

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [x] Documentation
- [ ] Other

## Validation

- 7 focused Jest suites / 146 tests
- npx tsc --noEmit
- npm run lint -- --quiet
- npm run build
- git diff --check

## Checklist

- [x] Existing issues have been referenced where applicable
- [x] No new selection control is introduced
- [x] All relevant tests pass

### DIFF
--- a/.github/workflows/caipe-ui-tests.yml
+++ b/.github/workflows/caipe-ui-tests.yml
@@ -352,6 +352,9 @@ jobs:
           MONGODB_URI: 'mongodb://mock:27017'
           MONGODB_DATABASE: 'mock'
           NEXTAUTH_SECRET: 'ci-nextauth-secret-for-playwright-regression-2026'
+          CAIPE_UNSAFE_RBAC_BYPASS: 'true'
+          CAIPE_CREDENTIALS_ENABLED: 'true'
+          CAIPE_USER_CONNECTIONS_ENABLED: 'true'
         run: npm run build
 
       - name: Start CAIPE UI server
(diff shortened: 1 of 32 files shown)


artdroz (changes requested):
The logical CAS/OpenFGA boundary is the right direction, but I found several issues that need resolving before merge:

1. Model rollout ordering: the UI performs the configured-app tuple reconciliation only once during startup, while the chart installs the updated OpenFGA model in a post-install/post-upgrade hook. The UI can therefore attempt the new agentic_app tuple writes against the old model, swallow the failure, and never retry after the model is installed. In enforce mode this leaves every opted-in app denied until the UI is restarted. Please establish ordering or add a bounded/periodic retry and observable readiness behavior.

2. Invalid configuration can silently bypass CAS: route casAction is accepted without a top-level authorization block, but both catalog and runtime checks invoke CAS only when authorization exists. Please reject this combination during parsing, or make casAction imply the authorization contract.

3. The catalog is read from the ConfigMap on every request, but CAS tuples are reconciled only once. Hot-added or newly opted-in apps immediately enforce without grants, while removed or de-opted apps can leave stale wildcard/admin tuples. Please reconcile catalog changes, including cleanup of absent/de-opted IDs, or change the documented lifecycle to require a rollout and provide deterministic cleanup.

4. The Weather example marks its preferences POST as casAction write, but startup grants ordinary users only use. The model deliberately does not let use satisfy write, so this sample mutation is denied for every non-admin. Please keep that sample route at use, seed/document an appropriate writer grant, or state that it is intentionally admin-only.

Please also add catalog-route coverage for CAS read filtering, launch denial/unavailability, and opt-out behavior, and list the catalog route in the enforcement manifest.

PR #2656, opened 2026-09-04
fix(agentic-apps): preserve root mount trailing slash

sriaradhyula:
## Summary

- preserve the trailing slash when proxying the root request for mounted agentic apps
- add regression coverage for root and nested mounted paths

## Why

When `preserveMountPath` is enabled, the proxy must retain the mounted app's root slash. Without it, upstream servers that serve a framework base path can treat the mount path as a different resource and return 404.

## Validation

- `npm test -- --runInBand src/lib/agentic-apps/__tests__/security.test.ts`
- `npm run lint`


artdroz (approved):
Reviewed against the current upstream routing implementation. The URL builder now distinguishes the root request from nested suffixes and preserves the terminal slash only when preserveMountPath is enabled. Stripped-root and nested-path behavior remain unchanged, and the focused tests cover both cases.

This looks ready once the branch is updated with current main if branch protection requires it.

Issue #626, opened 2026-01-14
[ai-platform-engineering-e2i] Add read_file_tool for reading files outside workspace


sriaradhyula:
**Beads ID:** `ai-platform-engineering-e2i`
**Priority:** P2
**Type:** task
**Status:** open

---

Add a tool that can read files from any path (not just workspace), needed for reading files from cloned repos in /tmp/.

sriaradhyula:
Closing as completed. This was part of the Jan-2026 beads planning batch, now superseded by shipped work on main: the deepagents migration (`deepagents==0.3.8`, `create_deep_agent`, `utils/deepagents_custom`), spec-kit adoption (`.specify/` with constitution + templates), and beads integration (`.beads/`) are all in place. Reopen if anything was missed.

PR #2663, opened 2026-09-06
feat(ui): configure release notes from commit diffs

sriaradhyula:
## Summary
- restore the release-notes popup entry point in the current settings architecture
- allow administrators to configure a GitHub repository and two commit SHAs
- render the configured compare range for standard and non-standard release tags

## Why
The current upstream settings retained the release-notes toggle and preview, but the preview no longer honored an explicitly configured commit range. This made mirror tags such as `1.0.0-outshift.1` fall back to version-based notes or show nothing.

## Behavior
- all three compare fields are required together and validated
- private repositories use `RELEASE_NOTES_GITHUB_TOKEN`, `GITHUB_TOKEN`, or `GITHUB_PERSONAL_ACCESS_TOKEN`
- when configured, the popup uses the GitHub compare API and includes a compare link
- when not configured, existing deployed-version release notes behavior remains

## Validation
- focused API/UI tests: 66 passed
- UI lint: passed
- `git diff --check`: passed

This ports the feature already present in the CAIPE mirror and complements mirror PR #626.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "ai-platform-engineering"
-version = "1.0.1-dev.1"
+version = "1.0.1-dev.2"
 description = "Multi-Agent Collaboration & Workflow Automation - AI agents collaborating across tools and teams to deliver high quality results"
 authors = [
   {name = "CAIPE Contributors"}
--- a/ui/src/app/api/admin/platform-config/__tests__/route.test.ts
+++ b/ui/src/app/api/admin/platform-config/__tests__/route.test.ts
@@ -120,6 +120,9 @@ describe("admin platform-config route", () => {
 
     expect(body.data.release_notes).toEqual({
       enabled: true,
+      repository_url: null,
+      previous_commit: null,
+      latest_commit: null,
     });
   });
 
@@ -316,13 +319,72 @@ describe("admin platform-config route", () => {
       { _id: "platform_settings" },
       expect.objectContaining({
         $set: expect.objectContaining({
-          release_notes: { enabled: false },
+          release_notes: {
+            enabled: false,
+            repository_url: null,
+            previous_commit: null,
+            latest_commit: null,
+          },
         }),
       }),
       { upsert: true },
     );
     expect(collection.updateOne.mock.calls[0][1].$set).not.toHaveProperty("default_agent_id");
-    expect(body.data.release_notes).toEqual({ enabled: false });
+    expect(body.data.release_notes).toEqual({
+      enabled: false,
+      repository_url: null,
+      previous_commit: null,
+      latest_commit: null,
+    });
+  });
+
+  it("persists a GitHub commit range for release notes", async () => {
+    const collection = {
+      findOne: jest.fn().mockResolvedValue(null),
+      updateOne: jest.fn().mockResolvedValue({ acknowledged: true }),
+    };
+    mockGetCollection.mockResolvedValue(collection);
+    const { PATCH } = await import("../route");
+
+    const response = await PATCH(
+      request("/api/admin/platform-config", {
+        method: "PATCH",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify({
+          release_notes: {
+            enabled: true,
+            repository_url: "https://github.com/example/repository",
+            previous_commit: "1111111",
+            latest_commit: "2222222",
+          },
+        }),
+      }),
+    );
+    const body = await response.json();
+
+    expect(response.status).toBe(200);
+    expect(body.data.release_notes).toEqual({
+      enabled: true,
+      repository_url: "https://github.com/example/repository",
+      previous_commit: "1111111",
+      latest_commit: "2222222",
+    });
+  });
+
+  it("rejects an incomplete release notes commit range", async () => {
+    const { PATCH } = await import("../route");
+
+    await expect(
+      PATCH(
+        request("/api/admin/platform-config", {
+          method: "PATCH",
+          headers: { "Content-Type": "application/json" },
+          body: JSON.stringify({
+            release_notes: { previous_commit: "1111111" },
+          }),
+        }),
+      ),
+    ).rejects.toMatchObject({ code: "INCOMPLETE_RELEASE_NOTES_COMPARE" });
   });
 
   it("checks coarse admin before system_config manage on updates", async () => {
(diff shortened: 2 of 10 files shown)


artdroz (changes requested):
## Request changes

Reviewed commit **8007180ddfa4a7a541c8ecbb7a01bf80cb2111e7**. Rechecked the unchanged PR head, the full diff and relevant authentication/settings context, and the failed CI job. Checked existing reviews on this PR; there was no prior artdroz review to duplicate.

### 1. P1 — Authenticate the configured private-repository comparison before returning its contents

**Location:** [ui/src/app/api/release-notes/route.ts:498–516](https://github.com/caipe-io/ai-platform-engineering/blob/8007180ddfa4a7a541c8ecbb7a01bf80cb2111e7/ui/src/app/api/release-notes/route.ts#L498-L516); [server-token use:239–247](https://github.com/caipe-io/ai-platform-engineering/blob/8007180ddfa4a7a541c8ecbb7a01bf80cb2111e7/ui/src/app/api/release-notes/route.ts#L239-L247).

**Scenario:** Configure a private GitHub repository, valid previous/latest SHAs, and a server token with access. A caller without a session can request `/api/release-notes?version=anything&compare=platform`. The handler only requires a nonempty version, reads the stored configuration, fetches using the server token, and returns the rendered commit subjects, SHAs and repository/PR links.

**Authentication recheck:** The exported GET handler has no authentication wrapper or session check. [ui/src/proxy.ts:35–37](https://github.com/caipe-io/ai-platform-engineering/blob/8007180ddfa4a7a541c8ecbb7a01bf80cb2111e7/ui/src/proxy.ts#L35-L37) matches only `/apps/:path*`, not this API. I also checked `ui/next.config.ts`, `ui/src/instrumentation.ts`, and `ui/src/lib/api-middleware.ts`: the config/instrumentation do not install a global API authentication gate, and the authentication helpers require explicit invocation, which this route does not perform.

**Impact:** Private commit metadata becomes available to unauthenticated callers who can reach the UI API. Restricting the repository to stored platform configuration prevents arbitrary-repository queries, but does not authenticate the reader. This is not token disclosure or source-code disclosure. An independently configured upstream authentication gateway could mitigate external reachability; this review does not establish the configuration of any deployed gateway.

**Correction:** Authenticate the configured-compare branch using the established server-side authentication mechanism before accessing configuration, cached notes, or GitHub. Apply the appropriate read authorization for the intended announcement audience while retaining access for ordinary authorized users; do not accidentally make announcements admin-only through a generic wrapper default. Preserve deliberate development/no-SSO behavior consistently with existing auth helpers.

**Validation:** Add route tests with SSO enabled for no session, invalid/unauthorized credentials, and an authorized ordinary user. Unauthenticated/unauthorized requests must fail before database/GitHub work and must also fail after an authorized request has warmed the compare cache. An authorized user should still receive the configured comparison. Verify both public and private-repository configurations.

**Evidence level:** Confirmed missing application-level auth by code inspection and request-path tracing; the private-repository response scenario is inferred from that code, not reproduced against a live deployment.

### 2. P2 — Fix the new preview import breaking the admin-page Jest suite

**Location:** [ui/src/components/settings/sections/PlatformAnnouncementsSettings.tsx:3](https://github.com/caipe-io/ai-platform-engineering/blob/8007180ddfa4a7a541c8ecbb7a01bf80cb2111e7/ui/src/components/settings/sections/PlatformAnnouncementsSettings.tsx#L3).

**Scenario and evidence:** The failed [test job](https://github.com/caipe-io/ai-platform-engineering/actions/runs/35607027594/job/106356516691) reports `FAIL src/app/(app)/admin/__tests__/admin-page.test.tsx` with `SyntaxError: Unexpected token export` at `node_modules/react-markdown/index.js:10`. Its stack explicitly follows `ReleaseUpgradeDialog → ReleaseNotesPreview → PlatformAnnouncementsSettings → admin/page → admin-page.test`. This PR adds the preview import to that chain.

**Impact:** The admin-page suite fails during module loading, before its tests run. This is a related CI regression, not an assumed unrelated flaky check. The job summary is **1 failed / 679 passed suites**, with **7,754 executed tests passing**; passing focused feature tests does not cover the failing suite.

**Correction:** Update the admin-page test boundary to mock the newly introduced preview/dialog where appropriate, or configure Jest to transform the relevant ESM dependency chain if those components are intended to be exercised in that suite. Preserve useful admin-page coverage rather than skipping the suite.

**Validation:** Rerun the admin-page suite and the complete UI test/coverage command used by this job; confirm all suites load and the CI job passes. Keep the focused release-notes/settings tests passing as well.

**Evidence level:** Confirmed from the actual CI failure and its import stack; no local test rerun was performed for this review. This finding does not assert a production bundler failure.

### 3. P2 — Secondary: snapshot submitted compare fields to avoid marking unsaved edits as saved

**Location:** [ui/src/components/settings/sections/PlatformAnnouncementsSettings.tsx:170–201](https://github.com/caipe-io/ai-platform-engineering/blob/8007180ddfa4a7a541c8ecbb7a01bf80cb2111e7/ui/src/components/settings/sections/PlatformAnnouncementsSettings.tsx#L170-L201), together with [in-place ref mutation:77–84](https://github.com/caipe-io/ai-platform-engineering/blob/8007180ddfa4a7a541c8ecbb7a01bf80cb2111e7/ui/src/components/settings/sections/PlatformAnnouncementsSettings.tsx#L77-L84).

**Scenario:** Enter valid range A and click Save. While PATCH is pending, edit the latest SHA to B. Inputs remain editable during saving. `const fields = compareRef.current` aliases the mutable object: JSON serialization has already sent A, but editing mutates that same object. When the request succeeds, `setSavedLatestCommit(fields.latestCommit)` records B as saved even though the server persisted A.

**Impact:** The page displays Saved, clears its dirty state and disables Save for B; reloading restores A. The later edit is silently unpersisted. The shared SaveButton disables the button during saving but does not disable these inputs, so it does not prevent the sequence.

**Correction:** Capture an immutable, preferably trimmed, snapshot of the submitted fields before the request, use that snapshot for both the payload and the saved baseline, and leave newer edits dirty. Alternatively, disable field editing while saving. Keep the baseline aligned with the actual submitted/server-accepted values.

**Validation:** Use a deferred PATCH response in a component test: submit A, edit to B before resolving the request, then resolve successfully. Assert the payload contains A and B remains dirty/saveable (or assert editing is blocked if that is the chosen fix). A subsequent Save must send B.

**Evidence level:** Confirmed aliasing and enabled-input behavior by code inspection; this timing scenario was not browser-reproduced. This is a secondary correctness issue; the authentication gap and confirmed CI regression are the primary blockers.

### Summary and scope

Please address the missing authentication, repair the related CI import failure, and correct the secondary save-state race. All three earlier findings remain supported after rechecking; the security claim is explicitly limited to application-level protection and conditional private-repository exposure, not a demonstrated live exploit. No source edits were made.

PR #2665, opened 2026-09-06
fix(setup-caipe): make Langfuse tracing non-fatal and pin the chart

sriaradhyula:
# Description

`deploy_langfuse` ran `helm upgrade --install langfuse langfuse/langfuse` **unpinned**, with output sent to `/dev/null` and the call **unguarded**. Under `set -euo pipefail` a failed Langfuse install therefore **silently aborted the entire installer** — before the core CAIPE chart was ever deployed.

This is now easy to hit: the current `langfuse/langfuse` 2.x chart hard-requires cert-manager **and** the ClickHouse operator (`ClickHouseCluster` / `KeeperCluster` CRs), so `helm upgrade` fails at render time on a plain cluster:

```
Error: execution error at (langfuse/templates/validations.yaml:107:8):
ClickHouse operator CRDs not found ... Install the ClickHouse operator ... before installing this chart
```

## Changes

- **Pin** `LANGFUSE_CHART_VERSION` to `1.5.41` — the last release that runs without the ClickHouse operator / cert-manager (bundles ClickHouse + MinIO as plain pods). Overridable via the env var.
- Stop redirecting the `helm` output to `/dev/null`; capture it and print it on failure. `deploy_langfuse` now **returns non-zero** on a helm/rollout failure.
- Run the tracing block in `main()` as an `if` condition (which disables `set -e` inside it); on failure `warn`, set `ENABLE_TRACING=false`, and **continue the core install**.

## Testing

- `bash -n setup-caipe.sh`
- Forced failure (`LANGFUSE_CHART_VERSION=9.9.9-nope … --tracing`): install prints the error and proceeds instead of exiting.
- Full non-interactive install on Kind: log shows `✗ Timeout waiting for pods in langfuse` → `continuing WITHOUT tracing` → `deploy_caipe` runs to completion (previously died at the Langfuse step).
- Real `helm upgrade --install langfuse … --version 1.5.41` deploys on Kind.

Follow-up (separate issue): opt-in Langfuse 2.x path (cert-manager + clickhouse-operator, Kind-tuned).

## Type of Change

- [x] Bugfix

## Checklist

### DIFF
--- a/setup-caipe.sh
+++ b/setup-caipe.sh
@@ -171,6 +171,12 @@ RAG_INGESTOR_OIDC_ISSUER=""
 RAG_INGESTOR_OIDC_CLIENT_ID=""
 LANGFUSE_PUBLIC_KEY=""
 LANGFUSE_SECRET_KEY=""
+# Langfuse Helm chart version. Pinned to the last 1.x release on purpose: the
+# 2.x line hard-requires cert-manager + the ClickHouse operator (ClickHouseCluster
+# / KeeperCluster CRs, ~15 extra pods, 100Gi+ PVCs) and bumps the minimum
+# Kubernetes version — far too heavy for a first-install laptop stack. 1.5.41
+# bundles ClickHouse + MinIO as plain pods. Override with LANGFUSE_CHART_VERSION.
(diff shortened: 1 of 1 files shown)


sibuthomasmathew (approved):
Nice catch!

PR #2666, opened 2026-09-06
fix(setup-caipe): use the caipe-io chart registry and validate versions

sriaradhyula:
# Description

`CAIPE_OCI_REPO` was hardcoded to `oci://ghcr.io/cnoe-io/charts/ai-platform-engineering`. That path has been **frozen at `0.6.0`** since the project moved to the `caipe-io` org — `1.0.0` is published under `ghcr.io/caipe-io/charts/...`. Result: every fresh `setup-caipe.sh` install silently gets the stale `0.6.0` chart, and typing `1.0.0` at the version prompt looked like "not found".

The version picker also accepted any typed string with no validation, so a nonexistent version sailed through to a failing `helm upgrade --install caipe --version <x>`.

## Changes

- `CAIPE_OCI_REPO` defaults to `oci://ghcr.io/caipe-io/charts/ai-platform-engineering` (matches `.github/workflows/helm.yml`, which pushes to `ghcr.io/${{ github.repository_owner }}/charts`). Still overridable via the env var.
- The `crane` tag listing derives its target from `CAIPE_OCI_REPO` instead of a second hardcoded path (can't drift again).
- `choose_chart_version` validates any free-form version against the registry with `helm show chart --version` and re-prompts, with a "use anyway" escape.
- Non-numeric version-menu input re-prompts instead of aborting under `set -e`.

## Testing

- `bash -n setup-caipe.sh`
- `helm show chart oci://ghcr.io/caipe-io/charts/ai-platform-engineering` → `version: 1.0.0`.
- Validation logic: `1.0.0` / `0.6.0` accepted, `9.9.9` / `1.0.1` rejected with re-prompt.

## Type of Change

- [x] Bugfix

## Checklist

### DIFF
--- a/setup-caipe.sh
+++ b/setup-caipe.sh
@@ -11,7 +11,10 @@ set -euo pipefail
 
 # ─── Defaults ────────────────────────────────────────────────────────────────
 CAIPE_CHART_VERSION="${CAIPE_CHART_VERSION:-}"
-CAIPE_OCI_REPO="oci://ghcr.io/cnoe-io/charts/ai-platform-engineering"
+# Helm chart OCI registry. Charts publish to ghcr.io/<repo owner>/charts (see
+# .github/workflows/helm.yml), which moved to the caipe-io org — the cnoe-io
+# path is frozen pre-migration at 0.6.0. Override with CAIPE_OCI_REPO.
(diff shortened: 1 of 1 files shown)


sibuthomasmathew (approved):
Considering users of the script is very unlikely to go back to version < `1.0.0`, this change makes a lot of sense

PR #2667, opened 2026-09-06
fix(setup-caipe): harden interactive input and Kind cluster selection

sriaradhyula:
# Description

Interactive-input robustness fixes, all observed during a `curl | bash` run.

## Changes

- **`_trim_input`** helper — strips CR/LF/tab, surrounding whitespace, and a single stray leading backslash from prompt answers. `curl | bash` reads prompts from `/dev/tty` while bash still consumes the script on stdin, so a stray keystroke (a shell line-continuation `\`, a trailing CR) turned `1` into `\1` → `Unknown choice '\1'`. Runtime, cluster-menu, chart-version, and Ollama-model reads route through it.
- **Kind cluster name collision** — "Create a new Kind cluster" now detects an existing cluster of the same name and offers **Reuse / Delete+recreate / Rename** in a loop, and re-prompts on `kind create` failure, instead of `ERROR: node(s) already exist ...` aborting the script under `set -e`.
- **Ollama model prompt** — trimmed, soft-validated against the suggested list (warn only, never blocks), and the resolved value is echoed before the multi-GB pull, so a typo is visible immediately.
- Non-numeric cluster-menu input re-prompts instead of erroring under `set -e`.

## Testing

- `bash -n setup-caipe.sh`
- `_trim_input` unit tests: `\1`→`1`, `  2  `→`2`, `1\r`→`1`, `qwen3.8:27b` unchanged, `\tall`→`all`.

## Type of Change

- [x] Bugfix

## Checklist

### DIFF
--- a/setup-caipe.sh
+++ b/setup-caipe.sh
@@ -292,6 +292,20 @@ tty_read() { read "$@" <&3; }
 # Returns 0 (true) when the user wants to go back — accepts "b", "back", "0"
 _is_back() { local _v; _v="$(echo "$1" | tr '[:upper:]' '[:lower:]')"; [[ "$_v" == "b" || "$_v" == "back" || "$1" == "0" ]]; }
 
+# Sanitise an interactive answer: strip CR/LF/tabs, trim surrounding spaces, and
+# drop a single leading backslash. When run via "curl | bash", prompts are read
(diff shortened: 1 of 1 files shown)


sibuthomasmathew (approved):
Ran locally and verified. lgtm

PR #2668, opened 2026-09-06
fix(setup-caipe): ask before running sudo

sriaradhyula:
# Description

Require consent before privileged installation and host-configuration operations in `setup-caipe.sh`. Previously, some paths invoked sudo without permission, treated passwordless sudo as consent, or accepted default approval when reading a prompt failed.

## Changes

- Resolve sudo consent once and cache the answer. Failed prompt reads and unattended runs without approval deny sudo; passwordless sudo alone does not grant permission.
- Support `--allow-sudo`, `--no-sudo`, and `CAIPE_ALLOW_SUDO`. Explicit denial overrides approval flags and `--yes`, regardless of argument order.
- Accept `0`/`1` and case-insensitive `false`/`true` environment values. Unset or empty values use normal consent handling; invalid values are rejected before flag parsing.
- Gate Docker installation, Helm system installation, prerequisite package installation, k9s installation, supported binary moves, and the Docker-access sudo probe behind consent.
- Gate Linux ingress host routing, `/etc/hosts` changes, and reboot persistence behind consent. Denial leaves manual configuration and port-forward guidance.
- Use `~/.local/bin` fallbacks for supported binary installers and make that directory available on the current PATH. Install jq only through supported package managers.
- Document consent policy, automation examples, and fallback behavior in CLI help and installation guides.
- Add 43 mocked consent regression scenarios and a GitHub Actions workflow for shell syntax checks and consent tests on Linux and macOS. The scenarios passed locally; hosted execution of these local commits remains pending.

## Testing

- `bash -n setup-caipe.sh`

## Type of Change

- [x] Bugfix

## Checklist

### DIFF
--- a/setup-caipe.sh
+++ b/setup-caipe.sh
@@ -166,6 +166,7 @@ KEYCLOAK_PORT=7080
 OPENFGA_PORT=18080
 INJECT_CORPORATE_CA=false
 CA_SSL_FIX_PROMPTED=false
+SUDO_CONSENT=""   # "", "yes" or "no" — cached answer for _sudo_consent (ask once)
 RAG_INGESTOR_SECRET_READY=false
 RAG_INGESTOR_OIDC_ISSUER=""
 RAG_INGESTOR_OIDC_CLIENT_ID=""
@@ -318,6 +319,23 @@ ask_yn() {
   [[ "$answer" =~ ^[Yy]$ ]]
 }
 
+# Consent gate for anything that shells out to `sudo`. Returns 0 only after the
+# user explicitly allows it — a working passwordless sudo is NOT treated as
+# permission. The answer is cached for the rest of the run so we prompt once.
+# --yes / AUTO_YES pre-consents (non-interactive convenience).
+_sudo_consent() {
+  local reason="${1:-a system change}"
+  case "$SUDO_CONSENT" in
+    yes) return 0 ;;
+    no)  return 1 ;;
+  esac
+  if $AUTO_YES; then SUDO_CONSENT="yes"; return 0; fi
+  if ask_yn "This step needs sudo (${reason}). Allow this script to run sudo?" "y"; then
+    SUDO_CONSENT="yes"; return 0
+  fi
+  SUDO_CONSENT="no"; return 1
+}
+
 wait_for_pods() {
   local ns="$1" timeout="${2:-300}" exclude_pattern="${3:-}" interval=5 elapsed=0
   local show_interval=10 next_show=10
@@ -455,7 +473,11 @@ _install_kubectl_linux() {
   ver=$(curl -sL https://dl.k8s.io/release/stable.txt)
   curl -sLo /tmp/kubectl "https://dl.k8s.io/release/${ver}/bin/linux/amd64/kubectl"
   chmod +x /tmp/kubectl
-  sudo mv /tmp/kubectl /usr/local/bin/kubectl || mv /tmp/kubectl "$HOME/.local/bin/kubectl"
+  if _sudo_consent "move kubectl into /usr/local/bin"; then
+    sudo mv /tmp/kubectl /usr/local/bin/kubectl || { mkdir -p "$HOME/.local/bin" && mv /tmp/kubectl "$HOME/.local/bin/kubectl"; }
+  else
+    mkdir -p "$HOME/.local/bin" && mv /tmp/kubectl "$HOME/.local/bin/kubectl"
+  fi
   log "kubectl ${ver} installed"
 }
 
@@ -516,7 +538,11 @@ _install_jq_linux() {
       ver=$(curl -sL https://api.github.com/repos/jqlang/jq/releases/latest | grep -o '"tag_name": "[^"]*' | cut -d'"' -f4)
       curl -sLo /tmp/jq "https://github.com/jqlang/jq/releases/download/${ver}/jq-linux-amd64"
       chmod +x /tmp/jq
-      sudo mv /tmp/jq /usr/local/bin/jq || mv /tmp/jq "$HOME/.local/bin/jq"
+      if _sudo_consent "move jq into /usr/local/bin"; then
+        sudo mv /tmp/jq /usr/local/bin/jq || { mkdir -p "$HOME/.local/bin" && mv /tmp/jq "$HOME/.local/bin/jq"; }
+      else
+        mkdir -p "$HOME/.local/bin" && mv /tmp/jq "$HOME/.local/bin/jq"
+      fi
       ;;
   esac
   log "jq installed"
@@ -528,7 +554,11 @@ _install_kind_linux() {
   ver=$(curl -sL https://api.github.com/repos/kubernetes-sigs/kind/releases/latest | grep -o '"tag_name": "[^"]*' | cut -d'"' -f4)
   curl -sLo /tmp/kind "https://kind.sigs.k8s.io/dl/${ver}/kind-linux-amd64"
   chmod +x /tmp/kind
-  sudo mv /tmp/kind /usr/local/bin/kind || mv /tmp/kind "$HOME/.local/bin/kind"
+  if _sudo_consent "move kind into /usr/local/bin"; then
+    sudo mv /tmp/kind /usr/local/bin/kind || { mkdir -p "$HOME/.local/bin" && mv /tmp/kind "$HOME/.local/bin/kind"; }
+  else
+    mkdir -p "$HOME/.local/bin" && mv /tmp/kind "$HOME/.local/bin/kind"
+  fi
   log "kind ${ver} installed"
 }
 
@@ -541,7 +571,11 @@ _install_kind_macos() {
     ver=$(curl -sL https://api.github.com/repos/kubernetes-sigs/kind/releases/latest | grep -o '"tag_name": "[^"]*' | cut -d'"' -f4)
     curl -sLo /tmp/kind "https://kind.sigs.k8s.io/dl/${ver}/kind-darwin-arm64"
     chmod +x /tmp/kind
-    sudo mv /tmp/kind /usr/local/bin/kind || mv /tmp/kind "$HOME/.local/bin/kind"
+    if _sudo_consent "move kind into /usr/local/bin"; then
+      sudo mv /tmp/kind /usr/local/bin/kind || { mkdir -p "$HOME/.local/bin" && mv /tmp/kind "$HOME/.local/bin/kind"; }
+    else
+      mkdir -p "$HOME/.local/bin" && mv /tmp/kind "$HOME/.local/bin/kind"
+    fi
     log "kind ${ver} installed"
   fi
 }
@@ -683,12 +717,11 @@ check_prerequisites() {
         esac
       done
 
-      # If any tools need sudo, check whether sudo is usable and user consents
+      # If any tools need sudo, get explicit consent first (a working
+      # passwordless sudo does not imply permission).
       if [[ ${#needs_sudo[@]} -gt 0 ]]; then
         local sudo_ok=false
-        if sudo -n true 2>/dev/null; then
-          sudo_ok=true
-        elif ask_yn "Installing ${needs_sudo[*]} requires sudo. Allow this script to run sudo?" "y"; then
+        if _sudo_consent "install ${needs_sudo[*]}"; then
           sudo_ok=true
         fi
 
@@ -782,9 +815,7 @@ check_prerequisites() {
   if ! command -v k9s &>/dev/null; then
     if [[ "$(uname -s)" == "Linux" ]]; then
       local _k9s_sudo_ok=false
-      if sudo -n true 2>/dev/null; then
-        _k9s_sudo_ok=true
-      elif ask_yn "Installing k9s (Kubernetes TUI) requires sudo. Allow?" "y"; then
+      if _sudo_consent "install k9s (Kubernetes TUI)"; then
         _k9s_sudo_ok=true
       fi
       if [[ "$_k9s_sudo_ok" == true ]]; then
--- a/setup-caipe.sh
+++ b/setup-caipe.sh
@@ -672,24 +672,42 @@ _check_kubeconfig() {
 }
 
 _check_docker_access() {
-  # Docker binary present but socket not accessible without sudo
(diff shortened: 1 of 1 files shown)


setup-caipe.sh
```diff
@@ -317,6 +318,23 @@ ask_yn() {
...
+# permission. The answer is cached for the rest of the run so we prompt once.
+# --yes / AUTO_YES pre-consents (non-interactive convenience).
+_sudo_consent() {
+  local reason="${1:-a system change}"
+  case "$SUDO_CONSENT" in
+    yes) return 0 ;;
+    no)  return 1 ;;
+  esac
+  if $AUTO_YES; then SUDO_CONSENT="yes"; return 0; fi
+  if ask_yn "This step needs sudo (${reason}). Allow this script to run sudo?" "y"; then
```
sibuthomasmathew, line 332: `--non-interactive` without `--yes` silently self-consents.

In that mode line 249 points fd 3 at `/dev/null`. `ask_yn` (line 310) then does `tty_read -r answer` → immediate EOF → `answer` is empty → `answer="${answer:-$default}"` makes it `"y"` → this gate returns 0 with nobody having answered anything.

So the one mode where an unexpected sudo is most likely to go unnoticed is the mode where the gate is weakest. Suggest defaulting to deny there:

```bash
if $AUTO_YES; then SUDO_CONSENT="yes"; return 0; fi
if $NON_INTERACTIVE; then SUDO_CONSENT="no"; return 1; fi
```

That keeps `--yes` as the explicit "you may sudo unattended" switch and makes `--non-interactive` alone safe.

---
_Generated by [Claude Code](https://claude.ai/code)_

setup-caipe.sh
```diff
@@ -317,6 +318,23 @@ ask_yn() {
...
 
+# Consent gate for anything that shells out to `sudo`. Returns 0 only after the
+# user explicitly allows it — a working passwordless sudo is NOT treated as
+# permission. The answer is cached for the rest of the run so we prompt once.
+# --yes / AUTO_YES pre-consents (non-interactive convenience).
+_sudo_consent() {
+  local reason="${1:-a system change}"
+  case "$SUDO_CONSENT" in
+    yes) return 0 ;;
+    no)  return 1 ;;
```
sibuthomasmathew, line 329: A cached `no` isn't binding on the script as a whole.

`SUDO_CONSENT=no` only stops the call sites that actually ask. `_install_docker_linux` (583-622) and the Linux ingress path (2033-2120, `_persist_iptables` 2128-2174) call `sudo` directly, so a user who answered "no" here still gets sudo'd later in the same run — arguably worse than before the gate existed, because now they were asked and believe they declined.

Either route every `sudo` through the gate, or narrow the prompt text so it does not read as a script-wide policy ("Allow this script to run sudo?" currently promises more than it delivers).

---
_Generated by [Claude Code](https://claude.ai/code)_

setup-caipe.sh
```diff
@@ -317,6 +318,23 @@ ask_yn() {
...
+  esac
+  if $AUTO_YES; then SUDO_CONSENT="yes"; return 0; fi
```
sibuthomasmathew, line 331: Minor: this branch is unreachable in effect — `ask_yn` already short-circuits to `return 0` under `AUTO_YES` (line 311), so line 332 would consent anyway. Harmless, but it only sets `SUDO_CONSENT="yes"` a little earlier than the next line would.

Related gap worth considering while you're here: there's no way to express "run unattended but never sudo". `--yes` currently means both "don't prompt" and "you may escalate". A `--no-sudo` flag (or `CAIPE_ALLOW_SUDO=0`) that pins `SUDO_CONSENT="no"` up front is the natural companion to this change, especially for CI.

---
_Generated by [Claude Code](https://claude.ai/code)_

setup-caipe.sh
```diff
@@ -454,7 +472,11 @@ _install_kubectl_linux() {
   ver=$(curl -sL https://dl.k8s.io/release/stable.txt)
   curl -sLo /tmp/kubectl "https://dl.k8s.io/release/${ver}/bin/linux/amd64/kubectl"
   chmod +x /tmp/kubectl
-  sudo mv /tmp/kubectl /usr/local/bin/kubectl || mv /tmp/kubectl "$HOME/.local/bin/kubectl"
+  if _sudo_consent "move kubectl into /usr/local/bin"; then
```
sibuthomasmathew, line 475: This prompt doesn't buy the user anything.

`check_prerequisites` already does `mkdir -p "$HOME/.local/bin"` and `export PATH="$HOME/.local/bin:$PATH"` (lines 768-769) *before* this installer loop runs. So the declined branch produces a working `kubectl` on `PATH` exactly like the accepted one — the only difference is where the file lands.

Prompting for sudo where the no-sudo path is equally good is how users get trained to hit "y" on every sudo prompt, which undercuts the point of the PR. Suggest just always using `~/.local/bin` for these three binaries and dropping the gate here.

Same applies to `_install_jq_linux` (540), `_install_kind_linux` (556) and `_install_kind_macos` (573).

---
_Generated by [Claude Code](https://claude.ai/code)_

setup-caipe.sh
```diff
@@ -682,12 +734,11 @@ check_prerequisites() {
...
 
-      # If any tools need sudo, check whether sudo is usable and user consents
+      # If any tools need sudo, get explicit consent first (a working
+      # passwordless sudo does not imply permission).
       if [[ ${#needs_sudo[@]} -gt 0 ]]; then
         local sudo_ok=false
-        if sudo -n true 2>/dev/null; then
-          sudo_ok=true
-        elif ask_yn "Installing ${needs_sudo[*]} requires sudo. Allow this script to run sudo?" "y"; then
+        if _sudo_consent "install ${needs_sudo[*]}"; then
```
sibuthomasmathew, line 741: The classification just above contradicts the installer it dispatches to.

Line 732 puts `kubectl` in `no_sudo`, so a run where `kubectl` is the only missing tool skips this gate entirely — and then `_install_kubectl_linux` asks for sudo anyway at line 475. The user gets a sudo prompt for a tool the script just classified as not needing one.

Worth making the two agree: if the binaries go to `~/.local/bin` (see my comment on 475), `kubectl`/`jq`/`kind` are all genuinely `no_sudo` and only the package-manager tools (`openssl`, `curl`) need this gate.

---
_Generated by [Claude Code](https://claude.ai/code)_

setup-caipe.sh
```diff
@@ -637,24 +671,42 @@ _check_kubeconfig() {
...
+  # Only relevant when the daemon can't be reached as the current user.
+  command -v docker &>/dev/null || return 0
+  docker info &>/dev/null 2>&1 && return 0
+
+  # macOS/Windows: Docker Desktop is per-user, never root-socket based — a
+  # failing `docker info` means the daemon isn't up, and `sudo docker` cannot
+  # fix it. Never prompt for a password here.
+  if [[ "$(uname -s)" != "Linux" ]]; then
+    err "Docker does not appear to be running (\`docker info\` failed)."
+    err "Start Docker Desktop, wait until \`docker info\` succeeds, then re-run."
```
sibuthomasmathew, line 683: This is a behavior change beyond the PR's stated scope: on non-Linux a failed `docker info` now `exit 1`s, where previously the function fell through and the run continued.

The reasoning in the comment above is sound (Docker Desktop is per-user, `sudo docker` can't fix it) and the hard exit is probably the better UX. But it isn't "ask before running sudo", and the PR description doesn't mention it — worth calling out in the description so a reviewer isn't surprised by a new exit path in what reads as a consent-only change.

---
_Generated by [Claude Code](https://claude.ai/code)_

sibuthomasmathew (commented):
Right fix, right direction — treating a working passwordless `sudo` as consent was a real bug, and `_sudo_consent` is the correct shape for the solution. My concern is that the invariant the PR advertises isn't enforced yet, so a user can decline and still be sudo'd.

Inline comments cover what's in the diff. Three things sit outside it:

**1. `_install_docker_linux` (583-622) is completely ungated**

This is the largest escalation on the install path — `sudo apt-get install`, an apt keyring written into `/etc/apt/keyrings`, `sudo systemctl enable --now docker`, `sudo usermod -aG docker "$USER"` — and it never touches `_sudo_consent`. Its only gate is `ask_yn "Install Docker now?"` (line 801), which doesn't mention sudo and doesn't set `SUDO_CONSENT`.

Concretely: decline the sudo prompt during prerequisites, then answer "y" to "Install Docker now?" and the script adds a system apt source and modifies your groups. That's the exact surprise the PR set out to remove.

**2. The Linux ingress path stays ungated**

`install_nginx_ingress` (sudo block 2033-2120) and `_persist_iptables` (2128-2174): `sudo sysctl -w`, iptables NAT rules plus `DOCKER`/`DOCKER-USER` chain edits, a rewrite of `/etc/hosts` (2118-2119) and `/etc/sysctl.conf`, `/etc/iptables/rules.v4`, and installing + `systemctl enable` on a new `iptables-restore.service` unit.

Deferring this to a separate PR is entirely reasonable — it's Linux-only and non-trivial. But it's the most persistent root footprint in the script and it fires mid-run, long after any prompt, so the PR title isn't fully true while it stands. A linked follow-up issue would be better than only a line in the description.

**3. `bash -n` isn't evidence for this change**

Syntax checking can't tell you whether a declined answer is honored. The repo already has the pattern: `tests/integration/test_setup_caipe_mongodb_password.sh` awk-extracts a single helper out of `setup-caipe.sh` and `eval`s it in isolation. `_sudo_consent` is testable the same way — that declining returns non-zero and stays cached, that `--non-interactive` alone doesn't self-consent, that `--yes` still pre-consents.

**Smaller notes**

- `docs/docs/getting-started/quick-start.md` walks through the installer flow and isn't updated for the new prompt.
- Unrelated but adjacent: `cmd_cleanup` never removes the `/etc/hosts` entry, the iptables rules, or `iptables-restore.service` — the script leaves root-level state it can't undo. Probably its own issue.

None of this argues against merging the direction; the docker-install hole (1) and the non-interactive default (inline, line 332) are the two I'd want closed before this lands, since together they mean "no" isn't currently a reliable answer.

---
_Generated by [Claude Code](https://claude.ai/code)_

sibuthomasmathew (changes requested):
On a linux machine(Ubuntu 24.04), inspite of denying `sudo`, it went ahead with installing `docker` and `helm`
```bash
━━━ Checking prerequisites ━━━
  ✓ Auto-installing missing tools on Linux: kubectl helm
  ✓ Installing kubectl...
  ▸ This step needs sudo (move kubectl into /usr/local/bin). Allow this script to run sudo? [Y/n] n
  ✓ kubectl v1.37.0 installed
  ✓ Installing helm...
  ✓ helm installed
  ! Docker is not installed — it is required to run kind clusters.
  ▸ Install Docker now? [Y/n] y

━━━ Installing Docker ━━━
  ✓ Installing Docker...
Synchronizing state of docker.service with SysV service script with /usr/lib/systemd/systemd-sysv-install.
Executing: /usr/lib/systemd/systemd-sysv-install enable docker
  ✓ Docker installed — run 'newgrp docker' or re-login to use without sudo
  ! Cannot verify Docker socket access without sudo.
  ! If kind fails, add yourself to the 'docker' group: sudo usermod -aG docker $USER && newgrp docker
  ✓ Installing kind...
  ✓ kind v0.33.0 installed
  ! Skipping k9s install — run manually: https://k9scli.io/topics/install/
  ✓ Prerequisites checked (docker, kubectl, helm, kind, openssl, curl, jq, k9s)
```
```bash
sibu@computeinstance-e01vy1hpfgv1ggn9ew:~$ which docker
/usr/bin/docker
sibu@computeinstance-e01vy1hpfgv1ggn9ew:~$ which helm
/usr/local/bin/helm
sibu@computeinstance-e01vy1hpfgv1ggn9ew:~$ 
```
Then triggered Claude Code to run a thorough check. Hence the noise.

sibuthomasmathew (changes requested):
The installation of `helm` happened in `/usr/local/bin` without prompting for `sudo` access:
```bash
sibu@computeinstance-e01tm57rhabp5va558:~/ai-platform-engineering$ ./setup-caipe.sh 


   ██████╗ █████╗ ██╗██████╗ ███████╗
  ██╔════╝██╔══██╗██║██╔══██╗██╔════╝
  ██║     ███████║██║██████╔╝█████╗
  ██║     ██╔══██║██║██╔═══╝ ██╔══╝
  ╚██████╗██║  ██║██║██║     ███████╗
   ╚═════╝╚═╝  ╚═╝╚═╝╚═╝     ╚══════╝

╔══════════════════════════════════════════════════════════════════════╗
║  Open Source AI Platform for All                                     ║
║                                                                      ║
║  CAIPE empowers individuals and teams with 🤖 agentic AI automation. ║
║                                                                      ║
║  https://github.com/caipe-io/ai-platform-engineering                 ║
╚══════════════════════════════════════════════════════════════════════╝


  How do you want to run CAIPE?
    1) Kind/Kubernetes (default; uses saved setup configuration when available)
    2) Docker Compose (uses docker-compose.yaml + .env)

  ▸ Select runtime [1]: 

━━━ Checking prerequisites ━━━
  ✓ Auto-installing missing tools on Linux: kubectl helm
  ✓ Installing kubectl...
  ▸ This step needs sudo (move kubectl into /usr/local/bin). Allow this script to run sudo? [Y/n] n
  ✓ kubectl v1.37.0 installed
  ✓ Installing helm...
  ✓ helm installed
  ! Docker is not installed — it is required to run kind clusters.
  ▸ Install Docker now? [Y/n] n
  ✗ Docker is required. Install it from https://docs.docker.com/engine/install/ and re-run.
sibu@computeinstance-e01tm57rhabp5va558:~/ai-platform-engineering$ 
```
```bash
sibu@computeinstance-e01tm57rhabp5va558:~/ai-platform-engineering$ which kubectl 
/home/sibu/.local/bin/kubectl
sibu@computeinstance-e01tm57rhabp5va558:~/ai-platform-engineering$ which helm
/usr/local/bin/helm
sibu@computeinstance-e01tm57rhabp5va558:~/ai-platform-engineering$ 
```

setup-caipe.sh
sibuthomasmathew, line 538: `helm` is missing the gating

cisco-erilutz (approved):

PR #2677, opened 2026-09-06
fix(setup-caipe): let caipe-ui trust the self-signed cert for OIDC login

sriaradhyula:
# Description

Login to the UI fails on any install that uses the setup-generated self-signed cert (the default for `*.localtest.me` / any domain without `--tls-cert`). You authenticate at Keycloak fine, then the callback dies:

```
[NextAuth] OAUTH_CALLBACK_ERROR: self-signed certificate; if the root CA is installed
locally, try running Node.js with --use-system-ca
  code: 'DEPTH_ZERO_SELF_SIGNED_CERT'   providerId: 'oidc'
```

NextAuth does the OIDC token exchange + JWKS fetch **server-side** from the `caipe-ui` pod, against the public HTTPS domain Keycloak advertises (`KC_HOSTNAME`). The pod reaches the ingress (hostAlias) but Node.js rejects the self-signed cert → callback fails → browser bounces back to the sign-in page with no useful error.

## Fix

- Add a `TLS_SELF_SIGNED` flag, set `true` only when setup generates the cert (never when `--tls-cert` supplies a real one).
- In `post_deploy_patches`, when `TLS_SELF_SIGNED`, `kubectl set env` `NODE_TLS_REJECT_UNAUTHORIZED=0` on `caipe-caipe-ui`.

Blunt but scoped: local-dev self-signed installs only; production installs with a real cert keep full TLS verification. (A `NODE_EXTRA_CA_CERTS` mount of `caipe-tls` would be tighter — happy to switch if preferred.)

## Testing

- `bash -n setup-caipe.sh`
- Live on Kind (`*.localtest.me`, self-signed): login looped on `OAUTH_CALLBACK_ERROR`. `kubectl set env deploy/caipe-caipe-ui NODE_TLS_REJECT_UNAUTHORIZED=0` → pod reaches `https://<domain>/realms/caipe/.well-known/openid-configuration` (200), no more callback errors, login completes.

## Type of Change

- [x] Bugfix

## Checklist

### DIFF
--- a/setup-caipe.sh
+++ b/setup-caipe.sh
@@ -4220,6 +4220,19 @@ post_deploy_patches() {
     else
       log "rag-server: No OIDC config found in caipe-ui-secret — skipping OIDC patch (no-SSO deployment)"
     fi
+
+    # Self-signed cert: pin rag-server's OIDC discovery / JWKS to the in-cluster
+    # HTTP Keycloak so token validation (ui + ingestor providers) doesn't fetch
+    # from the untrusted public https issuer and 500 on CERTIFICATE_VERIFY_FAILED.
+    if [[ "${TLS_SELF_SIGNED:-false}" == true ]]; then
+      local _kc_int="http://caipe-keycloak:8080/realms/caipe"
+      kubectl set env deployment/rag-server -n caipe \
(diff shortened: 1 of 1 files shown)


sibuthomasmathew (approved):
Tested and verified. CI is green

PR #2685, opened 2026-09-06
fix(ui): keep default agent settings available without webex

sriaradhyula:
# Description

Keeps the default-agent settings available when the optional Webex admin service is disabled or temporarily unavailable.

- Skip Webex bot discovery when Webex is disabled.
- Degrade Webex discovery failures to an empty Webex list instead of failing the entire preferences request.
- Add retry handling to the default-agent settings panel.
- Allow users to set their Web default agent from the new-chat agent menu.
- Preserve compact display for long default-agent names.

This is the reusable upstream portion of https://github.com/cisco-eti/ai-platform-engineering-mirror/pull/627.

## Type of Change

- [x] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation
- [ ] Other

## Temporary Helm Charts (Optional)

Not applicable; this PR changes the UI only.

## Checklist

- [x] Existing issues have been referenced where applicable
- [x] Functionality is documented in the UI
- [x] The existing agent picker is reused; no new base selection control is introduced
- [x] Targeted tests and the production UI build pass

## Validation

- npm test -- --runInBand src/app/api/user/preferences/__tests__/route.test.ts src/components/chat/__tests__/NewChatButton.test.tsx (33 tests passed)
- npm run lint
- npm run build

### DIFF
--- a/ui/src/app/api/user/preferences/route.ts
+++ b/ui/src/app/api/user/preferences/route.ts
@@ -134,16 +134,30 @@ export const GET = withErrorHandler(async (request: NextRequest) => {
     });
   }
   const tenantId = resolveTenant(session);
+  const integrations = getIntegrationAvailability();
   const [preference, platformDefaultAgentId, webexBots] = await Promise.all([
     getUserPreference({ tenantId, userId: subject }),
     getResolvedPlatformDefaultAgentId(),
-    buildWebexBotSettings(subject),
+    // Do not make the personal Web/Slack settings page depend on the Webex
+    // service when Webex is not part of this deployment. In that case the
+    // service call can fail even though the user preferences are healthy.
+    integrations.webex
+      ? buildWebexBotSettings(subject).catch((error: unknown) => {
+          // Web and Slack preferences remain usable during a Webex admin
+          // service outage. The Webex rows will reappear on the next load.
+          console.warn(
+            "[user-preferences] Webex settings unavailable",
+            error instanceof Error ? error.message : String(error),
+          );
+          return [];
+        })
+      : Promise.resolve([]),
   ]);
   return successResponse({
     ...preference,
     webex_bots: webexBots,
     platform_default_agent_id: platformDefaultAgentId,
-    integrations: getIntegrationAvailability(),
+    integrations,
   });
 });
 
--- a/ui/src/components/chat/NewChatButton.tsx
+++ b/ui/src/components/chat/NewChatButton.tsx
@@ -5,7 +5,7 @@ import { Button } from "@/components/ui/button";
 import { resolveUsableChatAgent } from "@/lib/chat-agent-selection";
 import { cn } from "@/lib/utils";
 import type { DynamicAgentConfig } from "@/types/dynamic-agent";
-import { ChevronDown,Loader2,Plus,Search } from "lucide-react";
+import { Check,ChevronDown,Loader2,Plus,Search,Star } from "lucide-react";
 import React,{ useEffect,useRef,useState } from "react";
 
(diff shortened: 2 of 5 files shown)


artdroz (approved):
Lightweight spot-check of the diff and relevant context: no blocking findings. Webex discovery gating and default-agent persistence look sound. Minor follow-up: separate save errors from agent-list errors so a failed preference save does not hide the loaded list. This is a scoped review, not exhaustive validation; no full local test suite was run.

ui/src/components/chat/NewChatButton.tsx
```diff
@@ -133,6 +136,31 @@ export function NewChatButton({ collapsed, onNewChat }: NewChatButtonProps) {
...
+      });
+      const data = await response.json() as { success?: boolean; error?: string };
+      if (!response.ok || !data.success) {
+        throw new Error(data.error || `Failed to save default agent (HTTP ${response.status})`);
+      }
+      setDefaultAgentId(agent._id);
+      setPersonalDefaultAgentId(agent._id);
+      setDefaultAgentName(agent.name);
+    } catch (err) {
+      setError(err instanceof Error ? err.message : "Failed to save default agent");
```
cisco-erilutz, line 158: **Confirmed bug.** `setError` here writes into the same `error` state that gates rendering of the whole agent list (`{!loading && !error && filteredAgents.map(...)}` at line 269, plus the two empty-state branches right below it).

So a failed default-agent save doesn't just show an inline error next to the item being saved — it hides the *entire* dropdown list, even though the agents themselves loaded fine. The user is left with only the error text and no way to pick a different agent or retry without closing and reopening the dropdown.

Suggest a separate `saveError` (or similar) state dedicated to this handler, distinct from the agent-list-load `error`.

ui/src/app/api/user/preferences/route.ts
```diff
@@ -134,16 +134,30 @@ export const GET = withErrorHandler(async (request: NextRequest) => {
...
+  const integrations = getIntegrationAvailability();
   const [preference, platformDefaultAgentId, webexBots] = await Promise.all([
     getUserPreference({ tenantId, userId: subject }),
     getResolvedPlatformDefaultAgentId(),
-    buildWebexBotSettings(subject),
+    // Do not make the personal Web/Slack settings page depend on the Webex
+    // service when Webex is not part of this deployment. In that case the
+    // service call can fail even though the user preferences are healthy.
+    integrations.webex
+      ? buildWebexBotSettings(subject).catch((error: unknown) => {
```
cisco-erilutz, line 145: **Confirmed bug.** This `.catch()` is unconditional, so it swallows *any* throw from `buildWebexBotSettings` — not just a Webex service outage, which is the stated intent per the comment above. A genuine bug in `buildWebexBotSettings` (e.g. a malformed document causing a `TypeError`) gets silently downgraded to `webex_bots: []`, indistinguishable from "no bots configured," with only a `console.warn` — nothing surfaces to the client or triggers an alert. That makes real regressions in this path easy to ship unnoticed.

Worth narrowing the catch to the specific outage conditions you actually want to tolerate (e.g. connection/timeout errors), or at least re-throwing/logging distinctly for unexpected error types so they don't look identical to an intentional empty state.

cisco-erilutz (commented):
Automated review pass. Two confirmed correctness bugs flagged inline below. Also spotted (not filed as top comments, happy to add if useful): the star button on an already-selected default agent still fires a redundant PUT instead of no-op'ing, and there's some duplicated PUT-preference logic between this component and `UserDefaultAgentsPanel.tsx`.

PR #2384, opened 2026-08-13
feat(rbac): add private personal resources

sriaradhyula:
# Description

Adds private ownership for dynamic agents, MCP servers, and credentials.

Private resources are only available to their owner in the web UI and in Slack or Webex direct messages. They cannot be used in channels or group spaces.

- Adds CAS-protected OpenFGA owner, creator, and private-marker tuples.
- Carries short-lived signed context through agent, MCP, and credential authorization.
- Denies group conversations, schedules, service accounts, delegated agents, and non-owner access.
- Adds Private/Team/Global UI controls and automated unit/live E2E coverage.
- Documents the policy, data model, threat model, and rollout behavior.

## Type of Change

- [ ] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [x] Documentation
- [ ] Other

## Validation

- Mirror full UI suite — 619 suites / 7,296 tests passed
- TypeScript, ESLint, and production build — passed
- OpenFGA bridge tests — 36 passed
- Live Playwright E2E against `https://tome-sri.dev.outshift.io` — passed

## Checklist

- [x] I have read the contributing guidelines
- [x] Existing issues/PRs have been referenced where applicable
- [x] I have verified this change is not present in another implementation PR
- [x] Functionality is documented
- [x] All code style checks pass
- [x] New code contribution is covered by automated tests
- [x] All relevant new and existing tests pass

PR #2695, opened 2026-09-08
fix(rbac): harden private resource ownership transitions

sriaradhyula:
# Description

Harden private-resource ownership, visibility, reconciliation, and execution across dynamic agents, MCP servers, skills, workflows, and credentials. MongoDB remains the desired-state source of truth while OpenFGA is projected safely.

This PR is intentionally based on the private-resources feature branch because private/personal agents are not yet present on upstream main. It is intended to be folded into #2384 before that feature merges.

This change:

- fixes team-to-private agent and direct MCP update conflicts
- persists monotonic authorization revisions in pending state before ownership, visibility, sharing, or agent-tool changes are projected
- verifies grants and revocations at higher consistency before marking the matching revision ready
- leaves failed transitions retriable and fails discovery/execution closed while keeping the builder available for repair
- applies owner-only private discovery and management to skills and workflows, including skill catalog content and workflow runs
- persists stable owner subjects and desired team shares for skill saves and ZIP imports
- prevents global/team agents and workflows from embedding narrower private dependencies; same-owner private composition remains supported
- preserves hub skills as global dependencies and supports legacy owner-email rows until stable subjects are backfilled
- shows a subtle authorization sync status beside Agent, MCP, Skill, and Workflow builder controls

Mirror counterpart: cisco-eti/ai-platform-engineering-mirror#646

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [x] Refactor
- [ ] Documentation
- [ ] Other

## Validation

- Full UI Jest suite: 572 suites, 6,984 tests passed
- Focused upstream authorization and builder matrix: 13 suites, 196 tests passed
- TypeScript: `tsc --noEmit` passed
- ESLint passed
- `git diff --check` passed

## Checklist

- [x] Existing upstream feature PR #2384 is referenced
- [x] All relevant existing tests pass

### DIFF
--- a/ui/src/app/api/__tests__/agent-skill-content.test.ts
+++ b/ui/src/app/api/__tests__/agent-skill-content.test.ts
@@ -44,16 +44,24 @@ jest.mock("@/lib/mongodb", () => ({
   isMongoDBConfigured: true,
 }));
 
+jest.mock("@/lib/rbac/skill-team-grants", () => ({
+  reconcileSkillTeamShares: jest.fn().mockResolvedValue({
+    teamSlugs: [],
+    writesPlanned: 0,
+    writesApplied: 0,
+    deletesPlanned: 0,
+    deletesApplied: 0,
+    enabled: false,
+  }),
+  readSkillSharedTeamSlugsFromOpenFga: jest.fn().mockResolvedValue([]),
+}));
+
 function createMockCollection() {
-  const findReturnValue = {
-    project: jest.fn().mockReturnValue({
-      toArray: jest.fn().mockResolvedValue([]),
-    }),
-    sort: jest.fn().mockReturnValue({
-      toArray: jest.fn().mockResolvedValue([]),
-    }),
-    toArray: jest.fn().mockResolvedValue([]),
-  };
+  const findReturnValue: Record<string, jest.Mock> = {};
+  findReturnValue.project = jest.fn().mockReturnValue(findReturnValue);
+  findReturnValue.sort = jest.fn().mockReturnValue(findReturnValue);
+  findReturnValue.limit = jest.fn().mockReturnValue(findReturnValue);
+  findReturnValue.toArray = jest.fn().mockResolvedValue([]);
 
   return {
     find: jest.fn().mockReturnValue(findReturnValue),
@@ -61,6 +69,7 @@ function createMockCollection() {
     insertOne: jest.fn().mockResolvedValue({ insertedId: "test-id" }),
     updateOne: jest.fn().mockResolvedValue({ matchedCount: 1, modifiedCount: 1, acknowledged: true }),
     deleteOne: jest.fn().mockResolvedValue({ deletedCount: 1 }),
+    deleteMany: jest.fn().mockResolvedValue({ deletedCount: 0 }),
     countDocuments: jest.fn().mockResolvedValue(0),
   };
 }
@@ -73,6 +82,7 @@ function userSession(email = "user@example.com") {
   return {
     user: { email, name: "Test User" },
     role: "user",
+    sub: "user-sub",
   };
 }
 
@@ -373,7 +383,7 @@ def hello():
     expect(insertedConfig.skill_content).toBe(SAMPLE_SKILL_CONTENT);
     expect(insertedConfig.is_quick_start).toBe(true);
     expect(insertedConfig.visibility).toBe("team");
-    expect(insertedConfig.shared_with_teams).toBeUndefined();
+    expect(insertedConfig.shared_with_teams).toEqual(["team-sre", "team-devops"]);
   });
 });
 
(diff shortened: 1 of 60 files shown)


artdroz (changes requested):
## Required fixes

Reviewed commit: `1f669e3bf55b3186941953e4bbd54c3c5f2eda84` (current head verified before submission). This review is limited to PR #2695. Both findings below were rechecked against the actual code and relevant callers. The triggering scenarios are code-derived; I did not execute an integration reproduction.

### 1. P1 — Reconcile outstanding credential share changes before marking a later revision ready

**Location:** [`ui/src/lib/credentials/secret-service.ts:612–614`](https://github.com/caipe-io/ai-platform-engineering/blob/1f669e3bf55b3186941953e4bbd54c3c5f2eda84/ui/src/lib/credentials/secret-service.ts#L612-L614), with completion at [lines 624–635](https://github.com/caipe-io/ai-platform-engineering/blob/1f669e3bf55b3186941953e4bbd54c3c5f2eda84/ui/src/lib/credentials/secret-service.ts#L624-L635).

**Observed code:** The service persists the desired share set and retains `authz_previous_state` before reconciliation. On failure it records `error`. A subsequent share/revoke request is allowed, but its reconciler only calls `deleteShare` or `reconcileShare` for that request’s single `teamId`. Success marks the entire revision ready and deletes `authz_previous_state`, without reconciling earlier outstanding operations. The production factory wires these callbacks to OpenFGA operations; team shares include both `metadata_reader` and `user` tuples.

**Concrete triggering scenario:** Start with a credential shared with team A. Revoke A, and have OpenFGA fail before deleting A’s tuples. Mongo now excludes A and records an error. After OpenFGA recovers, share the credential with team B instead of retrying A’s revoke. The second request writes B’s tuples and marks the credential ready, but never deletes A’s tuples. The credential is now team-visible, so the private-context restriction does not compensate for the stale grant.

**Impact (code-derived):** A retains authorization that the desired state says was revoked; marking the revision ready removes the execution sync gate. This is stale credential-use authorization, not a claim that a browser can directly retrieve plaintext: the retrieval service still requires its internal-caller checks.

**Recommended correction:** Reconcile the complete desired share state, including outstanding revocations from failed attempts, before clearing recovery metadata or marking ready. Account for partially applied attempts as well as the last successful state, for example by reading/reconciling actual share tuples or retaining a complete pending-operation set. Preserve owner grants when removing obsolete share tuples.

**Meaningful validation:** Use a stateful tuple-store fake or OpenFGA integration test: fail revoke A before deletion, then successfully share B. Assert A’s share tuples are absent, B’s are present, and a non-owner member of A without another grant is denied `use` after the resource becomes ready. Also cover a partially applied failed share followed by a different operation; an unrelated success must not discard unresolved changes.

### 2. P1 — Do not use the legacy email fallback when a stable skill owner subject exists

**Location:** [`ui/src/app/api/skills/route.ts:321–326`](https://github.com/caipe-io/ai-platform-engineering/blob/1f669e3bf55b3186941953e4bbd54c3c5f2eda84/ui/src/app/api/skills/route.ts#L321-L326).

**Observed code:** `ownsLegacyRow` compares email without checking that `owner_subject` is absent. The private-skill branch returns the skill when either subject or email matches, before the OpenFGA check. The GET handler supplies the authenticated email and applies this same filter to both metadata and content-enabled catalog requests ([lines 736–757](https://github.com/caipe-io/ai-platform-engineering/blob/1f669e3bf55b3186941953e4bbd54c3c5f2eda84/ui/src/app/api/skills/route.ts#L736-L757)). Other private-skill ownership checks in this PR treat the stable subject as authoritative.

**Concrete triggering scenario:** A ready private skill stores `owner_subject=old-sub` and `owner_id=person@example.com`. A different authenticated identity has `sub=new-sub` and that same email, such as after email reassignment or account recreation. The subject comparison fails, but the email comparison succeeds and returns the skill.

**Impact (code-derived):** The different identity can discover the private skill and receive its catalog content through a content-enabled request. This scenario requires an authenticated identity with the same email; it does not assume arbitrary email spoofing.

**Recommended correction:** If a stable `owner_subject` is present, require that subject to match. Use normalized email matching only for genuinely legacy rows without a stable owner subject. Prefer a shared ownership predicate so catalog and workspace routes agree.

**Meaningful validation:** Add tests for both catalog metadata and content-enabled requests using equal emails but conflicting subjects; neither may return the private skill, even if an OpenFGA mock would allow it. Retain positive tests for a matching stable subject after an email change and for a legacy row with no subject and a matching normalized email.

## CI context — not code blockers

The failed checks were investigated rather than attributed to this patch: `playwright-rbac-regression` failed while pulling `ghcr.io/cnoe-io/playwright:v1.61.0-noble` with a registry denial, before tests started; `build-and-push` reached image publication and failed with `installation not allowed to Create organization package`. These are registry-permission failures, not the basis for this REQUEST_CHANGES review. The required changes above are based on the authorization code paths.

PR #2715, opened 2026-09-09
fix(dynamic-agents): support structured response middleware

sriaradhyula:
## Summary

- add structured-response middleware for dynamic agents
- normalize structured output for runtime and custom SSE streaming
- add focused regression coverage

This ports the change from [cisco-eti/ai-platform-engineering-mirror#645](https://github.com/cisco-eti/ai-platform-engineering-mirror/pull/645).

## Testing

- `uv run ruff check` on all changed Python files
- `uv run pytest -q tests/test_structured_response.py` (3 passed)

### DIFF
--- a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py
+++ b/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py
@@ -89,6 +89,12 @@ from dynamic_agents.services.model_capabilities import (
     get_model_capabilities,
 )
 from dynamic_agents.services.skills import build_skills_files, detect_missing_skills, load_skills
+from dynamic_agents.services.structured_response import (
+    StructuredResponseFormat,
+    build_structured_response_instruction,
+    create_submit_structured_response_tool,
+    extract_response_format,
+)
 
 if TYPE_CHECKING:
     from dynamic_agents.services.mongo import MongoDBService
@@ -707,6 +713,8 @@ class AgentRuntime:
         self._failed_skills_error: str = ""  # Error message for display
         self._failed_workflows: list[str] = []  # Workflow config IDs not found
         self._failed_workflows_error: str = ""  # Error message for display
+        self._structured_response: dict[str, Any] | None = None
+        self._structured_response_schema_id: str | None = None
         self._mcp_credential_warnings: list[str] = []
         self._mcp_credential_failed_server_ids: set[str] = set()
         self._valid_workflow_configs: list[str] = []  # Validated workflow config IDs
@@ -976,6 +984,10 @@ class AgentRuntime:
             logger.error(f"Agent '{self.config.name}' failed to initialize: {exc}")
             raise RuntimeError(f"Agent '{self.config.name}' failed to initialize: {exc}") from exc
 
+        response_format = self._get_allowed_structured_response_format(client_ctx)
+        if response_format:
+            system_prompt += build_structured_response_instruction(response_format)
+
         # 5. Instantiate LLM
         logger.info(
             f"[llm] Instantiating LLM for agent '{self.config.name}': "
@@ -1235,6 +1247,50 @@ class AgentRuntime:
             f"tools={len(tools)}, subagents={len(subagents) if subagents else 0}"
         )
 
+    def _capture_structured_response(
+        self, payload: dict[str, Any], schema_id: str | None
+    ) -> None:
+        """Capture the latest validated structured response submitted by the agent."""
+        self._structured_response = payload
+        self._structured_response_schema_id = schema_id
+
+    def get_structured_response(self) -> dict[str, Any] | None:
+        """Return the latest validated structured response for this runtime turn."""
+        return self._structured_response
+
+    def get_structured_response_schema_id(self) -> str | None:
+        """Return the schema ID for the captured structured response."""
+        return self._structured_response_schema_id
+
+    def _get_allowed_structured_response_format(
+        self, client_context: dict | None
+    ) -> StructuredResponseFormat | None:
+        """Return requested structured output only when the agent allows it."""
+        response_format = extract_response_format(client_context)
+        if response_format is None or self.config.features is None:
+            return None
+
+        for entry in self.config.features.middleware:
+            if entry.type != "structured_response" or not entry.enabled:
+                continue
+
+            allowed_schema_ids = str(entry.params.get("allowed_schema_ids", "")).strip()
+            if not allowed_schema_ids:
+                return response_format
+
+            allowed = {item.strip() for item in allowed_schema_ids.split(",") if item.strip()}
+            if response_format.schema_id in allowed:
+                return response_format
+
+            logger.warning(
+                "Agent '%s': structured response schema '%s' not allowed",
+                self.config.name,
+                response_format.schema_id,
+            )
+            return None
+
+        return None
+
     def _build_builtin_tools(
         self,
         user: UserContext | None = None,
@@ -1256,7 +1312,25 @@ class AgentRuntime:
         tools = []
         config_summary: dict[str, Any] = {}
 
+        response_format = self._get_allowed_structured_response_format(client_context)
+        is_parent_agent = agent_config is None or config.id == self.config.id
+        if response_format and is_parent_agent:
+            tools.append(
+                create_submit_structured_response_tool(
+                    response_format=response_format,
+                    on_submit=lambda payload: self._capture_structured_response(
+                        payload,
+                        response_format.schema_id,
+                    ),
+                )
+            )
+            config_summary["submit_structured_response"] = {
+                "schema_id": response_format.schema_id
+            }
+
         if not config.builtin_tools:
+            if tools:
+                logger.info(f"Agent '{config.name}': added built-in tools: {config_summary}")
             return tools
 
         # fetch_url tool (disabled by default)
@@ -1759,6 +1833,8 @@ class AgentRuntime:
         assert encoder is not None, "encoder must be provided"
 
         self._cancelled = False
+        self._structured_response = None
+        self._structured_response_schema_id = None
 
         config = self._build_stream_config(session_id, user_id, trace_id)
         run_id = f"run-{uuid4().hex[:12]}"
@@ -1887,6 +1963,16 @@ class AgentRuntime:
             for frame in self._emit_interrupt(encoder, interrupt_data):
                 yield frame
             return
+
+        structured_response = self.get_structured_response()
+        if structured_response is not None:
+            on_structured_output = getattr(encoder, "on_structured_output", None)
+            if callable(on_structured_output):
+                for frame in on_structured_output(
+                    structured_response,
+                    self.get_structured_response_schema_id(),
+                ):
+                    yield frame
 
         # ── Core lifecycle: run finish ──
         logger.info(
@@ -2169,6 +2255,8 @@ class AgentRuntime:
         assert encoder is not None, "encoder must be provided"
 
         self._cancelled = False
+        self._structured_response = None
+        self._structured_response_schema_id = None
 
         config = self._build_stream_config(session_id, user_id, trace_id)
         run_id = f"run-{uuid4().hex[:12]}"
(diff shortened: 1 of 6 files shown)


cisco-erilutz (commented):
Automated review pass — 10 findings, most severe first. Two don't anchor to changed lines so noting them here:

1. **`ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py:1962`** (just outside the changed hunk, but interacts directly with the new code at line 1964): the HITL interrupt check returns early before the new structured-output emission block is reached. If structured output is submitted before a tool-approval interrupt in the same turn, the client never gets the `structured_output` SSE frame, and `_resume_impl` later resets `self._structured_response = None`, permanently discarding the payload.

2. **`ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/stream_encoders/agui_sse.py:77`** (file not touched by this PR): `AGUIStreamEncoder` never implements `on_structured_output`, unlike other encoders. Since `agent_runtime.py` calls it via `getattr(encoder, "on_structured_output", None)` + a callable check, this silently no-ops for AG-UI clients that request `response_format` — no error, no log, just a missing frame.

ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py
```diff
@@ -1759,6 +1833,8 @@ async def _stream_impl(
         assert encoder is not None, "encoder must be provided"
 
         self._cancelled = False
+        self._structured_response = None
```
cisco-erilutz, line 1836: `_structured_response` / `_structured_response_schema_id` are plain instance attributes on an `AgentRuntime` that `AgentRuntimeCache` pools by `(agent_id, session_id)`, with no lock around `_stream_impl`/`_resume_impl`. Concurrent requests for the same session share one cached instance, so one request's reset-to-`None` can wipe another's just-captured structured response, or read the wrong session's payload.

ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/structured_response.py
```diff
@@ -0,0 +1,122 @@
...
+        return "structured response payload must be a JSON object"
+
+    for field in response_format.required:
+        if field not in payload:
+            return f"structured response payload is missing required field '{field}'"
+
+    properties = response_format.schema.get("properties")
+    if not isinstance(properties, dict):
+        return None
+
```
cisco-erilutz, line 64: `_validate_payload` only checks the top-level JSON type of each property and never recurses into nested object/array properties (required fields, nested types), even though `jsonschema` is already a repo dependency. A payload with a nested object missing a required field (e.g. `meta: {}` where the schema requires `meta.id`) passes `isinstance({}, dict)` and gets accepted as conforming, even though it doesn't.

ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/structured_response.py
```diff
@@ -0,0 +1,122 @@
...
+    for field, rules in properties.items():
+        if field not in payload or payload[field] is None or not isinstance(rules, dict):
```
cisco-erilutz, line 66: An explicit `null` for a required field is treated as "skip validation" here (`payload[field] is None` → `continue`), so a required non-nullable field submitted as `null` passes both the required-presence check and the type check, and gets accepted with a null value where callers expect a real value.

ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/middleware.py
```diff
@@ -409,6 +409,18 @@ class MiddlewareSpec:
             "exit_behavior": "end|error|continue",
         },
     ),
+    "structured_response": MiddlewareSpec(
+        cls=object,
+        default_params={"allowed_schema_ids": "", "require_tool_submission": True},
```
cisco-erilutz, line 414: `require_tool_submission` is declared in the structured_response `MiddlewareSpec`'s `default_params`/`param_schema` (exposed as a UI toggle) but isn't read anywhere in the runtime logic. Toggling it in the agent config UI currently has zero effect.

ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py
```diff
@@ -976,6 +984,10 @@ async def initialize(self) -> None:
             logger.error(f"Agent '{self.config.name}' failed to initialize: {exc}")
             raise RuntimeError(f"Agent '{self.config.name}' failed to initialize: {exc}") from exc
 
+        response_format = self._get_allowed_structured_response_format(client_ctx)
```
cisco-erilutz, line 987: `_get_allowed_structured_response_format` is computed once in `initialize()` and again inside `_build_builtin_tools` for every agent config including subagents (where the result is discarded for non-parent agents). This redundantly re-parses `client_context` and re-scans the middleware list, and can double-log the same "not allowed" warning.

ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py
```diff
@@ -1256,7 +1312,25 @@ def _build_builtin_tools(
         tools = []
         config_summary: dict[str, Any] = {}
 
+        response_format = self._get_allowed_structured_response_format(client_context)
+        is_parent_agent = agent_config is None or config.id == self.config.id
```
cisco-erilutz, line 1316: `is_parent_agent` is derived as `agent_config is None or config.id == self.config.id`. Since `config = agent_config or self.config` on the prior line, `agent_config is None` already implies `config is self.config`, so the second clause only matters if a subagent's config id happens to collide with the parent's — an unusual, easy-to-misconfigure state that would incorrectly grant a subagent the structured-response tool meant only for the top-level agent. Worth an explicit is-root flag instead.

ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py
```diff
@@ -1256,7 +1312,25 @@ def _build_builtin_tools(
...
+                    ),
+                )
+            )
+            config_summary["submit_structured_response"] = {
+                "schema_id": response_format.schema_id
+            }
+
         if not config.builtin_tools:
+            if tools:
+                logger.info(f"Agent '{config.name}': added built-in tools: {config_summary}")
```
cisco-erilutz, line 1333: This duplicates the exact `logger.info("added built-in tools: ...")` line that already exists at the function's tail. Now there are two copies that can drift out of sync if the log format/fields change in only one.

ai_platform_engineering/dynamic_agents/tests/test_structured_response.py
```diff
@@ -0,0 +1,69 @@
...
+from dynamic_agents.services.structured_response import (
+    StructuredResponseFormat,
+    create_submit_structured_response_tool,
+    extract_response_format,
+)
+
+
+def _response_format() -> StructuredResponseFormat:
+    return StructuredResponseFormat(
+        schema_id="jira_project.dashboard.v1",
```
cisco-erilutz, line 18: This fixture uses the real product name `"jira"` in `schema_id="jira_project.dashboard.v1"` (repeated 6x in this file). Per this repo's CLAUDE.md test-data policy, tests/fixtures should use neutral identifiers (e.g. `example_project.dashboard.v1`) instead of real product names.

artdroz (changes requested):
## Request changes: structured-response validation

Reviewed head `234b4c6badb158759c444a67f7c7386c2c29c45a` (unchanged from the initial spot-check). Re-read the full diff and rechecked the validator findings against the request model, extraction logic, tests, PR discussion, and Dynamic Agents README/architecture/SSE documentation. I found no documented supported-schema subset excluding these examples. The existing extractor accepts each object-root schema below. This review is limited to the two confirmed validator findings.

### 1. P2 — Accepted nullable property schemas raise TypeError on valid values

**Location:** [structured_response.py:69–70](https://github.com/caipe-io/ai-platform-engineering/blob/234b4c6badb158759c444a67f7c7386c2c29c45a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/structured_response.py#L69-L70).

Concrete request schema and submission:
```json
{"type":"object","required":["summary"],"properties":{"summary":{"type":["string","null"]}}}
```
```json
{"payload":{"summary":"Healthy"}}
```
`extract_response_format` accepts this schema, but `_JSON_TYPE_NAMES.get(expected_type)` receives a list and raises `TypeError` (unhashable list). This is a standard JSON Schema nullable type declaration, and the submitted string is valid. The exception occurs before `on_submit`, so the tool cannot capture the valid output or return its normal validation result. This does not establish that the entire HTTP request necessarily fails; runtime tool-error handling can intercept the exception, but the requested structured submission still fails.

**Required correction:** Validate supported union types without using the raw list as a dictionary key, preferably through a JSON Schema validator. If the intended API deliberately supports a narrower subset, document and enforce it when accepting the response format, returning an explicit unsupported-schema error before tool execution rather than accepting a schema that later crashes.

**Meaningful tests:** Through `create_submit_structured_response_tool(...).invoke(...)`, submit both a string and null for `type: ["string", "null"]` and assert acceptance plus the exact captured value; submit an integer and assert rejection without capture. Also exercise response-format extraction/validation for unsupported schema constructs if a restricted subset is chosen.

### 2. P2 — Nonconforming payloads are accepted and delivered as validated output

**Location:** [structured_response.py:65–81](https://github.com/caipe-io/ai-platform-engineering/blob/234b4c6badb158759c444a67f7c7386c2c29c45a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/structured_response.py#L65-L81); acceptance/capture occurs at [lines 96–108](https://github.com/caipe-io/ai-platform-engineering/blob/234b4c6badb158759c444a67f7c7386c2c29c45a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/structured_response.py#L96-L108).

Two concrete examples:
- Schema `{"type":"object","required":["summary"],"properties":{"summary":{"type":"string"}}}` accepts payload `{"summary":null}`. Presence satisfies `required`, and the explicit `None` skip bypasses the non-nullable string type constraint.
- Schema `{"type":"object","required":["items"],"properties":{"items":{"type":"array","items":{"type":"integer"}}}}` accepts payload `{"items":["wrong"]}`. Only the outer list type is checked; item constraints are ignored.

Both return `None` (validation success) from the actual validator. Consequently the submission tool reports `accepted: true`, captures the invalid payload, and makes it available to the invoke/custom-SSE output paths. Apps relying on a string or integer-array contract receive incompatible values rather than a rejection that lets the model correct its answer.

**Required correction:** Enforce nullability and the constraints of every accepted schema, including nested object/array validation. A JSON Schema validator is preferable to expanding a shallow type map. A deliberately limited schema API must explicitly reject unsupported constraints instead of silently accepting and ignoring them. Align the system instruction at [lines 119–121](https://github.com/caipe-io/ai-platform-engineering/blob/234b4c6badb158759c444a67f7c7386c2c29c45a/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/structured_response.py#L119-L121) with schema nullability: unavailable data must not authorize null where the schema disallows it.

**Meaningful tests:** Assert `accepted: false` and no callback invocation for non-nullable null, an invalid array item, and a nested object missing a required property. Include conforming nested payloads and explicitly nullable values as positive controls. Cover an invalid submission followed by a corrected submission so only the corrected payload is captured/delivered. The current tests cover a simple top-level string/type mismatch but do not exercise these cases.

### Verification and CI (separate from code blockers)

Reproduced the three examples by fetching the source at this exact SHA with `gh api` and executing the unchanged extraction/validation functions in memory, excluding the unrelated LangChain import. This was a focused validator check, not a full runtime test-suite run.

The failed [Dynamic Agents prebuild run](https://github.com/caipe-io/ai-platform-engineering/actions/runs/35607983652) was investigated: both architecture builds reached registry publishing and failed with `denied: installation not allowed to Write organization package`. The manifest job then downloaded zero digest artifacts and failed because its `prebuild-digests` working directory did not exist. These are registry-permission and cascading publish failures, not code blockers attributed to this PR. Fix the publishing permissions and rerun separately. Reported lint and unit-test checks passed.

Both original validator findings are retained; no additional findings from other reviewers are asserted here. No repository files were edited.

PR #2469, opened 2026-08-24
fix(helm): harden workload defaults

kevkantes:
# Description

Hardens the Helm defaults surfaced by admission-policy checks and removes the values-type collision around additional resources.

- Replace the list-based `extraDeploy` contract with the stable-key `extraDeployMap` contract. This is an intentional breaking change with no compatibility shim.
- Add inherited CPU/memory defaults for MCP servers, OpenFGA hooks, Keycloak reconciliation jobs, and the Milvus components that previously rendered without limits.
- Disable Kubernetes API token automount by default where workloads do not use the API, and add least-privilege security contexts for Keycloak, Milvus, etcd, Neo4j, and the skill scanner.
- Refresh generated chart documentation for the current 1.1.0 release.

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [x] Breaking Change
- [ ] Refactor
- [x] Documentation
- [ ] Other (please describe)

## Pre-release Helm Charts (Optional)

This PR will use the `helm-prerelease` label so downstream profiles can validate the published chart before merge.

## Validation

- `helm lint charts/rag-stack`
- `helm lint charts/ai-platform-engineering`
- Full `helm template` renders for two map-based deployment profiles
- `npm run build` in `docs/`
- `git diff --check`

## Checklist

- [x] I have read the [contributing guidelines](CONTRIBUTING.md)
- [ ] Existing issues have been referenced (where applicable)
- [x] I have verified this change is not present in other open pull requests
- [x] Functionality is documented
- [x] All code style checks pass
- [ ] New code contribution is covered by automated tests
- [ ] All new and existing tests pass

PR #2745, opened 2026-09-14
fix(helm): restore extraDeploy compatibility

sriaradhyula:
## Summary
- restore rendering of the legacy `extraDeploy` list removed by #2469
- retain `extraDeployMap` as the preferred stable-key interface
- document the legacy field as deprecated and add CI coverage proving both forms render together

This prevents existing deployment values from silently dropping custom Kubernetes resources during a chart upgrade.

## Release note
The published 1.1.0 artifact is immutable. This fix should be included in a 1.1.1 patch release for existing 1.1 consumers; `main` is currently on the 1.2.0 development line.

## Validation
- rendered `templates/extra-deploy.yaml` with both `extraDeploy` and `extraDeployMap` and verified both ConfigMaps
- `git diff --check`

Related deployment mitigation: cisco-eti/platform-apps-deployment#1166


.github/workflows/helm-chart-test.yml
```diff
@@ -110,6 +110,30 @@ jobs:
           --set tags.basic=true \
           --set tags.rag-stack=false
 
+    - name: Test extraDeploy backward compatibility
```
kevkantes, line 113: extraDeploy is reserved by another chart we import so it causes issues. Imo, you should remove the backwards compat

sriaradhyula: @kevkantes - `extraDeploy` is convention. We have been using it. We can support both and keep `extraDeploy` as optional and `extraDeployMap` to avoid your collision, but I would retain backward compatibility as it would definitely break others using it.

kevkantes:
```
coalesce.go:237: warning: skipped value for ai-platform-engineering.extraDeploy: Not a table.
coalesce.go:237: warning: skipped value for etcd.extraDeploy: Not a table.
coalesce.go:237: warning: skipped value for ai-platform-engineering.extraDeploy: Not a table.
```

these are the warnings we get because of the etcd conflict. we should consider dropping support for extraDeploy in 1.3 and only use extraDeployMap

kevkantes (approved):
Approved in case this is blocking someone, but please be sure to have a plan for this in a later version

PR #2746, opened 2026-09-15
fix(helm): restore Milvus image compatibility

sriaradhyula:
## Summary

- stop forcing UID/GID 1000 on the stock Milvus image
- retain filesystem ownership, privilege-escalation, capability, and seccomp hardening
- add a regression test for the image-compatible security context

## Root cause

Milvus v2.6.1 ships /milvus/bin/milvus and /milvus/lib/libjemalloc.so as root:root with mode 0774. The chart's forced UID/GID 1000 cannot execute or preload those files, causing exit code 126 with 'exec milvus failed: Permission denied'.

## Validation

- python3 -m pytest tests/test_milvus_helm_security_context.py -q
- helm lint charts/rag-stack
- rendered the rag-stack chart and verified all five Milvus deployments omit forced identity fields while retaining the other security controls

The published 1.1.0 chart is affected. Please include this fix in the next 1.1.x patch release.


charts/rag-stack/values.yaml
```diff
@@ -280,17 +280,15 @@ milvus:
     annotations: {}
 
   securityContext:
-    runAsNonRoot: true
-    runAsUser: 1000
-    runAsGroup: 1000
+    # The stock Milvus image owns its executable and preload library as root:root
```
kevkantes, line 283: There's no way to customize this? Internally this flags as a security issue

kevkantes: Also, I don't think you need this comment if you're removing the runAs

sriaradhyula: Can you pass this security context in your values?

kevkantes: I think the upstream chart should be secure by default, and fix forward instead of reverting to admin. but if you want, you can just remove it for now and fix it later. 

I would just remove this comment

tests/test_milvus_helm_security_context.py
```diff
@@ -0,0 +1,28 @@
...
+
+import yaml
+
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+RAG_STACK_VALUES = REPO_ROOT / "charts" / "rag-stack" / "values.yaml"
+FORCED_IDENTITY_FIELDS = {"runAsNonRoot", "runAsUser", "runAsGroup"}
+
+
+def test_stock_milvus_image_does_not_force_incompatible_non_root_identity():
```
kevkantes, line 14: This test is just doing a yaml check?

charts/rag-stack/values.yaml
sriaradhyula: @kevkantes  Good point. I checked the current upstream release, and we should consider upgrading to Helm chart `5.0.28` (Milvus app version `3.0.1`) as the fix-forward path.

Milvus 3.0.1 improves this at the image level: it creates a `milvus:milvus` user, changes ownership of the required paths, and sets `USER milvus:milvus`. However, that user is created with `useradd -r`, so its numeric UID/GID is not guaranteed to be `1000`. Keeping `runAsUser: 1000` and `runAsGroup: 1000` would override the image user and could reproduce the permission failure.

For the 3.0.1 upgrade, we should therefore remove the hard-coded 1000 values, verify the image's actual numeric UID/GID if Kubernetes `runAsNonRoot: true` requires us to pin them, and retain the other hardening controls. Since this is a Milvus 2.6 to 3.0 major upgrade for a stateful component, I think it should be validated separately rather than folded into the 1.1.x compatibility patch.

kevkantes:
This was fixed in main, you can delete this branch/PR

PR #2759, opened 2026-09-16
fix(release): keep prereleases out of stable changelog

sriaradhyula:
## Description

Keep prerelease builds out of the stable changelog and prevent RCs from being published as full releases.

- Generate CHANGELOG.md entries only for stable semantic versions.
- Keep RC workflows build-only and out of the changelog and GitHub Release flow.
- Mark manually created RC releases as draft prereleases.
- Restrict documentation release posts to stable versions.
- Remove existing dev and RC headings from the current CHANGELOG.md.
- Add a dedicated release-policy workflow and regression script.

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation

## Validation

- `bash tests/integration/test_release_workflow_policy.sh`
- `git diff --check`
- Stable and prerelease workflow paths are covered by regression assertions
- DCO trailers included on all commits


tests/integration/test_release_workflow_policy.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
```
kevkantes: Do we really need a test that just checks for the word prerelease? it's testing for a negative condition that can't happen anymore in the above workflow. 

Imo this is contributing to the massive test bloat that we don't really understand.

sriaradhyula: Addressed in 0f5a2b338: removed the redundant release-policy test script and its dedicated workflow, since the grep-based assertions were not providing useful coverage.

kevkantes (approved):

PR #2760, opened 2026-09-16
fix(setup): harden no-ingress installer path

sriaradhyula:
## Description

Fix the CAIPE installer path for release-tag installs and explicit no-ingress SSH port-forward deployments.

- Preserve user-provided LLM endpoint and model overrides.
- Avoid unnecessary static AgentGateway CRDs unless Gateway API mode is selected.
- Propagate Gateway API CRD installation failures.
- Bound auxiliary namespace cleanup waits.
- Split browser-facing and in-cluster OIDC endpoints for no-ingress installs.
- Provision the local browser login and document the port-forward OIDC setup.
- Accept the shared `caipe-platform` audience used by interactive access tokens;
  keep `caipe-ui` as the browser OIDC client identifier.
- Enable the workflow-runner UI capability by default while preserving an
  explicit operator override.

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [ ] Documentation

## Validation

- `bash -n setup-caipe.sh`
- `bash tests/integration/test_setup_caipe_installer_regressions.sh`
- Dynamic-agent OBO and auth-config regression coverage added
- Standalone `caipe-ui` chart lint remains blocked by the existing `vpa.yaml`
  nil-map error when linted without the parent chart.
- DCO trailers included on all commits


setup-caipe.sh
```diff
@@ -4784,7 +4803,7 @@ JSON
 # that cause "Invalid parameter: redirect_uri" on login.
 update_keycloak_client_urls() {
   $ENABLE_RBAC_RUNTIME || return 0
-  [[ -n "${CAIPE_DOMAIN:-}" ]] || return 0
+  [[ -n "${CAIPE_DOMAIN:-}" || ! $ENABLE_INGRESS ]] || return 0
```
cisco-erilutz, line 4806: **Confirmed bug — breaks the no-ingress path this PR targets.**

`[[ -n "${CAIPE_DOMAIN:-}" || ! $ENABLE_INGRESS ]] || return 0`

Inside `[[ ]]`, a bare `$ENABLE_INGRESS` (literal string `"true"`/`"false"`) is a non-empty-string test, not a boolean — so `! $ENABLE_INGRESS` is always `false` regardless of the variable's value. The condition collapses to just `-n "${CAIPE_DOMAIN:-}"`.

Result: in `--port-forward-mode`/`--no-ingress` with no `CAIPE_DOMAIN` set, this returns early and never patches the `caipe-ui`/`caipe-platform` Keycloak redirect URIs to `http://localhost:$UI_PORT`, so browser login over port-forward fails with `invalid_redirect_uri` — the exact scenario this PR adds `_browser_ui_url()` to support.

Compare with the correct pattern used elsewhere in this same diff (outside `[[ ]]`, where `$ENABLE_INGRESS` is invoked as a command): `if [[ -z "$da_oidc_issuer" ]] && ! $ENABLE_INGRESS && ! _upstream_idp_configured; then`. Suggest: `[[ -n "${CAIPE_DOMAIN:-}" ]] || ! $ENABLE_INGRESS || return 0`.

setup-caipe.sh
```diff
@@ -4848,7 +4868,7 @@ update_keycloak_client_urls() {
 # BEARER_AUDIENCE_MISMATCH.
 provision_caipe_ui_audience_mapper() {
   $ENABLE_RBAC_RUNTIME || return 0
-  [[ -n "${CAIPE_DOMAIN:-}" ]] || return 0
+  [[ -n "${CAIPE_DOMAIN:-}" || ! $ENABLE_INGRESS ]] || return 0
```
cisco-erilutz, line 4871: **Confirmed bug — same guard pattern as line 4806, same fix needed.**

`[[ -n "${CAIPE_DOMAIN:-}" || ! $ENABLE_INGRESS ]] || return 0`

Same `! $ENABLE_INGRESS` inside `[[ ]]` issue: it's always false, so this function returns early whenever `CAIPE_DOMAIN` is unset, even in no-ingress mode. That means the `aud=caipe-ui` audience mapper never gets provisioned for `--port-forward-mode` installs, and the dynamic-agents streaming endpoint will 401 with `BEARER_AUDIENCE_MISMATCH` for that topology.

setup-caipe.sh
```diff
@@ -6481,12 +6592,26 @@ dynamic-agents:
...
+    if [[ -n "$da_oidc_issuer" ]]; then
+      local _da_cors_origin
+      _da_cors_origin="$(_browser_ui_url)"
       cat >> "$_da_values_file" <<DAEOF
     AUTH_ENABLED: "true"
     OIDC_ISSUER: "${da_oidc_issuer}"
+    OIDC_DISCOVERY_URL: "$(_internal_oidc_issuer)"
     OIDC_CLIENT_ID: "${da_oidc_client_id}"
-    CORS_ORIGINS: '["https://${CAIPE_DOMAIN}", "http://localhost:3000"]'
+    CORS_ORIGINS: '["${_da_cors_origin}"]'
```
cisco-erilutz, line 6614: **Potentially breaking, even for local testing setups.**

`CORS_ORIGINS: '["${_da_cors_origin}"]'` replaces the previous `'["https://${CAIPE_DOMAIN}", "http://localhost:3000"]'`. `_da_cors_origin` is `_browser_ui_url()`, which in ingress mode (`$ENABLE_INGRESS` true + `CAIPE_DOMAIN` set) returns only `https://$CAIPE_DOMAIN` — the `http://localhost:3000` origin that was previously always included is now dropped unconditionally for any ingress install.

This silently breaks a common local dev pattern: ingress-mode deploys reached directly via `kubectl port-forward` to `localhost:3000` (bypassing the ingress domain) will now get CORS-rejected by dynamic-agents, even though this worked before. Worth confirming this is intentional, and if so, calling it out in the PR description/changelog as a behavior change for anyone doing local ingress-mode testing via port-forward.

setup-caipe.sh
```diff
@@ -8808,13 +8989,20 @@ BANNER
...
     for _v in "${_llm_vars[@]}"; do
       local _val
       _val=$(_env_get "$ENV_FILE" "$_v")
-      [[ -n "$_val" && -z "${!_v:-}" ]] && export "$_v=$_val"
+      if [[ "$_v" == OPENAI_ENDPOINT ]]; then
+        [[ -n "$_val" && -z "${_OPENAI_ENDPOINT_EXPLICIT:-}" ]] && export "$_v=$_val"
+      elif [[ "$_v" == OPENAI_MODEL_NAME ]]; then
+        [[ -n "$_val" && -z "${_OPENAI_MODEL_NAME_EXPLICIT:-}" ]] && export "$_v=$_val"
+      else
+        [[ -n "$_val" && -z "${!_v:-}" ]] && export "$_v=$_val"
```
cisco-erilutz, line 9004: This generic `--env-file` override branch (`[[ -n "$_val" && -z "${!_v:-}" ]] && export "$_v=$_val"`) still applies to `ANTHROPIC_MODEL_NAME` and `AWS_BEDROCK_MODEL_ID`. Both already get non-empty defaults earlier in the script (`ANTHROPIC_MODEL_NAME="${ANTHROPIC_MODEL_NAME:-claude-haiku-4-5-20251001}"`, similarly for `AWS_BEDROCK_MODEL_ID`), so `${!_v:-}` is never empty by the time this loop runs — meaning an `--env-file` value for either is silently ignored, exactly the class of bug this PR just fixed for `OPENAI_ENDPOINT`/`OPENAI_MODEL_NAME` via the new `_EXPLICIT` flags. Might be worth the same `_EXPLICIT`-flag treatment (or dropping the premature default) for these two as well, unless it's intentionally out of scope for this PR.

ai_platform_engineering/dynamic_agents/src/dynamic_agents/auth/obo_exchange.py
```diff
@@ -72,9 +73,6 @@ class _CacheEntry:
 
 def _resolve_token_endpoint() -> Optional[str]:
     """Resolve the Keycloak token endpoint from issuer or discovery URL."""
-    issuer = os.environ.get("OIDC_ISSUER", "").strip()
-    if issuer:
-        return f"{issuer.rstrip('/')}/protocol/openid-connect/token"
     discovery = os.environ.get("OIDC_DISCOVERY_URL", "").strip()
```
cisco-erilutz, line 76: This reorders `_resolve_token_endpoint()` to check `OIDC_DISCOVERY_URL` before `OIDC_ISSUER` (previously `OIDC_ISSUER` was checked first). For any existing deployment that sets both env vars (e.g. to different hosts, such as an internal vs. browser-facing issuer in ingress mode), this silently changes which endpoint OBO token exchange targets. If that's intentional given the new `_internal_oidc_issuer()`/`_browser_oidc_issuer()` split introduced in this PR, might be worth a line noting the precedence change explicitly (beyond the docstring), since it's a behavior change for anyone already relying on the old precedence.

cisco-erilutz (commented):
Automated review pass. Two confirmed correctness bugs that break the no-ingress/port-forward path this PR is meant to harden, plus a couple of plausible breaking changes worth a second look before merge (flagging per request, especially the CORS localhost removal for local testing setups). Inline comments below; verified against the head commit (`9c72f61e2f4`).

Note: `tests/integration/test_setup_caipe_installer_regressions.sh` doesn't actually exercise whether `update_keycloak_client_urls`/`provision_caipe_ui_audience_mapper` execute their bodies under `--port-forward-mode` — it only checks for unrelated strings, which is how the guard bug below shipped without a failing test.

PR #2774, opened 2026-09-18
fix(auth): prevent OIDC refresh retry loops

sriaradhyula:
# Description

Production showed caipe-ui repeatedly retrying a Keycloak refresh token that had already become inactive. Server-side refreshes also followed the public token_endpoint advertised by OIDC discovery, sending pod traffic through a public WAF. Rate-limited HTML 403 responses were treated as expired refresh tokens and could put the browser into a sign-out and OAuth callback loop.

This change:

- routes server-side refreshes to OIDC_TOKEN_ENDPOINT when configured, otherwise derives the in-cluster Keycloak token endpoint from OIDC_DISCOVERY_URL
- adds the internal token endpoint to Helm and Docker deployment configuration
- marks invalid_grant and invalid_token as terminal so the same refresh token is never retried
- preserves a still-valid access token after terminal or transient refresh failures and applies bounded backoff for transient WAF, network, and provider failures
- prevents force-refresh keepalive calls for failed sessions
- makes failed-session sign-out idempotent across rerenders

OIDC_ISSUER remains browser-facing for redirects and issuer validation. No production resources are changed by this PR.

## Type of Change

- [x] Bugfix
- [ ] New Feature
- [ ] Breaking Change
- [ ] Refactor
- [x] Documentation
- [ ] Other

## Verification

- npm test -- --runInBand src/lib/__tests__/auth-config.test.ts src/components/__tests__/token-expiry-guard.test.tsx src/components/__tests__/auth-guard.test.tsx (132 passed)
- npx tsc --noEmit --pretty false
- npx eslint on all changed TypeScript and TSX files
- helm template for the caipe-ui subchart with required parent-value stubs
- git diff --check

## Checklist

- [x] Related incident and mirror PR referenced
- [x] Verified no existing PR for this branch
- [x] Functionality documented
- [x] Code style checks pass
- [x] New behavior is covered by automated tests
- [x] Focused existing and new tests pass
- [x] Selection-control checklist item is not applicable


cisco-erilutz (approved):
Reviewed via three parallel focused passes (core refresh logic in auth-config.ts, auth-guard/token-expiry-guard component lifecycle, and Helm/Docker Compose config wiring + test coverage). No correctness bugs found.

Minor non-blocking notes for a future follow-up:
- "Bounded backoff" for transient failures is a flat 60s cooldown rather than incremental backoff — functionally safe, just worth adjusting the description wording.
- The success-path type cast in `refreshAccessToken` doesn't validate `access_token` presence, but the existing `AccessTokenMissing` guard in the `session()` callback catches this downstream.
- No test drives a transient failure past its suppression window to verify a second cooldown cycle renews correctly.
- The new Helm `OIDC_TOKEN_ENDPOINT` default template value has no automated snapshot/unit test coverage, only the manual `helm template` step.

None of these block merge. LGTM.

PR #2787, opened 2026-09-20
fix(dynamic-agents): isolate interrupt state runtime

sriaradhyula:
# Description

The interrupt-state probe (GET conversation interrupt state) only needs
the durable LangGraph checkpoint to answer "is there a pending interrupt?",
but it was routed through the shared runtime cache via get_or_create.
Caching a runtime under the conversation's key from this read-only probe
risks poisoning that cache entry: if MCP server initialization degrades or
fails closed during the probe (e.g. a transient connection error, or
credentials not yet available in that request's context), the next real
chat call for the same conversation reuses the degraded, cached runtime
instead of getting a fresh one.

This switches the endpoint to cache.persistent(), a non-cached one-shot
runtime that still reads the same durable checkpoint but is cleaned up
immediately after the request and can never be reused by a later chat call.

## Type of Change

- [x] Bugfix

## Checklist


artdroz (changes requested):
## Required fix: [P2] Keep interrupt-state probes from deleting the live conversation skill filesystem

**Changed location:** [src/dynamic_agents/routes/conversations.py:122–127](https://github.com/caipe-io/ai-platform-engineering/blob/b1202704291a93e108f204ed3c64c0aa1d95b804/ai_platform_engineering/dynamic_agents/src/dynamic_agents/routes/conversations.py#L122-L127).

The cache-isolation goal is sound, but `cache.persistent()` is not a read-only runtime constructor. It unconditionally calls `initialize()` ([runtime_cache.py:278–299](https://github.com/caipe-io/ai-platform-engineering/blob/b1202704291a93e108f204ed3c64c0aa1d95b804/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/runtime_cache.py#L278-L299)). For a StoreBackend agent, initialization deletes every `/skills/` entry before loading replacement skills ([agent_runtime.py:1002–1042](https://github.com/caipe-io/ai-platform-engineering/blob/b1202704291a93e108f204ed3c64c0aa1d95b804/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py#L1002-L1042)). The default filesystem namespace is `(agent_id, session_id, "filesystem")`, so this one-shot runtime shares the existing chat runtime filesystem; creating a separate runtime does not isolate that storage ([namespace resolution](https://github.com/caipe-io/ai-platform-engineering/blob/b1202704291a93e108f204ed3c64c0aa1d95b804/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py#L789-L799)). `MongoDBGridFSStore.delete_by_key_prefix()` performs actual GridFS deletions, not local cache invalidation.

**Concrete scenario and impact:** A StoreBackend conversation has a healthy, non-stale cached chat runtime and populated skill files. Refreshing/opening its interrupt state now initializes another runtime for the same agent/conversation. The GridFS deletion succeeds, but a subsequent skill-catalog query fails transiently. `load_skills()` catches query errors and returns an empty list ([skills.py, load_skills](https://github.com/caipe-io/ai-platform-engineering/blob/b1202704291a93e108f204ed3c64c0aa1d95b804/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/skills.py#L263-L327)); `initialize()` then skips reseeding. Exceptions during skill building/seeding are also caught and logged without rollback ([agent_runtime.py:1054–1057](https://github.com/caipe-io/ai-platform-engineering/blob/b1202704291a93e108f204ed3c64c0aa1d95b804/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py#L1054-L1057)). Assuming the remaining graph initialization/checkpoint read succeeds, the GET can return normally while leaving the existing runtime without its persisted skill files. Later skill reads in that chat fail or lose the expected instructions. Cleanup releases runtime resources; it does not restore files. Previously, a valid cache hit skipped initialization and therefore avoided this destructive refresh. Cold-cache initialization already had this behavior; this change extends it to every probe, including conversations with healthy cached runtimes.

**Correction:** Read the durable checkpoint through a reader that does not run mutating agent initialization, or introduce a genuinely read-only initialization path that skips skill deletion/seeding and other filesystem writes. Preserve the configured checkpoint collections, conversation thread ID, response behavior, and cache isolation. Merely catching initialization errors or cleaning up the one-shot runtime does not undo the deletion.

**Regression expectation:** Exercise the real probe/reader path with a StoreBackend conversation, pre-existing skill files, and an existing healthy cached runtime. Inject a skill-catalog query failure (covering the loader empty-result behavior), and verify the probe preserves the original files, performs no skill deletion/seeding, leaves the cached runtime usable, and still reads the durable interrupt state. Cover both pending and absent interrupts. The current test replaces `persistent()` with a fake context and cannot detect initialization side effects.

**Evidence level:** Verified by tracing the implementation and its error handling at this commit; no live MongoDB failure or end-to-end reproduction was run. This is the sole blocking finding. No CI failure is being attributed to this change.

cisco-erilutz (approved):
Reviewed the diff: switches the GET interrupt-state probe from the shared cached runtime (`cache.get_or_create()`) to the already-existing one-shot `cache.persistent()` path, so a degraded probe-time runtime can never be reused by a later chat call on the same conversation. This bypasses the shared cache/pending state entirely rather than patching around a race, which structurally eliminates the cache-poisoning risk. MongoClient lifecycle and cleanup on early-return are correct. New unit test verifies `persistent()` is called instead of `get_or_create()`; no separate concurrency test needed since the fix removes the shared state the race depended on. LGTM.

sriaradhyula:
Fixed — thanks for the thorough trace, this was a real bug.

`AgentRuntime` now takes a `readonly: bool = False` flag. When set, `initialize()` skips the two writes that mutate the shared GridFS skill namespace (the stale-skill `delete_by_key_prefix` clear and the reseed `put` loop); everything else in initialization is unchanged. `RuntimeCache.reader()` is a new context manager alongside `persistent()`/`ephemeral()` that uses the real Mongo-backed checkpointer (so durable interrupt state is still visible) but constructs the runtime with `readonly=True`. `persistent()` itself is untouched — scheduled/webhook runs that need real skill files still get them — only the interrupt-state route now calls `reader()` instead.

Added `test_agent_runtime_readonly_skills.py`, which exercises the real `initialize()` skill pipeline (not a fake context, per your note) and asserts: a readonly runtime never calls `delete_by_key_prefix`/`put` on the store, and a normal runtime still does (control case, confirming the guard is specific to `readonly` and not a no-op everywhere).

PR #2788, opened 2026-09-20
fix(dynamic-agents): refresh cached runtime client context on reuse

sriaradhyula:
# Description

Depends on #2384 — this stacks on top of that branch's trusted-interaction
plumbing and should merge after it (or be rebased onto main once #2384
lands).

AgentRuntimeCache.get_or_create() reuses one cached runtime across requests
for the same conversation, but each request's client_context carries a
fresh, short-lived trusted-interaction token/signature. On a cache hit, the
cached runtime kept the client_context captured when it was first created,
so _trusted_interaction_headers() and the credential-exchange client kept
forwarding the token/signature from the first request instead of the
current one — meaning the proof goes stale after the very first message in
a conversation.

This adds AgentRuntime.refresh_client_context() and calls it from both the
cache-hit path and the pending single-flight-waiter path in
RuntimeCache.get_or_create(), so every caller gets a runtime whose client
context matches its own request.

## Type of Change

- [x] Bugfix

## Checklist


artdroz (changes requested):
## Required fix: P1 — Refresh the proof used by actual cached MCP tool calls

Reviewed head: `aacd8654f96a67c80da72cbec122d8f6dfb80a14` (rechecked before posting; unchanged). This review is limited to #2788.

**Location:** [agent_runtime.py:835–844](https://github.com/caipe-io/ai-platform-engineering/blob/aacd8654f96a67c80da72cbec122d8f6dfb80a14/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py#L835-L844), called by [runtime_cache.py:192–201](https://github.com/caipe-io/ai-platform-engineering/blob/aacd8654f96a67c80da72cbec122d8f6dfb80a14/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/runtime_cache.py#L192-L201).

Replacing `_client_context` updates the accessor, but does not update the connection configuration retained by the already-built MCP tools. Consequently, the stale-proof failure this PR intends to fix remains on the outbound tool-call path.

### Verified call path
1. `AgentRuntime.initialize()` takes a snapshot from `_trusted_interaction_headers()`, builds MCP connections, resolves credentials, and loads tools: [agent_runtime.py:869–910](https://github.com/caipe-io/ai-platform-engineering/blob/aacd8654f96a67c80da72cbec122d8f6dfb80a14/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py#L869-L910). Initialization returns early once `_initialized` is true.
2. `build_mcp_connection_config()` copies the token/signature strings into `X-CAIPE-Trusted-Interaction` and `X-CAIPE-Trusted-Interaction-Signature`: [mcp_client.py:262–275](https://github.com/caipe-io/ai-platform-engineering/blob/aacd8654f96a67c80da72cbec122d8f6dfb80a14/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/mcp_client.py#L262-L275). The runtime attribute is not a live reference for these values.
3. `get_tools_with_resilience()` passes that connection configuration to `MultiServerMCPClient.get_tools()`: [mcp_client.py:974–979](https://github.com/caipe-io/ai-platform-engineering/blob/aacd8654f96a67c80da72cbec122d8f6dfb80a14/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/mcp_client.py#L974-L979). I checked the pinned `langchain-mcp-adapters==0.2.1` implementation: [client.py](https://github.com/langchain-ai/langchain-mcp-adapters/blob/5f17d168729e1d8118c47cbf5c92c3d4408714ff/langchain_mcp_adapters/client.py) passes the connection to `load_mcp_tools`; [tools.py](https://github.com/langchain-ai/langchain-mcp-adapters/blob/5f17d168729e1d8118c47cbf5c92c3d4408714ff/langchain_mcp_adapters/tools.py) captures it in the tool coroutine and calls `create_session(effective_connection)` on each invocation; [sessions.py](https://github.com/langchain-ai/langchain-mcp-adapters/blob/5f17d168729e1d8118c47cbf5c92c3d4408714ff/langchain_mcp_adapters/sessions.py) forwards those headers and the factory to the HTTP/SSE transport. No tool interceptor is supplied here to replace the proof.
4. The custom HTTP factory copies the supplied headers and refreshes **agent-context** headers, but leaves **trusted-interaction** headers intact: [mcp_client.py:216–239](https://github.com/caipe-io/ai-platform-engineering/blob/aacd8654f96a67c80da72cbec122d8f6dfb80a14/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/mcp_client.py#L216-L239). Opening a fresh transport session therefore does not repair the stale proof. Subagent tools use the same snapshot pattern: [agent_runtime.py:1494–1507](https://github.com/caipe-io/ai-platform-engineering/blob/aacd8654f96a67c80da72cbec122d8f6dfb80a14/ai_platform_engineering/dynamic_agents/src/dynamic_agents/services/agent_runtime.py#L1494-L1507).

### Concrete scenario and impact
Initialize a cached conversation with valid proof A and successfully load a private MCP tool. Let A expire while keeping the runtime alive, then send a new message for the same agent/conversation with valid proof B. The cache hit assigns B to `_client_context`, but invoking the already-loaded tool still creates its HTTP/SSE session with A. Private MCP authorization consequently continues to fail despite the fresh request proof. This is an incomplete fix of the existing failure, not a claim that this PR newly introduced the original stale-header behavior.

**Evidence level:** confirmed by static tracing of this head and the exact pinned adapter implementation. I did not execute an end-to-end tool invocation or reproduce a live gateway denial; the denial described above is the expected consequence when the gateway rejects expired proofs.

### Required correction
Make the trusted-interaction token and signature used by each actual outbound MCP invocation come from that invocation’s current request context, including cached parent and subagent tools. A request-scoped transport/header provider or equivalent correctly scoped invocation mechanism can do this; replacing only the runtime field cannot. Remove stale captured trusted-interaction headers when the current request has no proof. Preserve request isolation if two callers overlap on the same cached runtime rather than making authorization depend on whichever caller last assigned a shared field.

Also carry the supplied resume proof through to the runtime/invocation: [_generate_resume_sse_events, chat.py:418–423](https://github.com/caipe-io/ai-platform-engineering/blob/aacd8654f96a67c80da72cbec122d8f6dfb80a14/ai_platform_engineering/dynamic_agents/src/dynamic_agents/routes/chat.py#L418-L423) calls `get_or_create()` without `client_context`, so the new cache-hit refresh receives `None` on resume. This is a related completeness requirement for the correction, not a separately reproduced outbound-header regression.

The credential-exchange wording should also reflect its actual lifecycle: the client and credential resolution shown above are created/run during initialization, not rerun by this setter. If refreshing credential resolution is part of the intended guarantee, implement and test that path explicitly.

### Required validation
Add a regression test that builds/loads an MCP tool using proof A, reuses the cached runtime with proof B, and **invokes the already-loaded tool** through the real adapter/transport setup. Capture the outbound HTTP headers with a local fake MCP endpoint or an HTTP transport test double; assert both trusted-interaction headers are B and neither is A. Do not mock away the connection capture or assert only `_trusted_interaction_headers()`/`refresh_client_context()` calls.

Cover parent and subagent tools, ordinary cache hits and genuinely pending initialization waiters, and resume with a fresh supplied proof. Include a no-proof request to ensure stale proof removal and overlapping requests to verify each invocation keeps its own proof. An expiry-aware fake gateway can additionally show A rejected and B accepted without a live deployment.

The three added tests currently verify the accessor or that a mock refresh method was called; they cannot detect this actual invocation-path failure. No other blocking finding is asserted in this review.

PR #2789, opened 2026-09-20
fix(audit): bound history query memory and concurrency

sriaradhyula:
# Description

Broad audit-history reads (GET /v1/audit/events) retained every matching
event in memory before sorting and slicing to the response limit. A wide
date range or low-selectivity filter could scan far more events than are
ever returned, holding the whole scan in memory for the life of the
request. S3 queries additionally submitted every matching object to the
thread pool at once via executor.map, so a slow object left every other
decoded result queued in memory for the entire range, and the response
body stream was never explicitly closed after reading.

- _QueryMatches retains only the newest `limit` records in a bounded
  min-heap while still counting every match for an exact `total`, so
  memory scales with the response size, not the scan size.
- S3AuditStore.query submits objects to the thread pool in windows of
  `_S3_IO_MAX_WORKERS` instead of all at once, and closes each response
  body after reading it.
- AUDIT_SERVICE_READ_CONCURRENCY (default 2) bounds how many history
  scans can run concurrently per process via a semaphore; a cancelled
  caller retains its slot until the background scan actually finishes,
  so a disconnect can't bypass the limit. Ingest and health endpoints
  are unaffected by this limit.

Synthetic 6,000-event scan: peak traced allocations fall from ~26.7 MB
to ~34 KB for the same total/order/truncation result.

This supersedes #2727, which bundled the same audit fix into a much
larger Admin Insights refactor and was closed unmerged; this PR is just
the audit_service slice, narrowly scoped for review.

## Type of Change

- [x] Bugfix

## Checklist

- [x] All new and existing tests pass (one pre-existing, unrelated failure: `test_get_storage_local` has a hardcoded fixture date that has aged past its retention window on unmodified main too)


artdroz (approved):
Approving based on a scoped, lightweight spot-check of the full five-file diff and relevant implementation context at 6fb3d65e7819bf5852eeba9822291077d270014b. This is not an exhaustive review or full test-suite validation. The bounded heap and windowed S3 reads look sound in this scope. Current checks have no failures, but the main unit-test log collects utils tests, not the new audit-service tests.

Per the requester’s explicit decision, accepting the following reproduced concurrency edge as a non-blocking follow-up:

- ai_platform_engineering/audit_service/main.py:157–159: after the first cancellation, `await task` is unshielded. If the request task is cancelled a second time while the storage thread is still scanning, cancellation propagates to the storage task and the semaphore slot is released even though the thread continues. A waiting query can then start above the configured limit. An in-memory reproduction using this commit’s exact helper started two scans with a limit of one.
- Suggested correction: give a cancellation-protected worker ownership of the semaphore until the underlying storage operation actually finishes, so repeated caller cancellation cannot release it early.
- Meaningful regression test: with concurrency one, block the first storage scan on an event, cancel its caller, let it enter cancellation cleanup, cancel it again, and queue a second query. Assert the second storage scan cannot enter until the first thread is released and finishes; then assert the queued query completes and the slot is reusable. The current test covers only one cancellation.

Minor documentation follow-up: ai_platform_engineering/audit_service/README.md:12 claims 512-row Parquet decoding batches, while `_from_parquet_bytes` still uses whole-table `pq.read_table(...).to_pylist()`. Correct that claim or implement and verify batching.

cisco-erilutz (approved):
Reviewed the bounded-memory query heap, S3 windowed submission, and semaphore/cancellation slot-holding logic in detail — traced tie-break behavior, heap comparison safety (no dict comparisons reached, since -total is always unique), and the cancelled-reader-holds-slot semantics against the new regression tests.

Checked out the branch and ran `uv run pytest ai_platform_engineering/audit_service`: 52 passed, 1 failed (`test_get_storage_local`) — confirmed that failure also reproduces on unmodified `main`, matching the PR description's note that it's a pre-existing, unrelated fixture-date issue. All CI checks pass.

One very minor, non-blocking note: in `_query_with_slot`, if the shielded task raises a real exception at the same moment the caller is cancelled, the caller receives that exception instead of `CancelledError` — harmless in practice since the client has already disconnected by then.

LGTM.

cisco-erilutz (approved):
Reviewed the diff: bounds audit history query memory via a min-heap top-K instead of a full-list sort, windows S3 object prefetch to avoid unbounded buffering, closes S3 response bodies in `finally` (even on decode exceptions), and adds a semaphore (default 2) capping concurrent history scans without blocking ingest/health, with cancellation-safe slot release (shields and awaits the background thread before releasing). Heap tie-break logic correctly reproduces the original stable-sort behavior. Limits are clamped to >=1 everywhere, so no deadlock risk. Test coverage is strong: tracemalloc-verified bounded memory (<2MiB vs >24MiB unbounded), exact ordering/truncation with ties, S3 prefetch windowing, concurrent-read semaphore bound, and cancellation-holds-slot. LGTM.

PR #2802, opened 2026-09-23
feat(admin): add user impersonation

kevkantes:
# Description

Replace the admin-only read-only "View As" simulator with Keycloak-backed, application-enforced read-only user impersonation for troubleshooting.

- Start impersonation from the selected user's detail modal under **Teams & Users → Users**; remove the redundant standalone Security & Policy page, nav item, and target-search endpoint.
- Require the administrator's email in `ADMIN_ALLOW_IMPERSONATION` and require bootstrap or OpenFGA `super-admins` membership. An empty allowlist disables the feature.
- Restrict impersonation in brokered realms to enabled, identity-provider-linked human users who satisfy `OIDC_REQUIRED_GROUP`; Slack-only Keycloak shells are not eligible.
- Confirm the read-only scope before starting, and show a persistent warning banner identifying the effective user.
- Enforce read-only browser traffic centrally across API, page, and embedded-app mutations. Block connected-credential and connector-linking endpoints entirely while impersonating.
- Hide the developer access-token copy affordance while impersonating.
- Exchange and renew a target-user access token while retaining the original administrator identity server-side for audit attribution.
- Record the target as the subject and the administrator as the actor on Web UI backend audit events and impersonation start/stop events, with an explicit impersonation marker and start time.
- Exit by ending impersonation, deleting the target token, clearing identity-owned chat state, and forcing sign-out/fresh administrator authentication.
- Show the latest Keycloak browser-session activity in the Teams & Users list while retaining federated identity linkage as the impersonation eligibility check.
- Clear identity-owned chat, active-conversation, notification, pagination, and streaming state across login, logout, and impersonation transitions so one user's conversation cannot appear in another user's chat list.
- Invalidate in-flight chat loads across identity changes and only auto-resume conversations owned by the effective user; direct admin-audit links remain read-only.
- Route Azure GPT-5 reasoning requests through the Responses API when agents use function tools, and skip retries for terminal HTTP 4xx errors.
- Keep the New Chat button's server and initial client markup stable while its default agent resolves.
- Provision the required Keycloak token-exchange and impersonation permissions in Docker Compose and Helm first-install paths.
- Remove the old simulation query parameters, route overrides, UI controls, and helper modules.
- Update admin, RBAC, Compose, Helm, and generated chart documentation.

The read-only restriction is enforced by the application. The exchanged Keycloak token does not yet carry a cryptographic `act` delegation claim, so this PR does not claim identity-provider-enforced delegation boundaries.

This is larger than the usual 500-line target because the security-sensitive workflow needs to land atomically across session lifecycle, request enforcement, API identity propagation, UI escape paths, Keycloak bootstrap configuration, tests, and removal of the previous implementation. The diff also removes the simulator implementation and tests.

AI assistance disclosure: Codex assisted with implementation, tests, documentation, and local verification under maintainer direction.

## Type of Change

- [x] Bugfix
- [x] New Feature
- [ ] Breaking Change
- [x] Refactor
- [x] Documentation
- [ ] Other (please describe)

## Temporary Helm Charts (Optional)

For chart changes, you can test temporary versions before merging:
- **Base repo contributors:** Create a branch starting with `prebuild/` for automatic prebuilds
- Temporary charts are published to `ghcr.io/caipe-io/charts` with PR-specific versions

## Verification

- `npm run lint`
- `npm run build`
- 7 focused impersonation/auth/request-guard/UI Jest suites: 263 tests passed
- `tests/test_validate_rbac_docs.py`: 32 tests passed
- Production and development Docker Compose config validation with the minimal first-install profiles
- Identity-boundary, chat-store, auth-provider, admin-users, chat navigation, New Chat, dynamic-agent LLM/retry, and audit-service suites from the earlier implementation passes
- Local Keycloak token exchange for a federated required-group test user
- Live Azure GPT-5 Responses smoke call with a bound function tool
- Clean Docker UI restart and chat reload without a hydration mismatch

## Checklist

- [x] This PR is under ~500 changed lines of hand-written code, or the description explains why it cannot be split
- [x] AI assistance used to produce this change is disclosed in the description
- [x] Review suggestions were applied locally, not with the "Commit suggestion" button (it adds a co-author who has not signed off, which fails DCO)
- [x] New selection controls follow the [selection-control decision table](https://github.com/caipe-io/ai-platform-engineering/blob/main/docs/docs/ui/selection-controls.md), or the custom interaction is justified


sriaradhyula:
@kevkantes - Is this restricted to admin page for admin debugging?

ui/src/lib/auth-config.ts
```diff
@@ -601,6 +675,146 @@ export const authOptions: NextAuthOptions = {
...
+          const stored = await getStoredTokens(activeImpersonation.storeKey);
+          const now = Math.floor(Date.now() / 1000);
+          if (
+            !stored?.accessToken
+            || activeImpersonation.expiresAt - now < IMPERSONATION_RENEWAL_WINDOW_SECONDS
+          ) {
+            const { mintImpersonatedUserToken } = await import("@/lib/auth/user-impersonation");
+            const minted = await mintImpersonatedUserToken(activeImpersonation.targetSub);
+            const groups = extractGroups(minted.claims);
+            if (!hasRequiredGroup(groups)) {
```
cisco-erilutz, line 790: **Revocation lag on token renewal.** This renewal block rechecks `hasRequiredGroup` on the *target* but never reruns `hasOrganizationAdmin` on the *actor* (unlike the start path at lines 706-712). Because `mintImpersonatedUserToken` exchanges via a static service-account credential rather than the admin's own token, renewal is fully decoupled from the admin's session health — there's no back-channel logout. If an admin's `can_manage` grant is revoked mid-impersonation, the session can keep renewing for up to the 24h JWT cookie `maxAge` with no recheck. This is the one admin action path that skips the live per-request admin recheck every other admin action gets. Suggest rechecking `hasOrganizationAdmin(actor)` here too.

ui/src/components/admin/teams/UserDetailModal.tsx
```diff
@@ -437,6 +435,31 @@ export function UserDetailModal({
       ? formatTs(user.createdAt)
       : "—";
 
+  const startImpersonation = async (): Promise<void> => {
+    if (!user || !user.enabled || user.id === session?.sub) return;
```
cisco-erilutz, line 439: **Missing guard against switching impersonation targets mid-session.** This guard only blocks `!user.enabled || user.id === session?.sub` — it doesn't check `session?.impersonation`. `UserImpersonationPanel.tsx` hides its picker while impersonation is active, but this modal doesn't have the equivalent guard, so a direct B→C target switch is reachable here (e.g. via Admin > Teams while impersonating a user who also has admin visibility). `clearImpersonation` still revokes B's token immediately — no capability leak — but no explicit "stop" audit event is logged for B, only a "start" event for C. Suggest adding the same `session?.impersonation` guard here, or better, always emit a stop event before starting a new impersonation regardless of entry point.

ui/src/app/api/admin/impersonation/users/route.ts
```diff
@@ -0,0 +1,89 @@
...
+    search,
+    enabled: true,
+    first: (page - 1) * pageSize,
+    max: pageSize,
+  });
+  const federationRequired = (await listIdpAliases()).some(
+    (provider) => provider.enabled !== false,
+  );
+  const excludedSubjects = new Set(
+    [session.impersonatedBySub, session.sub].filter(
```
cisco-erilutz, line 53: **Eligible-target list doesn't exclude other org-admins.** `excludedSubjects` only excludes self and the current impersonation actor — nothing stops an admin from impersonating another org-admin. Confirmed this grants no additional capability today (flat single-org OpenFGA model — every admin already has `can_manage`), so it's not an active vulnerability. But that "no extra power" property is emergent from today's flat model, not an enforced invariant — it would silently break if a scoped-admin tier is ever introduced. It's also a UX/hygiene gap: the picker gives no indication a candidate is an admin. Suggest excluding (or at least flagging) org-admins from the eligible-target list.

cisco-erilutz (commented):
Went through this with a fine-toothed comb on the security side (multiple independent passes, including adversarial attempts to break the admin-only gate). Overall: the core boundary is solid — `token.sub` is never reassigned, so a non-admin cannot start impersonation, and I couldn't find a privilege-escalation or cross-user data-leak path. Leaving 3 inline comments on files touched by this PR (cheap fixes worth doing before merge), plus 2 more as general PR comments since they point at files outside this diff.

cisco-erilutz:
**Audit-ingestion endpoint has no authentication and trusts client-supplied actor/impersonation fields** (`ai_platform_engineering/audit_service/main.py:229`, `POST /v1/audit/events`).

This route calls `AuditEvent.model_validate(raw)` directly on the request body with no auth check, so a caller reaching this port could set/omit `actor_ref`, `impersonation`, `impersonation_started_at` directly, bypassing all the server-side derivation this PR adds in the BFF. Docker Compose binds this to `127.0.0.1` (safe), but the Helm chart exposes it as a plain `ClusterIP` with no `NetworkPolicy` anywhere in the chart — any pod in-cluster could forge audit records in a real deployment.

This endpoint predates this PR and is out of scope for it — not something to fix here — but this PR is what makes those fields security-critical for impersonation accountability, so it's worth flagging. **This could be a good candidate for a followup issue** (a shared-secret/mTLS check on the endpoint, or a NetworkPolicy restricting ingress).

cisco-erilutz:
**`ImpersonationTokenMissing` isn't handled by `token-expiry-guard.tsx`'s error set** (`ui/src/components/token-expiry-guard.tsx:18-19`).

`SESSION_CREDENTIAL_ERRORS` only contains `RefreshTokenExpired`/`RefreshTokenError`; the `ImpersonationTokenMissing` error set in `auth-config.ts` isn't in this set, so it won't trigger a forced clean re-login. In practice this fails closed (downstream calls needing the missing token 401, and it self-heals on the next renewal cycle) rather than failing open, so it's low severity — but worth adding to this set for a cleaner recovery path instead of silent 401s.

ui/src/lib/auth-config.ts
kevkantes: Fixed in cfd83f45a. Target-token renewal now reruns `canStartUserImpersonation` for the original actor, which rechecks both the explicit email allowlist and current Super Admin membership before minting another target bearer. A failed recheck clears the stored bearer and ends impersonation; the regression test covers revocation before renewal.

ui/src/app/api/admin/impersonation/users/route.ts
kevkantes: Superseded by the final design. The standalone eligible-target endpoint and picker were removed; impersonation is only launched from a specific user modal. More importantly, the server now requires the actor to be both explicitly email-allowlisted and a current true Super Admin, and the impersonated session is read-only. That enforces the no-additional-capability invariant even if scoped admin tiers are added later, so excluding admin targets is no longer needed for privilege containment.

ui/src/components/admin/teams/UserDetailModal.tsx
kevkantes: Fixed in cfd83f45a. The modal no longer offers impersonation while `session.impersonation` is active, and the JWT callback independently rejects a second start request while an impersonation exists. This preserves the original target and bearer instead of silently switching; both layers have regression coverage.

ai_platform_engineering/dynamic_agents/tests/test_middleware_defaults.py
```diff
@@ -42,3 +50,24 @@ def test_build_middleware_uses_context_editing_guardrail_by_default():
     assert context_editing.edits[0].keep == 3
     assert any(isinstance(middleware, ToolResultInvariantMiddleware) for middleware in stack)
     assert any(isinstance(middleware, ContextUsageMiddleware) for middleware in stack)
+
+
+def test_model_retry_does_not_retry_terminal_client_errors():
+    middleware = TransientModelRetryMiddleware(
+        max_retries=5,
+        initial_delay=0,
+        on_failure=lambda exc: str(exc),
```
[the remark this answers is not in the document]
kevkantes: Addressed with `on_failure=str`; the fix is preserved after rebasing onto current `main`.

kevkantes:
> **`ImpersonationTokenMissing` isn't handled by `token-expiry-guard.tsx`'s error set** (`ui/src/components/token-expiry-guard.tsx:18-19`).
>
> `SESSION_CREDENTIAL_ERRORS` only contains `RefreshTokenExpired`/`RefreshTokenError`; the `ImpersonationTokenMissing` error set in `auth-config.ts` isn't in this set, so it won't trigger a forced clean re-login. In practice this fails closed (downstream calls needing the missing token 401, and it self-heals on the next renewal cycle) rather than failing open, so it's low severity — but worth adding to this set for a cleaner recovery path instead of silent 401s.

@cisco-erilutz Fixed in cfd83f45a. `ImpersonationTokenMissing` now enters the clean re-login flow, and the token-expiry guard test covers the recovery behavior.

kevkantes:
> **Audit-ingestion endpoint has no authentication and trusts client-supplied actor/impersonation fields** (`ai_platform_engineering/audit_service/main.py:229`, `POST /v1/audit/events`).
>
> This route calls `AuditEvent.model_validate(raw)` directly on the request body with no auth check, so a caller reaching this port could set/omit `actor_ref`, `impersonation`, `impersonation_started_at` directly, bypassing all the server-side derivation this PR adds in the BFF. Docker Compose binds this to `127.0.0.1` (safe), but the Helm chart exposes it as a plain `ClusterIP` with no `NetworkPolicy` anywhere in the chart — any pod in-cluster could forge audit records in a real deployment.
>
> This endpoint predates this PR and is out of scope for it — not something to fix here — but this PR is what makes those fields security-critical for impersonation accountability, so it's worth flagging. **This could be a good candidate for a followup issue** (a shared-secret/mTLS check on the endpoint, or a NetworkPolicy restricting ingress).

@cisco-erilutz Agreed. This predates the impersonation work and should not expand this PR into audit-service transport security. I opened #2816 to track authenticated ingestion, producer configuration, ingress restriction, and regression coverage. That issue remains open.

